PR #54, opened 2015-01-29
Added direct atomic support for x64 Unix replacing mintomics

phaynes:
Replaced requirement for Mintomics through direct implementation of lock support.
Key goal was JDK compatibility although a couple of mintomic's functions were copied. 
 
A mintomic's declaration remains to minimise other code changes at this point.


tmontgomery:
Awesome! Thanks! I'll be looking at this tonight and tomorrow.

phaynes:
Todd, when looking at the code, reading it with the JDK source would be handy.  A couple of functions have "+m" in the assembler. These are lifted from mintomics. for the getAndSetIntXX functions, I trolled through the unsafe.cpp function ran out of time trying to find where it was implemented to trace down behaviour.

RichardWarburton:
@phaynes You'll find that the guts of Unsafe are really implemented in hotspot as JVM intrinsics in order to avoid the overhead of a JNI call when you just want a single assembly instruction emitted. To the best of my knowledge this is all defined in `library_call.cpp` (http://hg.openjdk.java.net/jdk8/jdk8/hotspot/file/tip/src/share/vm/opto/library_call.cpp). If you care about get and set operations then see line 806 and 2797 onwards. Unfortunately it might be a bit hard to read what is going on because intrinsics in hotspot (as with many other VMs) are written as functions that generate JIT compiler IR sub-graphs rather than assembly, but it is the canonical source of truth.

phaynes:
@RichardWarburton Thanks very much for pointing this out. I have been experiencing a bit of of cognitive dissonance this past couple of weeks reading code purporting to implement Unsafe not 100% lining up with the instructions actually emitted from the JVM despite my test cases passing. library_call.cpp doesn't seem that much harder to reading than the JNI implementation with the added advantage that it what is actually called. I will continue check implementations.

PR #73, opened 2015-03-22
added troubleshooting and sample configuration doc

coderfi:
This helps diagnose why the samples might fail to run inside a Docker or in the situations when the sample programs are pointing to mismatched temporary directories, or if the temporary directory runs out of disk space.


mjpt777:
Thanks

PR #75, opened 2015-03-27
Fix #68, attempt to determine the protocol family and use it on Datagram...

### LINKED ISSUE #68: Linux and IPv4/IPv6 sockets
On Linux, subscribers appear to create UDP6 sockets.

Using the command line

`java -cp aeron-samples/build/libs/samples.jar -Daeron.dir.conductor=/tmp/aeron/conductor -Daeron.dir.data=/tmp/aeron/data -Daeron.sample.channel=udp://172.16.179.139:50000 uk.co.real_logic.aeron.samples.BasicSubscriber`

on a Linux VM, a listening socket was created, but reported by netstat as

`udp6 0 0 172.16.179.139:50000 [::]:*`

One (rather unsatisfactory) workaround is to disable IPv6 on Linux.

Doing some code sleuthing, the call to `DatagramChannel.open()` doesn’t specify a protocol family, and according to [this link](http://docs.oracle.com/javase/8/docs/api/java/nio/channels/DatagramChannel.html#open--):

> The ProtocolFamily of the channel's socket is platform (and possibly configuration) dependent and therefore unspecified.

An overloaded call, `DatagramChannel.open(ProtocolFamily family)`, allows the protocol family (e.g., IPv4 or IPv6) to be specified. If an IPv4 address is specified in the URI, then an IPv4 socket should be opened, and similarly for an IPv6 address.

I was able to work around this by adding

`-Djava.net.preferIPv4Stack=true`

to each Java command line (media driver, publisher, and subscriber)

mikeb01:
...Channel.open.

### DIFF
--- a/aeron-driver/src/main/java/uk/co/real_logic/aeron/driver/UdpChannel.java
+++ b/aeron-driver/src/main/java/uk/co/real_logic/aeron/driver/UdpChannel.java
@@ -22,13 +22,7 @@ import static uk.co.real_logic.aeron.common.NetworkUtil.filterBySubnet;
 import static uk.co.real_logic.aeron.common.NetworkUtil.findAddressOnInterface;
 import static uk.co.real_logic.aeron.common.Strings.isEmpty;
 
-import java.net.InetAddress;
-import java.net.InetSocketAddress;
-import java.net.NetworkInterface;
-import java.net.SocketException;
-import java.net.URI;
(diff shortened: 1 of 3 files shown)


mjpt777:
Thanks Mike.

dameiss:
Yes, this seems to fix the problem. Thanks!

PR #85, opened 2015-04-10
A trimmed down version of Javadoc

vksingh70:
A trimmed down version of Javadoc for Aeron APIs. Two new tasks 'docs' and 'docsJar' have been added to the 'build.gradle'. In addition, more clarifications have been added to client APIs.


RichardWarburton:
Docs look good. Is there any reason we don't make the `conclude()` method of the `Aeron.Context` class package scoped? This would discourage API users from calling it.

tmontgomery:
I have no issue making `conclude` package scope.

brianhorst:
It should probably be protected or be able to be called more than once. conclude is overridden from CommonContext which is in another package so it can't be package scoped.

PR #97, opened 2015-04-20
Change the embedded driver directories to work better on multi-user systems. 

rlankenau:
Now appends a UUID to the base directory, and deletes the base directory on exit.

Avoids issues due to concurrent builds by different users on a shared system.


rlankenau:
Too fast on my feet.  Looks like there's another place where a random dir generated that can leave a trailing /tmp/aeron or /tmp/shm/aeron with the original user's permissions as well.  I'll fix that and update the pull request.

PR #101, opened 2015-04-24
[Java] benchmark of writing to a DatagramChannel

David-Wales:
NIO performance test using write instead of send on a DatagramChannel


mjpt777:
Thanks for the PR. I had to fix compilation and checkstyle errors. I guess your fork has not by kept up to date.

`DatagramChannel.write(ByteBuffer)` is giving a ~0.3us reduction in latency compared to `DatagramChannel.send(ByteBuffer, InetSocketAddress)` so this could be a worthy addition to shaving a little further latency off Aeron. Results are from testing with Java 1.8.0_45 on Linux 3.19 kernel and 2.2 Ghz Ivy Bridge processor.

PR #108, opened 2015-04-30
Doc, packaging, and tools for Aeron

strangelydim:
This is a big one...  Tools for performance testing and graph generation, for testing Aeron in general with multiple publications/subscriptions and channels and sending/receiving threads, abuse testing, some new and updated sample apps, new doc, and a test plan.

Let me know if changes need to be made before acceptance.  (Or if you'd like to accept first just to get this all in and make changes later via another pull request.)

### DIFF
--- a/aeron-tools/src/test/java/uk/co/real_logic/aeron/tools/PubSubOptionsTest.java
+++ b/aeron-tools/src/test/java/uk/co/real_logic/aeron/tools/PubSubOptionsTest.java
@@ -159,7 +159,7 @@ public class PubSubOptionsTest
         final String[] args = {};
         opts.parseArgs(args);
         assertThat("FAIL: default input stream should be null.",
-                opts.input(), is(nullValue()));
+            opts.input(), is(nullValue()));
     }
 
     @Test
@@ -168,7 +168,7 @@ public class PubSubOptionsTest
         final String[] args = {};
         opts.parseArgs(args);
         assertThat("FAIL: Default output stream should be null",
-                opts.output(), is(nullValue()));
+            opts.output(), is(nullValue()));
     }
 
     @Test
@@ -177,7 +177,7 @@ public class PubSubOptionsTest
         final String[] args = {"-o", "NULL"};
         opts.parseArgs(args);
         assertThat("FAIL: Output stream should be null when set to the string 'null'",
-                opts.output(), is(nullValue()));
+            opts.output(), is(nullValue()));
     }
 
     @Test
@@ -186,7 +186,7 @@ public class PubSubOptionsTest
         final String[] args = {"--output", "StdOut"};
         opts.parseArgs(args);
         assertThat("FAIL: Expected output to be the standard out",
-                opts.output(), equalTo(System.out));
+            opts.output(), equalTo(System.out));
     }
 
     @Test
@@ -195,7 +195,7 @@ public class PubSubOptionsTest
         final String[] args = {"--output", "stderr"};
         opts.parseArgs(args);
         assertThat("FAIL: Expected output to be standard error.",
-                opts.output(), equalTo(System.err));
+            opts.output(), equalTo(System.err));
     }
 
     @Test
@@ -204,17 +204,17 @@ public class PubSubOptionsTest
         final String[] args = {"--channels", "udp://127.0.0.1:12345"};
         opts.parseArgs(args);
         assertThat("FAIL: Exactly one channel.",
-                opts.channels().size(), is(1));
+            opts.channels().size(), is(1));
 
         final ChannelDescriptor cd = opts.channels().get(0);
         final int[] streamIds = cd.streamIdentifiers();
 
         assertThat("FAIL: Exactly one stream identifier on the channel.",
-                streamIds.length, is(1));
+            streamIds.length, is(1));
         assertThat("FAIL: Channel is udp://127.0.0.1:12345",
-                cd.channel(), is("udp://127.0.0.1:12345"));
+            cd.channel(), is("udp://127.0.0.1:12345"));
         assertThat("FAIL: Stream ID is 1",
-                cd.streamIdentifiers()[0], is(1));
+            cd.streamIdentifiers()[0], is(1));
     }
 
     @Test
@@ -224,17 +224,17 @@ public class PubSubOptionsTest
         opts.parseArgs(args);
 
         assertThat("FAIL: Exactly one channel.",
-                opts.channels().size(), is(1));
+            opts.channels().size(), is(1));
 
         final ChannelDescriptor cd = opts.channels().get(0);
         final int[] streamIds = cd.streamIdentifiers();
 
         assertThat("FAIL: Exactly one stream identifier on the channel.",
-                streamIds.length, is(1));
+            streamIds.length, is(1));
         assertThat("FAIL: Channel is udp://127.0.0.1:12345",
-                cd.channel(), is("udp://127.0.0.1:12345"));
+            cd.channel(), is("udp://127.0.0.1:12345"));
         assertThat("FAIL: Stream ID is 100",
-                cd.streamIdentifiers()[0], is(100));
+            cd.streamIdentifiers()[0], is(100));
     }
 
     @Test
@@ -244,13 +244,13 @@ public class PubSubOptionsTest
         opts.parseArgs(args);
 
         assertThat("FAIL: Expected 3 channels.",
-                opts.channels().size(), is(3));
+            opts.channels().size(), is(3));
         assertThat("FAIL: Channel 1 incorrect.",
-                opts.channels().get(0).channel(), is("udp://127.0.0.1:12345"));
+            opts.channels().get(0).channel(), is("udp://127.0.0.1:12345"));
         assertThat("FAIL: Channel 2 incorrect",
-                opts.channels().get(1).channel(), is("udp://127.0.0.1:12346"));
+            opts.channels().get(1).channel(), is("udp://127.0.0.1:12346"));
         assertThat("FAIL: Channel 3 incorrect",
-                opts.channels().get(2).channel(), is("udp://127.0.0.1:12347"));
+            opts.channels().get(2).channel(), is("udp://127.0.0.1:12347"));
     }
 
     @Test
@@ -260,17 +260,17 @@ public class PubSubOptionsTest
         opts.parseArgs(args);
 
         assertThat("FAIL: Expected 1 channel.",
-                opts.channels().size(), is(1));
+            opts.channels().size(), is(1));
         final ChannelDescriptor cd = opts.channels().get(0);
 
         assertThat("FAIL: Expected 3 stream IDs on channel.",
-                cd.streamIdentifiers().length, is(3));
+            cd.streamIdentifiers().length, is(3));
         assertThat("FAIL: stream-id 1 is wrong value.",
-                cd.streamIdentifiers()[0], is(100));
+            cd.streamIdentifiers()[0], is(100));
         assertThat("FAIL: Stream-id 2 is wrong value.",
-                cd.streamIdentifiers()[1], is(101));
+            cd.streamIdentifiers()[1], is(101));
         assertThat("FAIL: Stream-id 3 is wrong value.",
-                cd.streamIdentifiers()[2], is(102));
+            cd.streamIdentifiers()[2], is(102));
     }
 
     /**
@@ -282,47 +282,47 @@ public class PubSubOptionsTest
     {
         ChannelDescriptor cd;
         final String[] args = {"--channels",
-                "udp://127.0.0.1:5000-5001#1-2,udp://224.9.10.11:6000-6001#600-601,udp://192.168.0.1:7000-7001#700-701"};
+            "udp://127.0.0.1:5000-5001#1-2,udp://224.9.10.11:6000-6001#600-601,udp://192.168.0.1:7000-7001#700-701"};
         opts.parseArgs(args);
 
         assertThat("FAIL: Expected 6 channels",
-                opts.channels().size(), is(6));
+            opts.channels().size(), is(6));
 
         cd = opts.channels().get(0);
         assertThat("FAIL: Wrong address for channel 1",
-                cd.channel(), is("udp://127.0.0.1:5000"));
+            cd.channel(), is("udp://127.0.0.1:5000"));
         assertThat("FAIL: Wrong number of stream IDs on channel 1",
-                cd.streamIdentifiers().length, is(2));
+            cd.streamIdentifiers().length, is(2));
 
         cd = opts.channels().get(1);
         assertThat("FAIL: Wrong address for channel 2",
-                cd.channel(), is("udp://127.0.0.1:5001"));
+            cd.channel(), is("udp://127.0.0.1:5001"));
         assertThat("FAIL: Wrong number of stream IDs on channel 2",
-                cd.streamIdentifiers().length, is(2));
+            cd.streamIdentifiers().length, is(2));
 
         cd = opts.channels().get(2);
         assertThat("FAIL: Wrong address for channel 3",
-                cd.channel(), is("udp://224.9.10.11:6000"));
+            cd.channel(), is("udp://224.9.10.11:6000"));
         assertThat("FAIL: Wrong number of stream IDs on channel 3",
-                cd.streamIdentifiers().length, is(2));
+            cd.streamIdentifiers().length, is(2));
 
         cd = opts.channels().get(3);
         assertThat("FAIL: Wrong address for channel 4",
-                cd.channel(), is("udp://224.9.10.11:6001"));
+            cd.channel(), is("udp://224.9.10.11:6001"));
         assertThat("FAIL: Wrong number of stream IDs on channel 4",
-                cd.streamIdentifiers().length, is(2));
+            cd.streamIdentifiers().length, is(2));
 
         cd = opts.channels().get(4);
         assertThat("FAIL: Wrong address for channel 5",
-                cd.channel(), is("udp://192.168.0.1:7000"));
+            cd.channel(), is("udp://192.168.0.1:7000"));
         assertThat("FAIL: Wrong number of stream IDs on channel 5",
-                cd.streamIdentifiers().length, is(2));
+            cd.streamIdentifiers().length, is(2));
 
         cd = opts.channels().get(5);
         assertThat("FAIL: Wrong address for channel 6",
-                cd.channel(), is("udp://192.168.0.1:7001"));
+            cd.channel(), is("udp://192.168.0.1:7001"));
         assertThat("FAIL: Wrong number of stream IDs on channel 6",
-                cd.streamIdentifiers().length, is(2));
+            cd.streamIdentifiers().length, is(2));
     }
 
     @Test
@@ -333,10 +333,10 @@ public class PubSubOptionsTest
         opts.parseArgs(args);
 
         assertThat("FAIL: Expected 1 channel",
-                opts.channels().size(), is(1));
+            opts.channels().size(), is(1));
         cd = opts.channels().get(0);
         assertThat("FAIL wrong address for channel 1",
-                cd.channel(), is("udp://[fe80::1234:2345:3456:4567]:12345"));
+            cd.channel(), is("udp://[fe80::1234:2345:3456:4567]:12345"));
     }
 
     @Test
@@ -349,10 +349,10 @@ public class PubSubOptionsTest
         assertThat(opts.channels().size(), is(2));
         cd = opts.channels().get(0);
         assertThat("FAIL: wrong address for channel 1",
-                cd.channel(), is("aeron:udp?remote=192.168.14.101:10000|local=192.168.14.102"));
+            cd.channel(), is("aeron:udp?remote=192.168.14.101:10000|local=192.168.14.102"));
         cd = opts.channels().get(1);
         assertThat("FAIL: wrong address for channel 2",
-                cd.channel(), is("aeron:udp?remote=192.168.14.101:10001|local=192.168.14.102"));
+            cd.channel(), is("aeron:udp?remote=192.168.14.101:10001|local=192.168.14.102"));
     }
 
     @Test
@@ -365,7 +365,7 @@ public class PubSubOptionsTest
         assertThat(opts.channels().size(), is(1));
         cd = opts.channels().get(0);
         assertThat("FAIL: wrong address for channel 1",
-                cd.channel(), is("aeron:udp?remote=[::1]:12345|local=[::1]"));
+            cd.channel(), is("aeron:udp?remote=[::1]:12345|local=[::1]"));
     }
 
     @Test
@@ -380,11 +380,12 @@ public class PubSubOptionsTest
     @Test
     public void numberOfStreamsAeronUri() throws Exception
     {
-        final String[] args = {
-            "--channels",
-            "aeron:udp?group=224.10.20.21:12345#1-10|interface=127.0.0.1,"+
-                "aeron:udp?remote=localhost:13333#1|local=127.0.0.1"
-        };
+        final String[] args =
+            {
+                "--channels",
+                "aeron:udp?group=224.10.20.21:12345#1-10|interface=127.0.0.1," +
+                    "aeron:udp?remote=localhost:13333#1|local=127.0.0.1"
+            };
 
         opts.parseArgs(args);
         assertThat(opts.numberOfStreams(), is(11));
@@ -393,7 +394,7 @@ public class PubSubOptionsTest
     @Test
     public void messageSizes() throws Exception
     {
-        final String[] args = { "--size", "100" };
+        final String[] args = {"--size", "100"};
         opts.parseArgs(args);
         final MessageSizePattern p = opts.messageSizePattern();
         assertThat(p.getNext(), is(100));
@@ -402,7 +403,7 @@ public class PubSubOptionsTest
     @Test
     public void messageSizeShortHand() throws Exception
     {
-        final String[] args = { "-s", "100" };
+        final String[] args = {"-s", "100"};
         opts.parseArgs(args);
         final MessageSizePattern p = opts.messageSizePattern();
         assertThat(p.getNext(), is(100));
@@ -411,7 +412,7 @@ public class PubSubOptionsTest
     @Test
     public void messageSizeRange() throws Exception
     {
-        final String[] args = { "--size", "101-102" };
+        final String[] args = {"--size", "101-102"};
         opts.parseArgs(args);
         final MessageSizePattern p = opts.messageSizePattern();
         assertThat(p.getNext(), both(greaterThanOrEqualTo(101)).and(lessThanOrEqualTo(102)));
@@ -420,7 +421,7 @@ public class PubSubOptionsTest
     @Test
     public void messageNumberAndSize() throws Exception
     {
-        final String[] args = { "--size", "1@100" };
+        final String[] args = {"--size", "1@100"};
         opts.parseArgs(args);
         final MessageSizePattern p = opts.messageSizePattern();
         assertThat(p.getNext(), is(100));
@@ -429,7 +430,7 @@ public class PubSubOptionsTest
     @Test
     public void messageNumberAndRange() throws Exception
     {
-        final String[] args = { "--size", "1@101-102" };
+        final String[] args = {"--size", "1@101-102"};
         opts.parseArgs(args);
         final MessageSizePattern p = opts.messageSizePattern();
         assertThat(p.getNext(), both(greaterThanOrEqualTo(101)).and(lessThanOrEqualTo(102)));
@@ -438,7 +439,7 @@ public class PubSubOptionsTest
     @Test
     public void messageNumberAndRangeCsv() throws Exception
     {
-        final String[] args = { "--size", "1@100,1@101-102,98@1000" };
+        final String[] args = {"--size", "1@100,1@101-102,98@1000"};
         opts.parseArgs(args);
         final MessageSizePattern p = opts.messageSizePattern();
         assertThat(p.getNext(), is(100));
@@ -449,7 +450,7 @@ public class PubSubOptionsTest
     @Test
     public void messageSizeBytesSuffix() throws Exception
     {
-        final String[] args = { "--size", "1@100B,1@101b" };
+        final String[] args = {"--size", "1@100B,1@101b"};
         opts.parseArgs(args);
         final MessageSizePattern p = opts.messageSizePattern();
         assertThat(p.getNext(), is(100));
@@ -459,7 +460,7 @@ public class PubSubOptionsTest
     @Test
     public void messageSizeKilobytesSuffix() throws Exception
     {
-        final String[] args = { "--size", "1@100K,1@101k,1@102KB,1@103kb" };
+        final String[] args = {"--size", "1@100K,1@101k,1@102KB,1@103kb"};
         opts.parseArgs(args);
         final MessageSizePattern p = opts.messageSizePattern();
         assertThat(p.getNext(), is(100 * 1024));
@@ -471,7 +472,7 @@ public class PubSubOptionsTest
     @Test
     public void messageSizeMegabytesSuffix() throws Exception
     {
-        final String[] args = { "--size", "1@100M,1@101m,1@102MB,1@103mb" };
+        final String[] args = {"--size", "1@100M,1@101m,1@102MB,1@103mb"};
         opts.parseArgs(args);
         final MessageSizePattern p = opts.messageSizePattern();
         assertThat(p.getNext(), is(100 * 1024 * 1024));
@@ -483,7 +484,7 @@ public class PubSubOptionsTest
     @Test
     public void messageSizeRangesWithSuffixes() throws Exception
     {
-        final String[] args = { "--size", "1@1023B-1KB,1@1023KB-1MB" };
+        final String[] args = {"--size", "1@1023B-1KB,1@1023KB-1MB"};
         opts.parseArgs(args);
         final MessageSizePattern p = opts.messageSizePattern();
         assertThat(p.getNext(), both(greaterThanOrEqualTo(1023)).and(lessThanOrEqualTo((1024))));
@@ -493,62 +494,62 @@ public class PubSubOptionsTest
     @Test
     public void rateMaximum() throws Exception
     {
-        final String[] args = { "--rate", "max" };
+        final String[] args = {"--rate", "max"};
         opts.parseArgs(args);
         assertThat(opts.rateIntervals().size(), is(1));
 
         final RateControllerInterval interval = opts.rateIntervals().get(0);
         assertThat(interval, instanceOf(SecondsAtBitsPerSecondInterval.class));
-        final SecondsAtBitsPerSecondInterval sub = (SecondsAtBitsPerSecondInterval)interval;
-        assertThat(sub.seconds(), is((double)Long.MAX_VALUE));
+        final SecondsAtBitsPerSecondInterval sub = (SecondsAtBitsPerSecondInterval) interval;
+        assertThat(sub.seconds(), is((double) Long.MAX_VALUE));
         assertThat(sub.bitsPerSecond(), is(Long.MAX_VALUE));
     }
 
     @Test
     public void rateMbps() throws Exception
     {
-        final String[] args = { "--rate", "1Gbps" };
+        final String[] args = {"--rate", "1Gbps"};
         opts.parseArgs(args);
         assertThat(opts.rateIntervals().size(), is(1));
 
         final RateControllerInterval interval = opts.rateIntervals().get(0);
         assertThat(interval, instanceOf(SecondsAtBitsPerSecondInterval.class));
-        final SecondsAtBitsPerSecondInterval sub = (SecondsAtBitsPerSecondInterval)interval;
-        assertThat(sub.seconds(), is((double)Long.MAX_VALUE));
+        final SecondsAtBitsPerSecondInterval sub = (SecondsAtBitsPerSecondInterval) interval;
+        assertThat(sub.seconds(), is((double) Long.MAX_VALUE));
         assertThat(sub.bitsPerSecond(), is(1_000_000_000L));
     }
 
     @Test
     public void rateMessagesPerSecond() throws Exception
     {
-        final String[] args = { "--rate", "100mps" };
+        final String[] args = {"--rate", "100mps"};
         opts.parseArgs(args);
         assertThat(opts.rateIntervals().size(), is(1));
 
         final RateControllerInterval interval = opts.rateIntervals().get(0);
         assertThat(interval, instanceOf(SecondsAtMessagesPerSecondInterval.class));
-        final SecondsAtMessagesPerSecondInterval sub = (SecondsAtMessagesPerSecondInterval)interval;
-        assertThat(sub.seconds(), is((double)Long.MAX_VALUE));
+        final SecondsAtMessagesPerSecondInterval sub = (SecondsAtMessagesPerSecondInterval) interval;
+        assertThat(sub.seconds(), is((double) Long.MAX_VALUE));
         assertThat(sub.messagesPerSecond(), is(100D));
     }
 
-    @Test (expected = ParseException.class)
+    @Test(expected = ParseException.class)
     public void rateNotValid() throws Exception
     {
-        final String[] args = { "--rate", "1.21 Giga Watts" };
+        final String[] args = {"--rate", "1.21 Giga Watts"};
         opts.parseArgs(args); // will throw ParseException
     }
 
     @Test
     public void rateMessagesAtBitsPerSecond() throws Exception
     {
-        final String[] args = { "--rate", "100m@0.5Mbps" };
+        final String[] args = {"--rate", "100m@0.5Mbps"};
         opts.parseArgs(args);
         assertThat(opts.rateIntervals().size(), is(1));
 
         final RateControllerInterval interval = opts.rateIntervals().get(0);
         assertThat(interval, instanceOf(MessagesAtBitsPerSecondInterval.class));
-        final MessagesAtBitsPerSecondInterval sub = (MessagesAtBitsPerSecondInterval)interval;
+        final MessagesAtBitsPerSecondInterval sub = (MessagesAtBitsPerSecondInterval) interval;
         assertThat(sub.messages(), is(100L));
         assertThat(sub.bitsPerSecond(), is(500000L));
     }
@@ -556,13 +557,13 @@ public class PubSubOptionsTest
     @Test
     public void rateMessagesAtMessagesPerSecond() throws Exception
     {
-        final String[] args = { "--rate", "10m@1mps" };
+        final String[] args = {"--rate", "10m@1mps"};
         opts.parseArgs(args);
         assertThat(opts.rateIntervals().size(), is(1));
 
         final RateControllerInterval interval = opts.rateIntervals().get(0);
         assertThat(interval, instanceOf(MessagesAtMessagesPerSecondInterval.class));
-        final MessagesAtMessagesPerSecondInterval sub = (MessagesAtMessagesPerSecondInterval)interval;
+        final MessagesAtMessagesPerSecondInterval sub = (MessagesAtMessagesPerSecondInterval) interval;
         assertThat(sub.messages(), is(10L));
         assertThat(sub.messagesPerSecond(), is(1D));
     }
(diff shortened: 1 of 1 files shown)


aeron-driver/src/main/java/uk/co/real_logic/aeron/driver/MediaDriver.java
```diff
@@ -195,6 +209,7 @@ public static MediaDriver launch(final Context ctx)
     /**
      * Shutdown the media driver by stopping all threads and freeing resources.
      */
+    @Override
```
mjpt777: What is the value in adding "@Override"? It just clutters up the code and is poorly defined for Java as most uses are implements rather than a true override.

aeron-samples/src/main/java/uk/co/real_logic/aeron/samples/AeronStat.java
```diff
@@ -29,31 +29,44 @@
...
+        // Check that command and control file exists, open file, and return MappedByteBuffer for entire file
         final MappedByteBuffer cncByteBuffer = IoUtil.mapExistingFile(cncFile, "cnc");
 
+        // Create a DirectBuffer for cnc's metadata
         final DirectBuffer metaDataBuffer = CncFileDescriptor.createMetaDataBuffer(cncByteBuffer);
 
+        // Extract version number of cnc from meta data
         final int cncVersion = metaDataBuffer.getInt(CncFileDescriptor.cncVersionOffset(0));
+
+        // Throw an exception if CNC version number mismatch is detected between this app and memory mapped cnc file
```
mjpt777: Really? Is a comment needed for things like this? I'd prefer no comments in a code base unless it is genuinely tricky to describe in code or referencing an external algorithm or paper.

aeron-samples/src/main/java/uk/co/real_logic/aeron/samples/MultipleSubscribersWithFragmentAssembly.java
```diff
@@ -0,0 +1,201 @@
...
+        // Register a SIGINT handler
+        SigInt.register(() -> running.set(false));
+
+        // Create an Aeron instance with client provided context configuration and connect to media driver
+        try (final Aeron aeron = Aeron.connect(ctx);
+                // Add a subscription to Aeron for a given channel and steam. Also,
+                // supply a dataHandler to be called when data arrives
+                final Subscription subscription1 = aeron.addSubscription(CHANNEL, STREAM_ID_1, dataHandler1);
+                final Subscription subscription2 = aeron.addSubscription(CHANNEL, STREAM_ID_2, dataHandler2))
+                {
```
mjpt777: Formatting seems a bit messed up in some of the files. Are you using tabs?

aeron-samples/src/main/java/uk/co/real_logic/aeron/samples/MultipleSubscribersWithFragmentAssembly.java
```diff
@@ -0,0 +1,201 @@
...
+            // Initialize a backoff strategy to avoid excessive spinning
+            final IdleStrategy idleStrategy = new BackoffIdleStrategy(
+                    100, 10, TimeUnit.MICROSECONDS.toNanos(1), TimeUnit.MICROSECONDS.toNanos(100));
```
mjpt777: Indent to 4 spaces.

aeron-samples/src/main/java/uk/co/real_logic/aeron/samples/StreamingPublisher.java
```diff
@@ -115,9 +121,11 @@ public static void main(final String[] args) throws Exception
 
                 printingActive = false;
             }
+            // Keep repeating the above loop if user answers 'Y' to "Execute again?"
+            // Otherwise, exit the loop
             while (barrier.await());
         }
-
+        // Halt the report
```
mjpt777: Please just don't do comments :-)

aeron-tools/src/main/java/uk/co/real_logic/aeron/tools/ChannelDescriptor.java
```diff
@@ -0,0 +1,58 @@
...
+ */
+package uk.co.real_logic.aeron.tools;
+
+/**
+ * This class is responsible for holding an Aeron channel and all the stream Ids that are on it.
+ * Created by bhorst on 3/4/15.
+ */
+public class ChannelDescriptor
+{
+    String channel;
```
mjpt777: Why package private?

aeron-tools/src/main/java/uk/co/real_logic/aeron/tools/MediaDriverOptions.java
```diff
@@ -0,0 +1,302 @@
...
+        }
+        else
+        {
+            // Use reflection to create the new IdleStrategy with no parameters.
+            try
+            {
+                final Class clazz = Class.forName(arg);
+                strategy = (IdleStrategy)clazz.newInstance();
+            }
+            catch (final ClassNotFoundException ex)
```
mjpt777: Could be improved with Java 7 multiple exception catch blocks.

aeron-tools/src/main/java/uk/co/real_logic/aeron/tools/MediaDriverOptions.java
```diff
@@ -0,0 +1,302 @@
...
+            maxYields = parseLong(params[1].trim());
+            minParkPeriodNs = parseLong(params[2].trim());
+            maxParkPeriodNs = parseLong(params[3].trim());
+
+        }
+        return makeBackoffIdleStrategy(maxSpins, maxYields, minParkPeriodNs, maxParkPeriodNs);
+    }
+
+    // Broken out into simple method for testing.
+    IdleStrategy makeBackoffIdleStrategy(final long maxSpins, final long maxYields,
```
mjpt777: All args on next line with hanging indent of 4.

aeron-tools/src/main/java/uk/co/real_logic/aeron/tools/MessageStream.java
```diff
@@ -0,0 +1,462 @@
...
+            System.out.printf("%02x ", buffer.getByte(i));
+            if (++pos % HEX_PRINT_WIDTH == 0)
+            {
+                System.out.println();
+            }
+        }
+        System.out.println();
+    }
+
+    /** Generates a message of random (within the constraints the MessageStream was
```
mjpt777: New line after double ** in javadoc

aeron-tools/src/main/java/uk/co/real_logic/aeron/tools/PubSubOptions.java
```diff
@@ -0,0 +1,1635 @@
...
+        final String messages;
+        final String output;
+        final String rate;
+        final String seed;
+        final String session;
+        final String size;
+        final String threads;
+        final String verify;
+
+        OptionValuesStruct(final String channels,
```
mjpt777: Hanging indent on next line for args.

aeron-tools/src/main/java/uk/co/real_logic/aeron/tools/PublisherTool.java
```diff
@@ -0,0 +1,495 @@
...
+ * limitations under the License.
+ */
+package uk.co.real_logic.aeron.tools;
+
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.util.ArrayList;
+
+import org.apache.commons.cli.ParseException;
+import org.slf4j.Logger;
```
mjpt777: Why introduce SLF for a dependency. I think this sets a bad example. We should support high performance logging and take no dependencies.

aeron-tools/src/main/java/uk/co/real_logic/aeron/tools/StatsDriver.java
```diff
@@ -0,0 +1,126 @@
...
+      output = new StatsConsoleOutput();
+    }
+    else if (command.hasOption("netstat"))
+    {
+      output = new StatsNetstatOutput();
+    }
+    else if (command.hasOption("csv"))
+    {
+      if (command.hasOption("file"))
+      {
```
mjpt777: Formatting screwed up.

aeron-tools/src/main/java/uk/co/real_logic/aeron/tools/TLRandom.java
```diff
@@ -0,0 +1,55 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package uk.co.real_logic.aeron.tools;
+
+import java.util.Random;
+import java.util.concurrent.ThreadLocalRandom;
+
+public class TLRandom
```
mjpt777: If using abbreviations do it with only the first letter capitalised. Or better just spell out for clarity.

aeron-tools/src/main/java/uk/co/real_logic/aeron/tools/MediaDriverOptions.java
```diff
@@ -0,0 +1,302 @@
...
+ */
+package uk.co.real_logic.aeron.tools;
+
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Properties;
+
+import org.apache.commons.cli.CommandLine;
```
mjpt777: Add 3rd party dependencies need to be agreed in advance.

aeron-tools/src/test/java/uk/co/real_logic/aeron/tools/RateControllerTest.java
```diff
@@ -0,0 +1,437 @@
...
+        final List<RateControllerInterval> ivlsList = new ArrayList<RateControllerInterval>();
+        final RateControllerInterval ivl = new MessagesAtMessagesPerSecondInterval(1, 1);
+        ivlsList.add(ivl);
+        rc = new RateController(new Callback(), ivlsList, 0);
+    }
+
+    @Test (expected = Exception.class)
+    public void createWithNegativeIterations() throws Exception
+    {
+        class Callback implements RateController.Callback
```
mjpt777: Why not just use a lambda expression?

aeron-samples/src/main/java/uk/co/real_logic/aeron/samples/AeronStat.java
brianhorst: This is a sample application. Everything it's doing should be clearly documented.

aeron-samples/src/main/java/uk/co/real_logic/aeron/samples/MultipleSubscribersWithFragmentAssembly.java
brianhorst: Yep the spacing is wrong, but it's just spaces since checkstyle rules prevents tabs.

aeron-samples/src/main/java/uk/co/real_logic/aeron/samples/AeronStat.java
mjpt777: Do you really think a comment like this adds value? :-)

aeron-tools/src/main/java/uk/co/real_logic/aeron/tools/ChannelDescriptor.java
brianhorst: Should be private. I checked the rest of the code and the accessors are used everywhere.

aeron-tools/src/main/java/uk/co/real_logic/aeron/tools/PublisherTool.java
brianhorst: Do you have a suggestion of what to use instead? The logging framework is not used in any performance critical areas, and the publisher tool is a functional verification application not a performance test.

aeron-driver/src/main/java/uk/co/real_logic/aeron/driver/MediaDriver.java
brianhorst: These are probably just added by the various IDEs. We can remove them.

aeron-tools/src/main/java/uk/co/real_logic/aeron/tools/PublisherTool.java
mjpt777: The problem is by adding the tools module it drags in other dependencies. Why not go for standard java.util.logging if not performance critical?

brianhorst: ok, we can do that.

johnpr01: Will be moving the subproject aeron-tools out of top-level build.gradle and moving to aeron-tools/build.gradle and moving third-party dependencies into the aeron-tools/build.gradle and out of the top-level build.gradle. We will later probably go back and swap out the SLF with java.util.logging.

aeron-samples/src/main/java/uk/co/real_logic/aeron/samples/MultipleSubscribersWithFragmentAssembly.java
tmontgomery: The spacing and formatting should be consistent with the rest of the Aeron code base.

aeron-client/src/main/java/uk/co/real_logic/aeron/FragmentAssemblyAdapter.java
```diff
@@ -73,6 +75,7 @@ public FragmentAssemblyAdapter(final DataHandler delegate, final int initialBuff
      * @param length of the data in bytes.
      * @param header representing the meta data for the data.
      */
+    @Override
```
tmontgomery: Remove @Overrides that got added.

aeron-driver/src/main/java/uk/co/real_logic/aeron/driver/Configuration.java
```diff
@@ -314,6 +314,9 @@
     public static final String THREADING_MODE_PROP_NAME = "aeron.threading.mode";
     public static final String THREADING_MODE_DEFAULT = DEDICATED.name();
 
+    /** Disable the NACKs from media driver (used for QA only)*/
```
tmontgomery: Newline after **

aeron-driver/src/main/java/uk/co/real_logic/aeron/driver/DriverConductor.java
```diff
@@ -15,6 +15,30 @@
  */
 package uk.co.real_logic.aeron.driver;
 
+import static java.util.stream.Collectors.toList;
```
tmontgomery: I believe that the style is "static" imports after other imports.

aeron-driver/src/main/java/uk/co/real_logic/aeron/driver/DriverConductor.java
```diff
@@ -161,6 +184,7 @@ public DriverConductor(final Context ctx)
         toDriverCommands.consumerHeartbeatTimeNs(clock.time());
     }
 
+    @Override
```
tmontgomery: Remove @Overrides

aeron-driver/src/main/java/uk/co/real_logic/aeron/driver/MediaDriver.java
```diff
@@ -15,36 +15,50 @@
  */
 package uk.co.real_logic.aeron.driver;
 
-import uk.co.real_logic.aeron.common.*;
+import static java.lang.Boolean.getBoolean;
```
tmontgomery: Should "static" imports come first before other imports?

aeron-driver/src/main/java/uk/co/real_logic/aeron/driver/LossDetector.java
```diff
@@ -151,6 +151,10 @@ private void onGap(final int termId, final UnsafeBuffer buffer, final int offset
     private void activateGap(final int termId, final int termOffset, final int length)
     {
         activeGap.reset(termId, termOffset, length);
+        if (determineNakDelay() == -1)
```
tmontgomery: Wouldn't it be better to disable NAKs as a static final and just skip the entire activation instead? @mjpt777 thoughts?

aeron-samples/src/main/java/uk/co/real_logic/aeron/samples/MultipleSubscribersWithFragmentAssembly.java
```diff
@@ -0,0 +1,201 @@
...
+     * @param channel           for the connection
+     * @param streamId          for the stream
+     * @param sessionId         for the connection publication
+     * @param position          in the stream
+     * @param sourceInformation that is transport specific
+     */
+    public static void eventNewConnection(
+        final String channel, final int streamId, final int sessionId, final long position, final String sourceInformation)
+    {
+        System.out.println(
```
tmontgomery: System.out.format instead?

aeron-samples/src/main/java/uk/co/real_logic/aeron/samples/MultipleSubscribersWithFragmentAssembly.java
```diff
@@ -0,0 +1,201 @@
...
+            System.out.println(
+                String.format(
+                    "message to stream %d from session %x term id %x term offset %d (%d@%d)",
+                    streamId, header.sessionId(), header.termId(), header.termOffset(), length, offset));
+            if (length != 10000)
+            {
+                try
+                {
+                    throw new Exception("Received message is not assembled properly received lenth " +
+                            length + " Expecting " + "10000");
```
tmontgomery: indent to 4 spaces

aeron-samples/src/main/java/uk/co/real_logic/aeron/samples/SimplePublisher.java
```diff
@@ -0,0 +1,97 @@
...
+                }
+            }
+            else
+            {
+                // Successfully sent the message
+                System.out.println(" yay !!");
+            }
+            System.out.println("Done sending.");
+        }
+        finally
```
tmontgomery: Might be better to use try-with-resources version. See the BasicPublisher, etc.

aeron-tools/src/main/java/uk/co/real_logic/aeron/tools/ChannelDescriptor.java
```diff
@@ -0,0 +1,58 @@
...
+    int[] streamIds;
+
+    ChannelDescriptor()
+    {
+        channel = null;
+        streamIds = null;
+    }
+
+    public String getChannel()
```
tmontgomery: Please avoid "get" and "set" for accessors.

aeron-tools/src/main/java/uk/co/real_logic/aeron/tools/MediaDriverOptions.java
```diff
@@ -0,0 +1,302 @@
...
+    private IdleStrategy conductorIdleStrategy;
+    private IdleStrategy senderIdleStrategy;
+    private IdleStrategy receiverIdleStrategy;
+    private IdleStrategy sharedNetworkIdleStrategy;
+    private IdleStrategy sharedIdleStrategy;
+
+    public MediaDriverOptions()
+    {
+        options = new Options()
+                .addOption("p", "properties", true, "A properties file containing Aeron options.")
```
tmontgomery: indent to 4 spaces.

aeron-tools/src/main/java/uk/co/real_logic/aeron/tools/MediaDriverOptions.java
```diff
@@ -0,0 +1,302 @@
...
+            // Inform application the user wants help message.
+            return 1;
+        }
+
+        if (command.hasOption("properties"))
+        {
+            this.properties = parseProperties(command.getOptionValue("properties"));
+        }
+        String defaultValue;
+        defaultValue = properties == null ? "null" : properties.getProperty("aeron.tools.mediadriver.sender");
```
tmontgomery: This feels like it should be simpler and cleaner. While I am sure it works, it is a bit hard to follow. And I can feel @mjpt777 twitching.

aeron-tools/src/main/java/uk/co/real_logic/aeron/tools/MessageStream.java
```diff
@@ -0,0 +1,462 @@
...
+
+        /* Assume we've already checked and it appears we have a verifiable message. */
+
+        /* Sequence number check first. */
+        final long receivedSequenceNumber = buffer.getLong(offset + SEQUENCE_NUMBER_OFFSET);
+        final long expectedSequenceNumber = sequenceNumber + 1;
+        if (receivedSequenceNumber != expectedSequenceNumber)
+        {
+            final Exception e = new Exception("Verifiable message stream received sequence number " +
+                    receivedSequenceNumber + ", but was expecting " +
```
tmontgomery: indent to 4 spaces

aeron-tools/src/main/java/uk/co/real_logic/aeron/tools/RateController.java
```diff
@@ -0,0 +1,624 @@
...
+            Thread.yield();
+            nanoSecondsRemaining = nanoseconds - (CLOCK.time() - startNanos);
+        }
+        while (nanoSecondsRemaining > TIME_NANOS)
+        {
+            nanoSecondsRemaining = nanoseconds - (CLOCK.time() - startNanos);
+        }
+    }
+
+    static
```
tmontgomery: Move static block up after members. Makes it easier to find.

aeron-tools/src/main/java/uk/co/real_logic/aeron/tools/SubscriberTool.java
```diff
@@ -0,0 +1,719 @@
...
+        int activeSubscriptionsLength = 0;
+
+        private final Publication controlPublication;
+        private final Subscription controlSubscription;
+        /* Doesn't use TLRandom, since this really does need to be random and shouldn't
+         * be affected by manually setting the seed. */
+        private final int controlSessionId = ThreadLocalRandom.current().nextInt();
+
+        /* channel -> stream ID -> session ID */
+        private final HashMap<String, Int2ObjectHashMap<Int2ObjectHashMap<MessageStream>>> messageStreams =
```
tmontgomery: check into BiInt2ObjectMap.

aeron-tools/src/main/java/uk/co/real_logic/aeron/tools/ThwackerTool.java
```diff
@@ -0,0 +1,933 @@
...
+        private int lastSessionId;
+        private long lastThreadId;
+
+
+        ThwackingElement(final String chan, final int stId, final boolean verifiable, final boolean createSubscriber)
+        {
+            channel = chan;
+            streamId = stId;
+            verify = verifiable;
+            buffer = new ThreadLocal<UnsafeBuffer>()
```
tmontgomery: There is a new Java8 technique for this now. ThreadLocal.withInitial(). Might be a little bit more readable.

aeron-tools/src/main/java/uk/co/real_logic/aeron/tools/log_analysis/LogAnalyzer.java
```diff
@@ -0,0 +1,196 @@
...
+import javafx.scene.paint.Color;
+import javafx.stage.DirectoryChooser;
+import javafx.stage.Stage;
+
+import javax.swing.JPanel;
+import javax.swing.JTextField;
+
+public class LogAnalyzer extends Application implements Observer
+{
+    private MenuBar menuBar = null;
```
tmontgomery: null initializers not need.

aeron-tools/src/main/java/uk/co/real_logic/aeron/tools/log_analysis/LogModel.java
```diff
@@ -0,0 +1,210 @@
...
+            final LogBuffers logBuffers = new LogBuffers(path);
+            final UnsafeBuffer[] atomicBuffers = logBuffers.atomicBuffers();
+            final DataHeaderFlyweight dataHeaderFlyweight = new DataHeaderFlyweight();
+            final UnsafeBuffer logMetaDataBuffer = atomicBuffers[PARTITION_COUNT * 2];
+            termLength = atomicBuffers[0].capacity();
+            System.out.format("Initial term id: %d\n", initialTermId(logMetaDataBuffer));
+            System.out.format(" Active term id: %d\n", activeTermId(logMetaDataBuffer));
+            System.out.format("   Active Index: %d\n", indexByTerm(initialTermId(logMetaDataBuffer),
+                    activeTermId(logMetaDataBuffer)));
+            System.out.format("    Term Length: %d\n", termLength);
```
tmontgomery: might want to add MTU length as it has been added to the log meta data

aeron-tools/src/main/java/uk/co/real_logic/aeron/tools/log_analysis/LogModel.java
```diff
@@ -0,0 +1,210 @@
+/*
```
tmontgomery: how is this different than LogInspector?

aeron-tools/src/main/java/uk/co/real_logic/aeron/tools/perf_tools/AeronPing.java
```diff
@@ -0,0 +1,378 @@
...
+import uk.co.real_logic.aeron.Publication;
+import uk.co.real_logic.aeron.Subscription;
+import uk.co.real_logic.aeron.common.concurrent.logbuffer.BufferClaim;
+import uk.co.real_logic.aeron.common.concurrent.logbuffer.Header;
+import uk.co.real_logic.agrona.DirectBuffer;
+import uk.co.real_logic.agrona.MutableDirectBuffer;
+import uk.co.real_logic.agrona.concurrent.UnsafeBuffer;
+
+/**
+ * Created by philip on 4/7/15.
```
tmontgomery: Probably can remove these style comments since git works well to know this. :)

aeron-tools/src/main/java/uk/co/real_logic/aeron/tools/perf_tools/AeronPing.java
```diff
@@ -0,0 +1,378 @@
+/*
```
tmontgomery: What does this and Pong do differently than the samples Ping and Pong?

Also, why does this do its own calculations instead of using HdrHistogram?

build.gradle
```diff
@@ -195,6 +206,24 @@ project(':aeron-samples') {
     build.doLast {
         tasks.shadowJar.execute()
     }
+
```
tmontgomery: I think these tasks clutter up the build. It's not too bad to execute java at the command line and bring in a tools.jar instead. Somewhat like we do with samples.

aeron-samples/src/main/java/uk/co/real_logic/aeron/samples/MultipleSubscribersWithFragmentAssembly.java
vksingh70: Agree, Will fix that

aeron-tools/src/main/java/uk/co/real_logic/aeron/tools/perf_tools/AeronPing.java
johnpr01: This version is more for the performance testing framework. I didn't want to clutter up the sample version (which I think most people would use). This version is run after builds and then the data is fed through gnuplot to generate our latency graphs per build. I will clean it up and change a couple of method names since I've started using gnuplot instead of having the tool generate scatterplots itself.

aeron-tools/src/main/java/uk/co/real_logic/aeron/tools/log_analysis/LogModel.java
johnpr01: Correct. This will be removed. I was unable to get to it prior to the pull request being made.

aeron-samples/src/main/java/uk/co/real_logic/aeron/samples/SimplePublisher.java
vksingh70: This shows that we can do it without AutoClosable. 
Also, show the explicit need to close publication and aeron. There may be a case where we wanted to store publication in a variable beyond the try block and eventually need to be closed.

aeron-samples/src/main/java/uk/co/real_logic/aeron/samples/MultipleSubscribersWithFragmentAssembly.java
vksingh70: Sure

aeron-samples/src/main/java/uk/co/real_logic/aeron/samples/MultipleSubscribersWithFragmentAssembly.java
vksingh70: Sure

aeron-tools/src/main/java/uk/co/real_logic/aeron/tools/TLRandom.java
strangelydim: Java already has a ThreadLocalRandom, spelled out.  It was named differently to distinguish it from that.  TLRandom exists to allow the random seed to be set or gotten via a callback, which for unknown idiotic reasons, Java's built-in ThreadLocalRandom doesn't allow.  Do you have a suggestion for a better name that isn't too long or silly?  CoolAeronToolsThreadLocalRandom ?

aeron-tools/src/main/java/uk/co/real_logic/aeron/tools/MediaDriverOptions.java
strangelydim: In the future, we'll be sure to run them by you for anything we're planning on doing a PR for; how about for now?  Can we agree on using Apache's CommandLine now?  I don't think there's a license incompatibility.  The alternative is for us to write and include our own getopt parsing class, which seems a little silly.

aeron-driver/src/main/java/uk/co/real_logic/aeron/driver/DriverConductor.java
strangelydim: A lot of the import re-organization was done automagically by Eclipse upon saving the files, so the style we used is whatever the default style Eclipse thinks is good.  If you guys don't like the default Eclipse style, we can change it.  I assume it's the Eclipse default because it's some kind of standard practice, though.

aeron-samples/src/main/java/uk/co/real_logic/aeron/samples/SimplePublisher.java
strangelydim: Yeah, we had this as a try-with-resources originally, and changed it just to make the code much more explicit and give us a place to comment on cleaning up resources.  So not using try-with-resources was very much on purpose.

aeron-tools/src/main/java/uk/co/real_logic/aeron/tools/SubscriberTool.java
strangelydim: Iiiinteresting.  Didn't know that existed, but it looks like what I was wanting.

build.gradle
strangelydim: I think Phil was meaning to remove those for the PR and just hadn't gotten to them.

aeron-tools/src/main/java/uk/co/real_logic/aeron/tools/SubscriberTool.java
strangelydim: Hm, maybe not.  It does look useful, but I wanted the ability to distinguish between having an entry with a given stream ID, but no session ID yet, and not having the stream ID at all... So I think I still need the 3-step lookup.  The full lookup - starting with the channel String - doesn't happen except during setup, so it shouldn't be too expensive.

aeron-samples/src/main/java/uk/co/real_logic/aeron/samples/AeronStat.java
strangelydim: Yep, we really do.

aeron-tools/src/main/java/uk/co/real_logic/aeron/tools/MessageStream.java
strangelydim: I'll fix that in the files in the PR - but a grep through the main Aeron source shows missing newlines all over the place, too.

strangelydim: As I'm going through and adding newlines, it's looking really ridiculous doing it for one-line docs like this:

```
/** Application should print advanced usage guide with help */
boolean showUsage;
```

Do you really want those to look like this instead?:

```
/**
 * Application should print advanced usage guide with help */
boolean showUsage;
```

aeron-tools/src/main/java/uk/co/real_logic/aeron/tools/ChannelDescriptor.java
brianhorst: I can do this, but I'll dearly miss typing .get and having intellisense show me all my options.

build.gradle
johnpr01: I have already moved all of this into the build.gradle in aeron-tools. I also create a separate tools.jar which packages everything together just like samples.jar.

aeron-tools/src/main/java/uk/co/real_logic/aeron/tools/TLRandom.java
mjpt777: Can you give it a name that signifies why it is different?

aeron-tools/src/main/java/uk/co/real_logic/aeron/tools/perf_tools/AeronPing.java
```diff
@@ -0,0 +1,371 @@
...
+            }
+            out.close();
+        }
+        catch (final Exception e)
+        {
+            e.printStackTrace();
+        }
+    }
+
+    public void onNewConnection(final String channel, final int streamId,
```
mjpt777: Align args

aeron-tools/src/main/java/uk/co/real_logic/aeron/tools/perf_tools/AeronPong.java
```diff
@@ -0,0 +1,171 @@
...
+            return;
+        }
+        if (pongPub.tryClaim(length, bufferClaim) >= 0)
+        {
+            try
+            {
+                final MutableDirectBuffer newBuffer = bufferClaim.buffer();
+                newBuffer.putBytes(bufferClaim.offset(), buffer, offset, length);
+            }
+            catch (final Exception e)
```
mjpt777: What is the exception you are expecting to catch?

aeron-tools/src/main/java/uk/co/real_logic/aeron/tools/perf_tools/AeronThroughputencyPublisher.java
```diff
@@ -0,0 +1,434 @@
...
+import uk.co.real_logic.agrona.DirectBuffer;
+import uk.co.real_logic.agrona.MutableDirectBuffer;
+import uk.co.real_logic.agrona.concurrent.BusySpinIdleStrategy;
+import uk.co.real_logic.agrona.concurrent.IdleStrategy;
+import uk.co.real_logic.agrona.concurrent.UnsafeBuffer;
+
+/**
+ * Created by philipjohnson1 on 4/2/15.
+ */
+public class AeronThroughputencyPublisher implements RateController.Callback
```
mjpt777: What is this name? :-)

aeron-tools/src/main/java/uk/co/real_logic/aeron/tools/perf_tools/AeronThroughputencyPublisher.java
```diff
@@ -0,0 +1,434 @@
...
+
+        try
+        {
+            subThread.join();
+        }
+        catch (final Exception e)
+        {
+            e.printStackTrace();
+        }
+        pub.close();
```
mjpt777: Why not try with resources?

aeron-tools/src/main/java/uk/co/real_logic/aeron/tools/perf_tools/AeronThroughputencyPublisher.java
```diff
@@ -0,0 +1,434 @@
...
+        //g2.setColor(Color.orange);
+        //plotSubset(g2, start, end, "3M msgs/sec", 100 + width * 6, width, stepY, means[6]);
+
+        try
+        {
+            ImageIO.write(image, "png", imageFile);
+        }
+        catch (final Exception e)
+        {
+            e.printStackTrace();
```
mjpt777: Why not re-throw and have top level handler. Don't like local handling that outputs without control.

aeron-tools/src/main/java/uk/co/real_logic/aeron/tools/perf_tools/AeronThroughputencySubscriber.java
```diff
@@ -0,0 +1,141 @@
...
+            {
+                System.out.println("Took too many tries: " + iterations);
+            }
+            try
+            {
+                final MutableDirectBuffer newBuffer = bufferClaim.buffer();
+                final int newOffset = bufferClaim.offset();
+                 newBuffer.putBytes(newOffset, buffer, offset, length);
+            }
+            catch (final Exception e)
```
mjpt777: Again what exception are you expecting?

aeron-tools/src/main/java/uk/co/real_logic/aeron/tools/TLRandom.java
strangelydim: How about "SeedableThreadLocalRandom"?

aeron-tools/src/main/java/uk/co/real_logic/aeron/tools/MessageStream.java
mjpt777: For one liners it is OK. If 2 or more then top and tail. IntelliJ will do this for you.

aeron-tools/src/main/java/uk/co/real_logic/aeron/tools/TLRandom.java
mjpt777: Better. It is starting to say what it is.

aeron-samples/src/main/java/uk/co/real_logic/aeron/samples/SimplePublisher.java
mjpt777: Sign of too many silly comments :-)

aeron-tools/src/main/java/uk/co/real_logic/aeron/tools/SubscriberTool.java
mjpt777: When you ref another class then use the JavaDoc @link tag.

aeron-driver/src/main/java/uk/co/real_logic/aeron/driver/LossDetector.java
mjpt777: Definitely turn things off at runtime with public final static fields.

aeron-tools/src/main/java/uk/co/real_logic/aeron/tools/MediaDriverOptions.java
mjpt777: use "new" rather than "make" in the method name when creating a new object.

aeron-tools/src/main/java/uk/co/real_logic/aeron/tools/perf_tools/AeronThroughputencyPublisher.java
mjpt777: I think it is an anti-pattern to put an author name on a file. Collective code ownership. If you want to know who actually made a change then that is what version control is for.

aeron-tools/src/test/java/uk/co/real_logic/aeron/tools/MessageSizePatternTest.java
```diff
@@ -0,0 +1,173 @@
...
+ */
+public class MessageSizePatternTest
+{
+    MessageSizePattern p;
+    @Test
+    public void createWithMessageSize() throws Exception
+    {
+        p = new MessageSizePattern(1000);
+        assertThat("FAIL: getNext() didn't return correct value",
+                p.getNext(), is(1000));
```
mjpt777: Hanging indent of 4.

aeron-tools/src/test/java/uk/co/real_logic/aeron/tools/MessageStreamTest.java
```diff
@@ -0,0 +1,184 @@
...
+import static org.hamcrest.core.Is.is;
+
+import java.nio.ByteOrder;
+
+import org.junit.Test;
+
+import uk.co.real_logic.agrona.concurrent.UnsafeBuffer;
+
+/**
+ * Created by ericb on 3/27/15.
```
mjpt777: No authors in code.

aeron-tools/src/test/java/uk/co/real_logic/aeron/tools/MessagesAtMessagesPerSecondTest.java
```diff
@@ -0,0 +1,81 @@
...
+    class Callback implements RateController.Callback
+    {
+        public int onNext()
+        {
+            return 0;
+        }
+
+    }
+
+    Callback callback = new Callback();
```
mjpt777: Use a lambda expression.

aeron-tools/src/main/java/uk/co/real_logic/aeron/tools/MediaDriverOptions.java
mjpt777: We can go with Apache CommandLine for now. 

Do you really need to parse that many options to pull in a lib?

aeron-tools/src/main/java/uk/co/real_logic/aeron/tools/perf_tools/AeronThroughputencyPublisher.java
johnpr01: Throughputency refers to a type of test. It is a combination of Throughput and Latency, sometimes also called a Latency Under Load test. We defined this term a few years ago while still 29W.

mjpt777: Why not call it LatencyUnderLoadTest then? I like things to be more obvious.

aeron-tools/src/main/java/uk/co/real_logic/aeron/tools/SubscriberTool.java
strangelydim: Not sure what you mean - I should be putting a `@link` tag somewhere in this comment:

```
/* channel -> stream ID -> session ID */
```

?

aeron-driver/src/main/java/uk/co/real_logic/aeron/driver/Configuration.java
strangelydim: Elsewhere Martin commented that it's fine to have everything on one line if it's just a one-line comment.  So... ?

aeron-tools/src/main/java/uk/co/real_logic/aeron/tools/MediaDriverOptions.java
strangelydim: Yep.

aeron-tools/src/main/java/uk/co/real_logic/aeron/tools/SubscriberTool.java
mjpt777: Use things like `{@link Receiver}` in javadoc so the class is linked and refactoring works. Don't just use a raw class name.

strangelydim: Oh, gotcha.  Todd's original comment here was for a line a few lines down.

strangelydim: Should I use @link even in comments that are normal, non-Javadoc-style code comments like this one?  Will those still get found and changed during a refactor?  Should I convert normal comments to /** javadoc-style comments?  I would expect not, since I don't think we'd want to clutter up the actual generated doc like that.

mjpt777: I'd ask a different question. Why do you need comments on normal fields? It is always a smell to me when things need comments. Over time the code changes and the comment become misleading and even plain wrong.

If the code is so unclear then how will a comment help? If the person cannot write clear enough code that it requires comments then what makes you think the comments are going to be better than then code?

brianhorst: I think a comment here is worth having. what makes the code unclear is the list of generic types. Saying what's actually used to access the per producer object is very helpful. It may be more clear if we created a new class just to hold and access that information, but that might not be worth it. We thought for a second one existed in BiInt2ObjectMap, but that wasn't the case.

As for javadoc and refactoring, it won't work without being in the standard /** form (at least it doesn't in intellij). This isn't in javadoc and isn't specifically identifying any classes so there shouldn't be any problems there.

As an additional side note, you don't need to repeat types when creating the new hash map. I.e. it could just be "messageStreams = new HashMap();".

brianhorst: I also noticed subscriber tool has some “incorrect” whitespace formatting. Want me to clean that up?

strangelydim: Go for it.

Issue #107, opened 2015-04-30
Publication IndexOutOfBoundsException when calling offer() on different threads with random message sizes


mst4:
In running the ThwackerTool on Windows, my Publication will spit out IndexOutOfBoundsExceptions when I run with more than one sending thread (sending on the same Publication) using random message sizes.

```
java -cp aeron-tools\build\libs\tools.jar uk.co.real_logic.aeron.tools.ThwackerTool -v yes -d 10000 -e 10 -s 2 --max-size 100 --min-size 64
```

This will cause it to run verifiable streams for 10 seconds with 10 publications, 10 subscriptions, 2 sending threads, and random messages of size 64 to 100 bytes.

It reports multiple of these exceptions:

```
java.lang.IndexOutOfBoundsException: index=65520, length=24, capacity=65536
	at uk.co.real_logic.agrona.concurrent.UnsafeBuffer.boundsCheck(UnsafeBuffer.java:811)
	at uk.co.real_logic.agrona.concurrent.UnsafeBuffer.putBytes(UnsafeBuffer.java:740)
	at uk.co.real_logic.aeron.common.concurrent.logbuffer.LogAppender.appendPaddingFrame(LogAppender.java:306)
	at uk.co.real_logic.aeron.common.concurrent.logbuffer.LogAppender.appendUnfragmentedMessage(LogAppender.java:218)
	at uk.co.real_logic.aeron.common.concurrent.logbuffer.LogAppender.append(LogAppender.java:144)
	at uk.co.real_logic.aeron.Publication.offer(Publication.java:192)
	at uk.co.real_logic.aeron.tools.ThwackerTool$ThwackingElement.trySendMessage(ThwackerTool.java:884)
	at uk.co.real_logic.aeron.tools.ThwackerTool.sendOnRandomPub(ThwackerTool.java:469)
	at uk.co.real_logic.aeron.tools.ThwackerTool.lambda$createAndStartThreads$14(ThwackerTool.java:299)
	at uk.co.real_logic.aeron.tools.ThwackerTool$$Lambda$63/1068824137.run(Unknown Source)
	at java.lang.Thread.run(Thread.java:745)
```

It always seems to be at the end of the TermBuffer length dealing with adding padding.  The ThwackerTool uses a TermBuffer length of 65536.  It will only happen with multiple sending threads and random message sizes, as using just one sending thread with random message sizes or 2 sending threads with uniform message size will not cause any issues.

[...]

PR #114, opened 2015-05-06
Fixed issue #107 by ensuring there's enough space to write a padding …

strangelydim:
…frame at the end of the term buffer if padding is needed.


mjpt777:
Nice spot Eric! Thanks.

PR #123, opened 2015-05-12
Fixed bug where a random InputStream was always used instead of possi…

strangelydim:
…bly a given InputStream.


strangelydim:
A nice small PR this time!

PR #125, opened 2015-05-14
Randomize control channel for SubscriberTool

strangelydim:
This allows the SubscriberTool to play a bit more nicely with other SubscriberTool instances on the same machine.  And also serves to bring the commit history into sync between real-logic/master and strangelydim/master.


strangelydim:
Hold off on merging this for just a second... we're fixing one more bug.

strangelydim:
OK, this is good to merge now.

PR #134, opened 2015-05-20
Perf fix

johnpr01:
I have a number of very small fixes so that we can run perf tests as part of our Jenkins CI setup. Perf tests are now run on actual machines instead of a VM using loopback.


tmontgomery:
Update the PR and I can merge it for updating the tools.

Issue #141, opened 2015-06-01
Stopping and then restarting a subscription causes it to never receive data again


jessefugitt:
Using the latest code from GitHub, I've been testing a stop and restart subscription scenario where both subscriptions are on the same channel but different streams.  On Machine 1, I have a media driver app, a publication app on stream 1, and a publication app on stream 2.  On Machine 2, I have a media driver app, a subscription app on stream 1, and a subscription app on stream 2.  When I start everything up, the two subscription apps are receiving data:

NEW CONNECTION: channel "aeron:udp?remote=192.168.14.14:31111", stream 1, session 1496040322
1.000034: 334.554 Kmsgs/sec 85.6 Mbps
0.999995: 1.049 Mmsgs/sec 268.5 Mbps
1.000004: 1.059 Mmsgs/sec 271.1 Mbps
1.000001: 1.017 Mmsgs/sec 260.3 Mbps
...

NEW CONNECTION: channel "aeron:udp?remote=192.168.14.14:31111", stream 2, session -578364638
1.000047: 1.671 Mmsgs/sec 427.8 Mbps
0.999997: 1.576 Mmsgs/sec 403.4 Mbps
1.000010: 1.054 Mmsgs/sec 269.7 Mbps
1.000009: 1.040 Mmsgs/sec 266.4 Mbps
...

Then, I stop and restart the subscription app on stream 2.  It gets a new connection message but it never receives data:

NEW CONNECTION: channel "aeron:udp?remote=192.168.14.14:31111", stream 2, session -578364638
1.000362: 0.000 msgs/sec 0 bps
0.999624: 0.000 msgs/sec 0 bps
1.000001: 0.000 msgs/sec 0 bps
1.000000: 0.000 msgs/sec 0 bps
...

The publication app seems to be sending a small bit of data every 10 seconds or so but it never gets delivered to the subscription app:

0.999975: 1.036 Mmsgs/sec 265.1 Mbps
0.999999: 1.090 Mmsgs/sec 279.1 Mbps
1.000002: 1.085 Mmsgs/sec 277.9 Mbps
1.000000: 231.696 Kmsgs/sec 59.3 Mbps
1.000001: 0.000 msgs/sec 0 bps
0.999999: 0.000 msgs/sec 0 bps
1.000008: 0.000 msgs/sec 0 bps
0.999992: 2.340 Kmsgs/sec 599.0 Kbps
1.000002: 0.000 msgs/sec 0 bps
[...]

PR #143, opened 2015-06-01
Issue 141 System Test Pull Request

jessefugitt:
Added a new system test class that demonstrates the bug described in issue #141 and causes one of the included tests to fail (obviously causing the build to break as well).


jessefugitt:
Closing this pull request because Aeron had breaking changes to the API and this system test will be submitted in another pull request.

Issue #127, opened 2015-05-14
 When a subscriber is started after the publisher (and it is not the first subscriber), it will not report inactive connection when the publisher exits


antrongladney:
Minor issue.
Summary:
When a subscriber is started after the publisher (and it is not the first subscriber), it will not report inactive connection when the publisher exits
when the publisher exits.

All apps running on same machine.
Steps to reproduce:
1. Start MediaDriverTool using the following command:
java -cp tools.jar -Daeron.dir=/dev/shm/ag/aeron uk.co.real_logic.aeron.tools.MediaDriverTool

2. Start the PublisherTool using the following command:
java -cp tools.jar -Daeron.dir=/dev/shm/ag/aeron uk.co.real_logic.aeron.tools.PublisherTool -c 'aeron:udp?address=192.168.14.34|group=224.17.17.13:32511#5' -m 100000 -r 1000m@5mps

3. Wait for a few seconds and start 3 Subscribers using the following command on the same machine as the Publisher and the MediaDriver
java -cp tools.jar -Daeron.dir=/dev/shm/ag/aeron uk.co.real_logic.aeron.tools.SubscriberTool -c 'aeron:udp?address=192.168.14.34|group=224.17.17.13:32511#5'

Results:
All subscribers will receive messages from the Publisher.
When the publisher completes the sending of all messages, it will exit.
The first subscriber will report the following

INACTIVE CONNECTION: channel "aeron:udp?address=192.168.14.31|group=224.17.17.13:32511", stream 5, session -6695520

Subscribers 2 and 3 do not report the inactive connection as expected.

tmontgomery:
Can you retest with the latest set of changes? Let us know if it is still happens.

brianhorst:
I ran this scenario and I still see the unusual behavior.

One additional thing I noticed is the Subscribers all report the inactive notification, as expected, if they are running _before_ the publisher starts. There must be something on that initial wire up that causes some to be left out.

I'm going to attempt to write a system test to reproduce this.

PR #146, opened 2015-06-09
Issue #127 fix and unit test

brianhorst:
These changes fix and test issue 127. In order to unit test the change I needed to make DriverConductor able to use mock network connections. Creating real network connections made it impossible to test in isolation.


aeron-driver/src/main/java/uk/co/real_logic/aeron/driver/DriverConductor.java
```diff
@@ -367,6 +368,45 @@ public void onCloseResource(final AutoCloseable resource)
             .collect(toList());
     }
 
+    NetworkConnection newNetworkConnection(
```
tmontgomery: Why is this necessary for the unit test?

tmontgomery:
I've incorporated this fix with a new unit test. So, closing the PR.

PR #147, opened 2015-06-09
Fix issue 141 and add system test

jessefugitt:
This pull request contains a proposed fix for issue #141 by marking a connection inactive on the driver conductor thread and then preventing new subscriptions from being associated with any network connections that are not active.

However, there may be a cleaner way to fix this issue which would involve taking a close look at the which of the threads in the driver should really be updating the network connections in various places and addressing some of those windows that currently exist as a connection moves from active to inactive to linger (especially as it relates to processing add/remove events in parallel on the driver conductor thread).

Also, should the following line in DataPacketDispatcher's removeSubscription method still be there after applying this fix?
```
connectionBySessionIdMap.values().forEach(NetworkConnection::ifActiveGoInactive);
```


aeron-driver/src/main/java/uk/co/real_logic/aeron/driver/DriverConductor.java
```diff
@@ -683,6 +684,11 @@ private void onRemoveSubscription(final long registrationId, final long correlat
         final int refCount = channelEndpoint.decRefToStream(subscription.streamId());
         if (0 == refCount)
         {
+            connections
```
tmontgomery: When the Receiver does a removeSubscription from the dispatcher, I think it does this check already. Why do it here?

jessefugitt: It wasn't clear which thread should own the ability to write state to the NetworkConnection objects (currently multiple threads appear to).

This code could be removed from the Receiver class and just moved here to the DriverConductor but I'm not sure if other calls expect it to also happen in the Receiver code (like in response to network events).  As reported in issue #141, if it isn't done here, it triggers the bug described in the issue because the "work" will be put on a queue for another thread to do later, which opens a window where new "add subscriber" messages from the client will temporarily get attached to this NetworkConnection before it gets marked inactive on the other thread and then they never recover or receive messages.

tmontgomery: This a little bit weird. But [here](https://github.com/real-logic/Aeron/blob/master/aeron-driver/src/main/java/uk/co/real_logic/aeron/driver/NetworkConnection.java#L264) is where a little is explained. Having the conductor set the state for this creates some cache traffic on the Receiver thread.

PR #168, opened 2015-09-20
Start on port of media driver to C++

mikeb01:
I've started porting the media driver from Java to C++.  It is far from complete, but is a initial pass at a number of classes from the Java code base.  I've started with the AeronUri and UdpChannel and associated classes.  It handles uri parsing, interface lookup and has some simple abstractions in place to handle the differences between IPv4 and IPv6.

Includes the build infrastructure, e.g. CMakeLists.txt for the main and test code for the media driver.  I've introduced a dependency from the driver to the client code, which is consistent with the direction of the dependency in the Java code (I think).  This is to reuse the util code in the client code base.

### DIFF
--- a/aeron-driver/src/main/java/uk/co/real_logic/aeron/driver/event/EventDissector.java
+++ b/aeron-driver/src/main/java/uk/co/real_logic/aeron/driver/event/EventDissector.java
@@ -22,28 +22,26 @@ import uk.co.real_logic.agrona.MutableDirectBuffer;
 
 import java.net.InetAddress;
 
-import static java.lang.ThreadLocal.withInitial;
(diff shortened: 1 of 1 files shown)


mikeb01:
Don't take this pull request yet.  Doesn't compile on Linux or with GCC.  Fixing.

mikeb01:
The update should allow it to compile on Linux with GCC (#169 no withstanding).

tmontgomery:
Awesome stuff!

I will not be able to review until tomorrow afternoon. But will do so as soon as I can and merge then.

mikeb01:
Closing this as I'm moving stuff about in my repo, will reopen if desired.

PR #205, opened 2016-03-01
fix formatting of bytes to ascii in LogInspector. Negative bytes to be formatted to zero

langera:
That seemed to me to be the intention in the previous code, however, the widening of the byte to an integer made sure that 'b' would never be negative and therefore the condition was redundant.


mjpt777:
Thanks

PR #228, opened 2016-05-30
Bump Aeron version for cmake

vdidenko:
CMake to be aware of proper Aeron version

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -28,8 +28,8 @@ project("aeron")
 
 # version info
 set(aeron_VERSION_MAJOR 0)
-set(aeron_VERSION_MINOR 1)
-set(aeron_VERSION_PATCH 0)
+set(aeron_VERSION_MINOR 9)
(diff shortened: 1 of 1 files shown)


vdidenko:
I am just following the existing pattern in CMakeLists.txt. The better CMake way would be:

    project("aeron" VERSION 0.9.8)

In the [project definition line](https://github.com/real-logic/Aeron/blob/7912588769ec0a92491aa17119d73e2b752df3de/CMakeLists.txt#L27), per [CMake docs](https://cmake.org/cmake/help/v3.0/command/project.html)

tmontgomery:
@vdidenko how would `version.txt` be pulled in for this?

If that can work well, then I would be all for bumping to CMake 3.0.2 and adding the version that way.

vdidenko:
OK, I have it [in another branch](https://github.com/vdidenko/Aeron/tree/project_version_in_cmake), will make that a pull request in a sec.

PR #259, opened 2016-07-14
[C++] Added setting clientId on remove messages to the driver

lirm:
It appears that clientId is not set on the flyweights for removeSubscription and removePublication sent to the driver. Doesn't seem to impact the functionality, however.


mjpt777:
The remove code does not use these fields though they are probably useful for debugging, e.g. if an exception occurs in the driver then the log entry will have the client id.

I noticed the Java takes a different approach and can reuse a buffer with them already set. Give me a good opportunity to simplify things a little on the Java side as I'd not looked at it in a while.

https://github.com/real-logic/Aeron/commit/f07738b2a6edacaf9b090e3dbeda24db2bae6313

PR #265, opened 2016-08-07
Add ThreadFactories for each thread in MediaDriver to support program…

kdkavanagh:
Adds configurations to MediaDriver Context to pass in ThreadFactories for each of the threads spun up by the MediaDriver.  Will allow users to easily create ThreadFactories that pin indivudal MediaDriver threads to cores.  (Dependent on Agrona PR https://github.com/real-logic/Agrona/pull/69)


mjpt777:
How about adding the thread factories to the context like you suggest and then pass them appropriately to startOnThread rather than hold a reference in the AgentRunner?

tmontgomery:
I agree with @mjpt777 .

kdkavanagh:
No longer dependent on any changes to Agrona

mjpt777:
Thanks!

PR #267, opened 2016-08-16
SubscriptionLink's clientLivenessTimeoutNs from MediaDriver context i…

grzegorz-gierlach:
…nstead of directly from Configuration (and System properties)


mjpt777:
Thanks

PR #279, opened 2016-09-15
aeron-agent: improve logging detachment; allow to enable/disable logging at runtime via MX bean

raphw:
The Aeron agent now reads its argument and registers an MX bean to the running platform's MBean server if `mbean` is contained in the agent arguments. Using this exposed MX bean, it is now possible to enable or disable logging during runtime and to check the current logging state of a running application.

### DIFF
--- a/aeron-agent/src/main/java/io/aeron/agent/EventLogAgent.java
+++ b/aeron-agent/src/main/java/io/aeron/agent/EventLogAgent.java
@@ -18,6 +18,7 @@ package io.aeron.agent;
 import io.aeron.driver.EventLog;
 import net.bytebuddy.ByteBuddy;
 import net.bytebuddy.agent.builder.AgentBuilder;
+import net.bytebuddy.agent.builder.ResettableClassFileTransformer;
 import net.bytebuddy.description.type.TypeDescription;
 import net.bytebuddy.dynamic.DynamicType;
 import net.bytebuddy.dynamic.scaffold.TypeValidation;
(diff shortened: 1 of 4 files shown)


RichardWarburton:
It is worth noting that communicating over MXBeans has always been garbagey for me in the past, and their default transport layer is over RMI which periodically forces a full gc. Is this really the right means of control in an otherwise steady-state allocation free system? I'm not really up to date with the logging side of things, so feel free to disregard me if you've already addressed this issue in previous discussion ;) just thought I would raise the question.

mjpt777:
Yes it is nice that this can be turned on automatically but enabling RMI MXBeans is not a good option.

raphw:
Right, did not think of that.

An alternative would be to apply the reset on agent attachment when some `reset` parameter is set. Any agent can be attached multiple times.

mjpt777:
Would it be possible to have an example of resetting with an agent attachment rather than MXBeans?

The system can be tested by starting the driver with [agent](https://github.com/real-logic/Aeron/blob/master/aeron-samples/scripts/logging-media-driver) script and a test [publisher](https://github.com/real-logic/Aeron/blob/master/aeron-samples/scripts/basic-publisher) and [subscriber](https://github.com/real-logic/Aeron/blob/master/aeron-samples/scripts/basic-subscriber) started after.

raphw:
It is possible to runtime attach an agent multiple times. Depending on the parameter, the agent could choose to rollback a previous transformation. 

The question is however if this is an actual use case. Alternatively, the resetting mechanism could be dropped altogether. I added it because it was easy to do but did not think of the performance impact of enabling JMX.

mjpt777:
Thanks for the ideas. I'll close this one for now. We do not switch logging at runtime. I normally use BTrace for dynamically connecting to a running system when necessary.

raphw:
Sure, entirely forgot about this. Makes perfect sense.

PR #288, opened 2016-11-06
Make artifacts valid OSGi bundles.

oheger:
I would like to use the artifacts of this project in an OSGi environment. (Actually, they come in as transitive dependencies of the most recent akka-remote version.) So far, the produced jars do not contain any OSGi meta data.

This patch is pretty trivial. It just adds the Gradle OSGi plugin to the relevant projects. This produces standard OSGi meta data which is sufficient to use the jars as bundles in an OSGi container; I could verify that they could be deployed and used successfully in a local project.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -170,6 +170,8 @@ subprojects {
 }
 
 project(':aeron-client') {
+    apply plugin: 'osgi'
+
     dependencies {
         compile 'org.agrona:Agrona:0.5.6'
     }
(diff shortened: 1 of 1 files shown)


mjpt777:
Thanks. I've limited knowledge of OSGi. I understand this replaces the default manifest. What are the implications of this?

oheger:
Many thanks :-)

PR #296, opened 2016-12-11
[C++]: Fixed segfault due to null ifa_addr from getifaddrs

jh0x:
`BsdInterfaceLookup::lookupIPv4` and `BsdInterfaceLookup::lookupIPv6` can segfault on a machine with interfaces that have `NULL` `ifa_addr` (e.g. OpenVPN `tun` interface).

From the documentation of `getifaddrs`:

> `ifa_addr` and other fields can actually be `NULL` if the interface has no address, and no link-level address is returned if the interface has an IP address assigned.


mikeb01:
Thank you for the fix.

PR #301, opened 2017-01-03
enable Aeron on non-x86-64 platforms with GCC and C++11 functions

erichocean:
I have tested this with the aeron-client tests using ctest and all tests pass. I have not done any heavy load testing, concurrency testing, etc.


erichocean:
It's possible I was too conservative with the memory ordering in places, especially the `cmpxchg` functions.

PR #317, opened 2017-02-18
Made finding the next power of 2 more portable.

nicklauslittle:
The __lzcnt instruction is not returning the correct value. (Apparently, when compiling with VS it only works on newer CPUs.) I modified the find next power of 2 function so that it doesn't rely on this function. I also removed the count leading zeros function, since it will generate incorrect information on some CPUs. (Although, it could be left in or modified.) With this change all the C++ tests are passing on Windows.

### DIFF
--- a/aeron-client/src/main/cpp/util/BitUtil.h
+++ b/aeron-client/src/main/cpp/util/BitUtil.h
@@ -72,6 +72,10 @@ namespace BitUtil
         return current - 1;
     }
 
+    /* Counts the leading number of zeros in a value.
+     * (Note: this only works on 32-bit types, when compiling with GCC
+     * or on newer x64 processors when compiling with Visual Studio.)
+     */
     template<typename value_t>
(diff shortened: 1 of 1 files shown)


mh321:
Wouldn't it make more sense to use the builtin whenever possible and fallback to the slow version in case of not supported architectures/compilers?

nicklauslittle:
Normally, yes. But in this case I would say no, simply because we are looking for the next power of 2 rather than the actual leading zero count. By the time compiler optimizations are done (loop-unrolling, etc.), the part that replaces the leading zero count boils down to simple bitwise OR's and shifts, which are extremely fast. Even though the LZCNT instruction would be faster, I would think the overhead of checking for support would make it slower than the generic version.

If actually counting the leading zeros is needed, then yes, it would probably be beneficial to make the appropriate optimizations. I am going to go ahead and add the function back in, but will probably add a comment that it only works on 32-bit types using gcc and on some processors using Visual Studio. (i.e. It's not what I would consider a robust function.)

PR #319, opened 2017-02-25
[C++] Improvements to file operations on Windows

nicklauslittle:
This allows files larger than 4GB to be mapped on Windows, and fixes a couple warnings that Visual Studio 2015 identifies.


mjpt777:
The driver will not map files larger than 1GB due to the design of the protocol and restrictions in Java.

nicklauslittle:
Okay. I was mainly going through fixing the warnings in Visual Studio and noticed only the parameters for the lower 32 bits were used. Would it be better to just fix the warnings and leave the existing code as is?

mjpt777:
I'd just focus on the warnings rather than changing behaviour. @tmontgomery what do you thing?

nicklauslittle:
Alright. I pushed a new commit that doesn't have the changed functionality, just the warning fixes.

PR #364, opened 2017-06-06
Refactored the Function<T,Integer> to ToIntFunction<T>

Harshrajsinh:
Hello,

I am a graduate student at Oregon State University and as a part of my research and subject CS562 Applied Software Engineering project (Study and Refactoring of Functional Interface in Java), 
I have done refactoring of the Function<T,Integer> to ToIntFunction<T>.  
My project deals with the boxing and unboxing of Wrapper class and the primitive data-types. 
I want to know that whether the open source community is ready to accept these micro-optimizing refactorings or not. 

In the file ClientConductorTest, the Function<MutableDirectBuffer,Integer> is consumed as primitive int in the line **final int length = filler.apply(buffer)**.  So, Java Compiler has to autobox the Integer to int and because of that, the performance reduces. So, instead of using Function<T,Integer>, Java 8 has provided a specialized primitive functional interface to ToIntFunction.

Thank you,
Harsh Thakor


mjpt777:
Happy to merge. These things don't matter so much in tests. Best if in hot path of code.

PR #390, opened 2017-08-25
Corrected reading of the 3 boolean JVM options.

mogrifier:
The file io.aeron.samples.SampleConfiguration was not reading boolean JVM options. This prevented BasicPublisher and BasicSubscriber from working right, since you could never start the embedded media driver using a command line argument. The fix was to read the system property and then parse the boolean value. Previously, the system property was not being read at all, so null was being returned.


mjpt777:
Have you looked at the internal implementation of `Boolean.getBoolean()`? It is doing what your change does.

Can you further explain what is going wrong?

mogrifier:
Had not looked. What a horrible method name. I don't know why my IDE was not setting the property right, and thinking getBoolean had nothing to do with properties, I changed it and voila, it started working right so I figured that was a bug.

Oh well. At least I made my first pull request ever even if not needed. You can ignore the pull request (as you already know).

mjpt777:
Interesting learning experience at least :-)

PR #403, opened 2017-09-23
[Java] Makes tail initialisation atomic

franz1981:
A TermAppender fallen behind (ie in a past active partition), but that has already passed the backpressure check, could perform the claim when the tail is going to be reinitialised (ie the partition reactivated), hence the initialisation is needed to be atomic (a memory relaxed operation).
It is more likely to happen with a lot of producer threads (ie much more than the core count)and very small term capacity.

### DIFF
--- a/aeron-client/src/main/java/io/aeron/logbuffer/LogBufferDescriptor.java
+++ b/aeron-client/src/main/java/io/aeron/logbuffer/LogBufferDescriptor.java
@@ -552,7 +552,7 @@ public class LogBufferDescriptor
     public static void initialiseTailWithTermId(
         final UnsafeBuffer logMetaData, final int partitionIndex, final int termId)
     {
-        logMetaData.putLong(TERM_TAIL_COUNTERS_OFFSET + (partitionIndex * SIZE_OF_LONG), packTail(termId, 0));
+        logMetaData.putLongOrdered(TERM_TAIL_COUNTERS_OFFSET + (partitionIndex * SIZE_OF_LONG), packTail(termId, 0));
     }
 
     /**


mjpt777:
Thanks.

mjpt777:
BTW it would only happen if the producer was 3 or more terms behind as the result of an interrupt taken between reading the active index and then the tail.

franz1981:
Got it! Just to know: what could happen to such claim if it will land right after an activation beyond the producer limit?

mjpt777:
Can you try and ask the question another way? I don't quite understand.

franz1981:
Np, my fault: I'll try to explain it better.

Suppose to have a `Publication::offer` that go through the [back-pressure check](https://github.com/real-logic/aeron/blob/master/aeron-client/src/main/java/io/aeron/Publication.java#L393) and is going to perform the [(XADD) claim](https://github.com/real-logic/aeron/blob/master/aeron-client/src/main/java/io/aeron/logbuffer/TermAppender.java#L149) on the term.
Suppose that it will perform the claim right after the tail has been [reinitialised](https://github.com/real-logic/aeron/blob/master/aeron-client/src/main/java/io/aeron/Publication.java#L594) by another producer, but in a termId different from the one used for the backpressure check.
`(A)` If the term will have enough capacity the offer operation will succeed.

Now, suppose to be already in a overclaiming scenario ie where multiple producers has already surpassed the backpressure check and were able to (over)claim until the end of a term: there will be a producer that will trip the term and perform the initialisation of the next term tail before changing the active term.
Other concurrent offers will fail by design and be backpressured (we are already beyond the producer limit!).
If a "late" claim (described in `(A)`) will land right after the tail initialisation there won't be any counter-measures to stop that offer to succeed. it is correct?

PR #408, opened 2017-09-29
[C]: Created macro condition for glibc<5.21 recvmmsg function signature

bcaimano:
This is the small patch as discussed in gitter. Should be pretty straightforward, let me know if I've drifted from the code standards or something.


tmontgomery:
Awesome! Thanks!

Issue #414, opened 2017-10-16
Potential archive segment filename collision


zyulyaev:
Because of a typo in this line:
https://github.com/real-logic/aeron/blob/b55b005d6210f0622585793d8f80ef1a5488db6c/aeron-archive/src/main/java/io/aeron/archive/Archive.java#L756
one may get the same filename for different recordingId and segmentIndex pair. E.g. recordingId=1, segmentIndex=2 and recordingId=2,segmentIndex=1 gives you the same filename.

Solution is to replace single quotes with double quotes, so the line looks like:
```
return recordingId + "-" + segmentIndex + Configuration.RECORDING_SEGMENT_POSTFIX;
``` 
I believe this is what it was intended to be.

mjpt777:
Thanks. Yes you are correct. I've pushed a fix.

PR #415, opened 2017-10-16
[Java] Fix archive segment filename collision

zyulyaev:
#414


mjpt777:
Thanks. I was already fixing it.

Issue #331, opened 2017-04-03
Support units for size specification in properties


nitsanw:
It is more pleasant to specify in human-ish units so we can say:
``` aeron.term.buffer.length=512m ```
instead of:
``` aeron.term.buffer.length=536870912 ```

Support for g/m/k units would be nice.
Similar support for time units is probably also a good idea.

mjpt777:
👍 +1

PR #418, opened 2017-10-18
[Java]: Add magnitudes to property files

lfn3:
This should fix half of #331. Namely, it allows you to specify byte based sizes with suffixes of k, m, or g.

I'll put together a pr for the other half (time units) if or once you're happy with this. Any other tests or anything you reckon this needs?

### DIFF
--- a/aeron-driver/src/main/java/io/aeron/driver/Configuration.java
+++ b/aeron-driver/src/main/java/io/aeron/driver/Configuration.java
@@ -15,9 +15,13 @@
  */
 package io.aeron.driver;
 
-import io.aeron.*;
+import io.aeron.Aeron;
+import io.aeron.CommonContext;
+import io.aeron.Image;
+import io.aeron.Publication;
 import io.aeron.driver.exceptions.ConfigurationException;
-import io.aeron.driver.media.*;
+import io.aeron.driver.media.ReceiveChannelEndpoint;
+import io.aeron.driver.media.SendChannelEndpoint;
 import io.aeron.logbuffer.FrameDescriptor;
 import io.aeron.protocol.DataHeaderFlyweight;
 import org.agrona.BitUtil;
@@ -27,14 +31,13 @@ import org.agrona.concurrent.ControllableIdleStrategy;
 import org.agrona.concurrent.IdleStrategy;
 import org.agrona.concurrent.broadcast.BroadcastBufferDescriptor;
 import org.agrona.concurrent.ringbuffer.RingBufferDescriptor;
-import org.agrona.concurrent.status.*;
+import org.agrona.concurrent.status.CountersReader;
+import org.agrona.concurrent.status.StatusIndicator;
 
 import java.io.IOException;
 import java.net.StandardSocketOptions;
 import java.nio.channels.DatagramChannel;
 import java.util.concurrent.TimeUnit;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
 
 import static io.aeron.driver.ThreadingMode.DEDICATED;
 import static io.aeron.logbuffer.LogBufferDescriptor.PAGE_MAX_SIZE;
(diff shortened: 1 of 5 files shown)


mjpt777:
Looks good on initial review. I'll review in more detail tomorrow. Thanks!

aeron-driver/src/main/java/io/aeron/driver/Configuration.java
```diff
@@ -1096,4 +1098,68 @@ public static void validatePageSize(final int pageSize)
...
+            return defaultValue;
+        }
+    }
+
+    private static long getSize(final String propertyKey, final long defaultValue)
+    {
+        final String propertyValue = getProperty(propertyKey);
+        if (propertyValue != null)
+        {
+            return parseSize(propertyKey, propertyValue);
```
mjpt777: The parseSize method below only supports int. Should do a long method with a range check for if the int value is allowed.

aeron-driver/src/main/java/io/aeron/driver/Configuration.java
```diff
@@ -1096,4 +1098,68 @@ public static void validatePageSize(final int pageSize)
...
+
+    private static final Pattern SIZE_REGEX = Pattern.compile("(?<value>\\d+)(?<magnitude>\\w)?");
+    private static final String PARSE_SIZE_MAGNITUDE_FAILURE_FORMAT = "Couldn't parse value: %s for property %s. " +
+            "Trailing character should be one of k, m, or g, but was %s";
+    private static final String PARSE_SIZE_VALUE_FAILURE_FORMAT = "Couldn't parse value: %s for property %s. " +
+            "Should be a positive number followed by one of k, m, or g.";
+
+    public static int parseSize(final String propertyKey, final String sizeString)
+    {
+        final Matcher matcher = SIZE_REGEX.matcher(sizeString);
```
mjpt777: I think it would be simpler and more efficient to simply check the last character and see if it is not a digit.

aeron-driver/src/main/java/io/aeron/driver/Configuration.java
```diff
@@ -1096,4 +1098,68 @@ public static void validatePageSize(final int pageSize)
             throw new ConfigurationException("Page size not a power of 2: page size=" + pageSize);
         }
     }
+
+    private static int getSize(final String propertyKey, final int defaultValue)
+    {
+        final String propString = getProperty(propertyKey);
```
mjpt777: propValue is a better name than propString like in the long overload.

aeron-driver/src/main/java/io/aeron/driver/Configuration.java
```diff
@@ -1096,4 +1098,68 @@ public static void validatePageSize(final int pageSize)
...
+        {
+            throw new IllegalArgumentException(String.format(PARSE_SIZE_VALUE_FAILURE_FORMAT, sizeString, propertyKey));
+        }
+
+        final int value = Integer.valueOf(matcher.group("value"));
+        final String magnitude = matcher.group("magnitude");
+
+        if (magnitude != null)
+        {
+            switch (magnitude)
```
mjpt777: A switch on a char is more efficient than on a String.

aeron-driver/src/main/java/io/aeron/driver/Configuration.java
```diff
@@ -1096,4 +1098,68 @@ public static void validatePageSize(final int pageSize)
...
+            {
+                case "k":
```
mjpt777: Upper and lower case can be checked with a fall through.
```java
    case 'k':
    case 'K':
        return value * 1024;
```

aeron-driver/src/main/java/io/aeron/driver/Configuration.java
```diff
@@ -1096,4 +1098,68 @@ public static void validatePageSize(final int pageSize)
...
+                    return value * 1024;
+                case "m":
+                    return value * 1024 * 1024;
+                case "g":
```
mjpt777: In the `int` case the only valid value is `1`. It would be best to range check  as this can also check for overflow.

aeron-driver/src/main/java/io/aeron/driver/Configuration.java
mjpt777: This code will be ported to C so keeping it simple makes that easier by not using a regex. Startup time also matters and simple code is faster in the interpreter plus it generates less garbage.

aeron-driver/src/main/java/io/aeron/driver/Configuration.java
```diff
@@ -1096,4 +1098,68 @@ public static void validatePageSize(final int pageSize)
...
+                    return value * 1024 * 1024 * 1024;
+                default:
+                    throw new IllegalArgumentException(
+                            String.format(PARSE_SIZE_MAGNITUDE_FAILURE_FORMAT, sizeString, propertyKey, magnitude));
+            }
```
mjpt777: Profiling has shown us that String.format is very costly. I appreciate that this is an exception case but in general we are trying to avoid its use.

lfn3:
So maybe a more general explanation of why I did it this way is in order.
I wanted to try and make the time units version more or less the same, even if code isn't shared between them. Imo it's nicer if you can just play spot the difference between the two bits of code. 

Since the time units will be something like `ms`, `ns`, and maybe even `s`, it'll be matching on strings (although we could do something a bit sketch and just match on the first character?). Actually that suggests how we could keep it to a switch on char in that case as well.

mjpt777:
Thanks. I made a couple of commits to refactor the code to be more in keeping.

PR #421, opened 2017-10-22
[Java] parse times from properties with suffixes of ns, us, ms or s.

lfn3:
This is the other half of #331.


aeron-driver/src/main/java/io/aeron/driver/Configuration.java
```diff
@@ -1187,4 +1186,76 @@ public static long parseSize(final String propertyName, final String propertyVal
...
+        if (propertyValue != null)
+        {
+            final long value = parseTime(propertyName, propertyValue);
+            if (value < 0)
+            {
+                throw new ConfigurationException(
+                        "Value " + value + " for property " + propertyName + " must be positive");
+            }
+
+            if (value == Long.MAX_VALUE)
```
lfn3: This bit is here rather than below since all of the `TimeUnits...toNanos` calls return `Long.MAX_VALUE` when they hit positive overflow. I feel like not allowing `Long.MAX_VALUE` is an ok trade-off?

mjpt777:
Thanks. I've merged and move to Agrona so it can be applied beyond the driver configuration.

PR #422, opened 2017-10-24
[Java] have ExclusivePublication extend Publication.

### LINKED ISSUE #404: Have ExclusivePublication extend Publication
This will allow for substitution in APIs.  Don't use an interface to avoid the invoke interface calls in Java.

lfn3:
This should close #404.

Not sure if this is the approach you had in mind or not. (Figure you might have meant having an abstract class or duplicating the logic or something else?) Let me know if this fits with what you want, I'm totally happy to rework it if you had something else in mind.

### DIFF
--- a/aeron-client/src/main/java/io/aeron/ExclusivePublication.java
+++ b/aeron-client/src/main/java/io/aeron/ExclusivePublication.java
@@ -15,13 +15,14 @@
  */
 package io.aeron;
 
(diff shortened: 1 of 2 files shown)


mjpt777:
I've did a little more on this and pushed a change. The approach taken here has resulted in ~10% drop in throughput. I've a good idea why and will restructure the code to see if I can get the performance back.

PR #437, opened 2017-11-23
Fix linux/osx build warnings, and a windows test failure

cesaref:
I've enabled warnings as errors in the unix cmake build, and applied fixes for the various problems found. Mostly this has been a warning about _BSD_SOURCE being depricated, so i've moved this to only be defined on FreeBSD - this may be the wrong platform as i am unaware which you are targeting with this.

Publication/Exclusive publication had a bug on windows caused by long being 32 bit (not 64 bits). I've re-written this code to fix the code and hence the tests.

The use of dlsym() in the C code was problematic because of a void * to function pointer conversion that is apparently not portable (that's what the pedantic flag was saying). So, i've written a macro to perform the logic wrapping a nasty void * cast to avoid this.

There's an unused variable in aeron_loss_detector.c that i've commented out rather than removing - not sure whether this is a logic error or if it's just cruft and should be removed.

hdr_histogram has again a problem with a zero length array in a struct, and the pedantic option shows this up. I've submitted a pull to the project to get this fixed, and in the meantime wrapped this include in logic to disable this test whilst including it.


tmontgomery:
I like treating warnings as errors. However, due to how we have to handle older platforms with both clang and gcc, we can't do it easily. So, we can't really use it.

BSD_SOURCE is needed for all Linuxes. While it is deprecated currently, it has not always been such. So, it needs to stay as is.

dlsym is a known issue for gcc. Wrapping it this way is not how we should handle it, though. So, please leave as is.

Go ahead and remove the calculated_n from the loss detector.c file. It is a holdover from an earlier part.

The change in transport_poller.c is odd. Why was that added?

cesaref:
Yeah, dealing with older platforms is something i'm not looking at, and i'm not sure about that whole BSD_SOURCE thing anyway, why it was needed before. I'll certainly stick those back in. Is it going to be possible to wrap them in a #if so that we can optionally remove them on some platforms/builds? That way, they stay in place for the older linux platforms, but can be excluded, say, for debug builds with warnings as errors flagged?

The dlsym thing was a bit of a leap, and I agree it isn't pretty. I'll see if there is a better way of disabling this one warning, as it makes no sense on modern architectures (I think it's a hangover from segmented memory models from way back).

transport_poller.c was an 'unused local variable' in the case where neither HAVE_EPOLL or HAVE_POLL is defined (and the old capacity needs to be kept from before AERON_ARRAY_ENSURE_CAPACITY is called since it modifies the capacity). 

I'll pull out the simple fixes and we can deal with these separately.

cesaref:
I'll close this, i've raised a simpler one for the windows fixes

PR #455, opened 2018-01-17
[Java] Javadoc typos.

JPWatson:


aeron-client/src/main/java/io/aeron/Publication.java
```diff
@@ -103,7 +103,7 @@ protected Publication(
         final long registrationId,
         final int maxMessageLength)
     {
-        final UnsafeBuffer logMetaDataBuffer = logBuffers.metaDataBuffer();
```
mjpt777: This change would involve repeated field access rather than reuse from a register.

JPWatson: reverted.

PR #467, opened 2018-02-09
Use configured aeron directory for Archiver & Cluster clients

JPWatson:
Issue: ClustedMediaDriver/ArchivingMediaDriver fail if you specify and alternative aeron directory because the spawned aeron clients continue to use the default directory.

Not currently an issue in the ConsensusModule because it doesn't allow explicit override of the aeronDirectoryName.


mjpt777:
What if it is explicitly set as a system property?

JPWatson:
That's the case I'm trying to fix - at the moment, if it's set as a system property then it's being overridden by the default value for the Archiver/Archiver client/Cluster client.

PR #474, opened 2018-03-06
Specify charset explicitly for Admin Endpoint

JPWatson:


mjpt777:
If we do specify I feel it would be better to restrict to US-ASCII.

tmontgomery:
This is changing. Drastically. It'll be the same format as ClusterMember.parse. Including the memberId.

tmontgomery:
I've pushed the format change. I agree with @mjpt777 , the format used should be US-ASCII.

JPWatson:
Done

PR #478, opened 2018-03-10
Recovery from snapshot

JPWatson:
1. Fix bug in the RecoveryPlan where the wrong recording was fetched.
2. Adjust snapshot recovery so that if no log entry follows it, the correct logPosition is used for the new leadership term.


tmontgomery:
This broke the build for me. ClusterNodeRestartTest broken.

JPWatson:
Sorry Todd I should have labelled it better, it was a work in progress for Martin to have a look at in the morning.

PR #490, opened 2018-03-16
Notify ClusteredService that recovery has ended.

JPWatson:
onReplayEnd is not called when starting from clean state or recovering snapshot only.


tmontgomery:
Fine with me. I've requested Martin to review.

Issue #486, opened 2018-03-13
Improve endpoint parsing, i.e. don't use regex


mjpt777:
Improve endpoint parsing so it is less allocation and CPU intensive.

https://github.com/real-logic/aeron/blob/master/aeron-driver/src/main/java/io/aeron/driver/media/SocketAddressUtil.java#L26

mikeb01:
I'll have a look at this.

PR #491, opened 2018-03-19
[Java] Remove regular expressions from socket address parsing.

mikeb01:
Replace regex with state machine based parser, #486.


mjpt777:
Thanks Mike.

PR #526, opened 2018-07-12
set serviceId in serviceheartbeat counter key

benwimpory:
No obvious place to add a test for this, other than creating a new multi-service test, or creating a dedicated unit test for the ServiceHeartbeat class.  Happy to add if you let me know where.


benwimpory:
@mjpt777 fixed this

PR #531, opened 2018-07-25
Change the round-robin mechanism of Subscription's poll methods

denizevrenci:
To be able to continue from the next `Image` in subsequent calls.

### DIFF
--- a/aeron-client/src/main/cpp/Subscription.h
+++ b/aeron-client/src/main/cpp/Subscription.h
@@ -144,20 +144,24 @@ public:
         Image *images = imageList->m_images;
         int fragmentsRead = 0;
 
-        std::size_t startingIndex = m_roundRobinIndex++;
-        if (startingIndex >= length)
-        {
-            m_roundRobinIndex = startingIndex = 0;
-        }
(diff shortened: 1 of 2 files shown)


mjpt777:
What is the goal to be achieved with the change?

denizevrenci:
If the number of fragments in earlier images in the list is larger than `fragmentLimit`, processing the fragments in the last image in the list takes `N` calls to `poll()` as we lose the position (`i`) in the list. After this change, the subsequent calls "remember" the last processed position (`i`). So `poll()` can iterate the list faster.

With a small number of images, the effect is probably unnoticeable. And if each image has more than `fragmentLimit` number of unprocessed fragments, this change does not affect the behaviour.

It is a micro-optimization that will probably affect very few use cases. If you think it does not serve a purpose, please feel free to close this PR.

mjpt777:
This broke the build and so reverted it.

denizevrenci:
Sorry to pollute the commit history. Seems like cluster tests are failing with index out of bounds errors. I'll investigate and update the PR.

PR #555, opened 2018-09-25
Use relative path in SourcedException

lukepalmer:
Today, SourcedException uses the absolute path of wherever aeron was built. I'd instead like it to refer to the relative path within the source tree.

It's probably misleading for a potential production exception to refer to /home/luke/aeron... or something to that effect.


mjpt777:
Thanks but note the C driver build failed due to missing flag for the C build.

PR #559, opened 2018-10-05
Multiple commit position counters

JPWatson:
Close commitPosition to avoid multiple counters after an election:

```
81:                3,968 - rec-pos: 0 -1324327184 100 aeron:udp?control=192.168.222.12:10007|sparse=true|session-id=-1324327184
 82:                3,968 - commit-pos: leadershipTermId=1
 83:                   34 - Sequence Number
 84:               99,999 - Current Value
 85:                1,408 - commit-pos: leadershipTermId=0
 86:                    0 - pub-pos (sample): 110 -1324327176 102 aeron:udp?endpoint=192.168.222.10:10000
```


mjpt777:
Thinking of races with the service running a bit behind. I think a single commit position counter should be used now the log is monotonic across leadership terms. This way the counter carries forward. Closing would cause issues as the service is catching up.

mjpt777:
Code is now migrated to a single commit position counter.

PR #563, opened 2018-10-09
Initialise variable (in case std::cin is closed or otherwise read from it fails)

### LINKED ISSUE #562: [C++] Read of possibly uninitialised variable
Picked up by `clang --analyze`:

```
In file included from 
include/concurrent/AtomicBuffer.h:25:
include/util/StringUtil.h:146:13: warning: The right operand of '==' is a garbage value
    if ('y' == response || 'Y' == response)
            ^  ~~~~~~~~
```

And it's not wrong:

```cpp
inline bool continuationBarrier(const std::string& label)
{
    bool result = false;
    char response;
    std::cout << std::endl << label << " (y/n): ";
    std::cin >> response;  // <--------- if the cin fails here (e.g. stdin has been closed), response will be left uninitialised

    if ('y' == response || 'Y' == response)
    {
        result = true;
    }

    return result;
}
```

mattgodbolt:
Fixes #562 (which addresses a `clang --analyze` static code check failure).

### DIFF
--- a/aeron-client/src/main/cpp/util/StringUtil.h
+++ b/aeron-client/src/main/cpp/util/StringUtil.h
@@ -139,7 +139,7 @@ std::string strconcat (Ts... vs)
 inline bool continuationBarrier(const std::string& label)
 {
     bool result = false;
-    char response;
+    char response = '\0';
     std::cout << std::endl << label << " (y/n): ";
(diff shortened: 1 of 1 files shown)


mjpt777:
Thanks

Issue #560, opened 2018-10-08
[C++] Build fails after release 1.11.2


mo2menelzeiny:
Migrating from 1.11.1 to 1.11.2 produced a failed build, my guess it's related to the relative path feature merged in #555.

I'm installing the library to `/usr/local/`, does that  miss-configure the relative path substitution?  https://github.com/real-logic/aeron/blob/44e7e43e820fb34accbc73e7d5b67cd952798a8b/CMakeLists.txt#L59 

#### Build script: 
```
wget https://github.com/real-logic/aeron/archive/1.11.2.tar.gz
tar zvxf 1.11.2.tar.gz
rm 1.11.2.tar.gz
cd aeron-1.11.2
mkdir -p cppbuild/Debug
cd cppbuild/Debug
cmake -DBUILD_AERON_DRIVER=ON ../..
make -j 4
make -j 4 install PREFIX=/usr/local
```

#### Build error log: 
```
In file included from /usr/local/include/concurrent/AtomicBuffer.h:25:0,
                 from /usr/local/include/concurrent/logbuffer/TermReader.h:22,
                 from /usr/local/include/concurrent/AgentRunner.h:24,
                 from /usr/local/include/Context.h:22,
                 from /home/mo2menelzeiny/Projects/arbito/include/Messenger.h:6,
                 from /home/mo2menelzeiny/Projects/arbito/src/Messenger.cpp:2:
/usr/local/include/util/StringUtil.h: In function ‘valueType aeron::util::parse(const string&)’:
/usr/local/include/util/StringUtil.h:63:29: error: expected primary-expression before ‘(’ token
         throw ParseException(std::string("Failed to parse: ") + input, SOURCEINFO);
                             ^
In file included from /usr/local/include/Context.h:21:0,
                 from /home/mo2menelzeiny/Projects/arbito/include/Messenger.h:6,
                 from /home/mo2menelzeiny/Projects/arbito/src/Messenger.cpp:2:
/usr/local/include/util/StringUtil.h:63:72: error: expected ‘)’ before ‘__FILENAME__’
         throw ParseException(std::string("Failed to parse: ") + input, SOURCEINFO);
                                                                        ^
In file included from /usr/local/include/concurrent/logbuffer/TermReader.h:22:0,
                 from /usr/local/include/concurrent/AgentRunner.h:24,
                 from /usr/local/include/Context.h:22,
                 from /home/mo2menelzeiny/Projects/arbito/include/Messenger.h:6,
[...]

PR #571, opened 2018-10-25
Add CI builds

lukepalmer:
I've added configuration for travis builds covering linux and OSX.
If accepted, the maintainers have to press a button at https://travis-ci.org to activate.

This change is motivated because in #560 I apparently created a build issue that I can't reproduce.

Incidentally, I find that OSX builds currently fail on io.aeron.MemoryOrderingTest. I've marked OSX as allowed to fail in the submission for now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,118 @@
+language: java
+
+os:
+- linux
+
+before_script:
+- export CXX=$COMPILER
+# Add an IPv6 config - see the corresponding Travis issue
+# https://github.com/travis-ci/travis-ci/issues/8361
+- if [ "${TRAVIS_OS_NAME}" == "linux" ]; then
+  sudo sh -c 'echo 0 > /proc/sys/net/ipv6/conf/all/disable_ipv6';
+  fi
+
+# don't run gradle during C++ builds
+install: skip
+
+matrix:
+  include:
+  - name: "Oracle Java 8"
+    language: java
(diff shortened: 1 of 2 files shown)


mjpt777:
Thanks @lukepalmer. I've been considering adding a CI build for Aeron. I have some concerns about sufficient CPUs being available for the system tests for example.

If we add the support for C/C++ then the C driver should also be built and with a matrix of compilers. The build for SBE could serve as a template. https://github.com/real-logic/simple-binary-encoding/blob/master/.travis.yml

lukepalmer:
Compiler matrix sounds reasonable, I can take a shot at that.

Note that the submitted change does do both the java and C/C++ builds. So far, the system test does appear to work reliably on the 2 cores travis gives us.

Is the OS X failure known? Here's an example build. It's consistent.
https://travis-ci.org/lukepalmer/aeron/builds/446126370

mjpt777:
The build failure is a timeout. Likely due to insufficient CPU resource. I also note it is building with Java 10.0.1 rather than 8.

lukepalmer:
Ok there we are. All linux builds for now, but we have Oracle JDK 8 and 11 and a full range of C/C++ builds. 

I did see a single timeout failures on JDK11 but worked all other times.

mjpt777:
Thanks. I've swapped Oracle JDK 11 for OpenJDK 11 and all seems to be passing now.

PR #581, opened 2018-11-09
Treat Egress Channel as NOT_CONNECTED on InvalidChannelException

JPWatson:
A channel that was resolvable at connection time can subsequently become unresolvable which causes a cascade of errors when a node becomes leader or the cluster is coming up after shutdown.

```
 io.aeron.exceptions.RegistrationException: errorCodeValue=1 java.net.UnknownHostException: could not resolve endpoint address: xxx.host:9020
    at io.aeron.ClientConductor.onError(ClientConductor.java:169)
    at io.aeron.DriverEventsAdapter.onMessage(DriverEventsAdapter.java:85)
    at org.agrona.concurrent.broadcast.CopyBroadcastReceiver.receive(CopyBroadcastReceiver.java:120)
    at io.aeron.DriverEventsAdapter.receive(DriverEventsAdapter.java:56)
    at io.aeron.ClientConductor.service(ClientConductor.java:718)
    at io.aeron.ClientConductor.awaitResponse(ClientConductor.java:762)
    at io.aeron.ClientConductor.addExclusivePublication(ClientConductor.java:407)
    at io.aeron.Aeron.addExclusivePublication(Aeron.java:253)
    at io.aeron.cluster.ClusterSession.connect(ClusterSession.java:130)
    at io.aeron.cluster.ConsensusModuleAgent.becomeLeader(ConsensusModuleAgent.java:1113)
    at io.aeron.cluster.Election.leaderTransition(Election.java:653)
    at io.aeron.cluster.Election.doWork(Election.java:209)
    at io.aeron.cluster.ConsensusModuleAgent.doWork(ConsensusModuleAgent.java:272)
    at org.agrona.concurrent.AgentRunner.doDutyCycle(AgentRunner.java:268)
    at org.agrona.concurrent.AgentRunner.run(AgentRunner.java:161)
    at java.base/java.lang.Thread.run(Thread.java:844)
```

### DIFF
--- a/aeron-cluster/src/main/java/io/aeron/cluster/ClusterSession.java
+++ b/aeron-cluster/src/main/java/io/aeron/cluster/ClusterSession.java
@@ -21,7 +21,10 @@ import io.aeron.CommonContext;
 import io.aeron.Publication;
 import io.aeron.cluster.client.ClusterException;
 import io.aeron.cluster.codecs.CloseReason;
+import io.aeron.driver.exceptions.InvalidChannelException;
+import io.aeron.logbuffer.BufferClaim;
 import org.agrona.CloseHelper;
+import org.agrona.DirectBuffer;
 import org.agrona.collections.ArrayUtil;
 
 import java.util.Arrays;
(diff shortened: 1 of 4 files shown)


JPWatson:
This was the simplest way I could see to solve this.

I wonder whether the behaviour might be handled instead by the MediaDriver which could continue to create a 'null publication' that never connects?

mjpt777:
I think we need to consider some sort of timeout on this so that it is remove eventually, like it does in the initial connect case. For example, if ingress is still working then keepalive is not applied if egress is not connected.

JPWatson:
I was working on the basis that the session would be timed out anyway because there would be no messages would be coming in.

mjpt777:
We need to consider if there is so it is covered.

Issue #579, opened 2018-11-07
[C++] LogBufferDescriptor::indexByPosition triggers integer overflow


denizevrenci:
When I reassign a publication to the same channel by

```cpp
publication = aeron.findPublication(regId)
```

Thereby, the previous publication object managed by the `shared_ptr` is destroyed and a new object is started to be managed by the `shared_ptr`. The subscription side receives the new image event before the unavailable image event.

During this time, if I poll the subscription, the soon-to-be-unavailable image is still accessible by the subscription. However, when `poll` traverses up the stack, it seems to try to access destructed memory. Below is the relevant part of the stack trace.

```
Thread 1 "test" received signal SIGSEGV, Segmentation fault.
0x000000000040cf30 in aeron::concurrent::atomic::getInt32Volatile (source=0x100000001)

#0  0x000000000040cf30 in aeron::concurrent::atomic::getInt32Volatile (source=0x100000001)
    at ../thirdparty/aeron/source/aeron-client/src/main/cpp/concurrent/atomic/Atomic64_gcc_x86_64.h:67

#1  0x000000000040cf19 in aeron::concurrent::AtomicBuffer::getInt32Volatile (this=0x7ffff0001458, offset=0)
    at ../thirdparty/aeron/source/aeron-client/src/main/cpp/concurrent/AtomicBuffer.h:242

#2  0x000000000040740c in aeron::concurrent::logbuffer::FrameDescriptor::frameLengthVolatile (logBuffer=..., frameOffset=0)
    at ../thirdparty/aeron/source/aeron-client/src/main/cpp/concurrent/logbuffer/FrameDescriptor.h:149

#3  0x000000000040fa03 in aeron::concurrent::logbuffer::TermReader::read<...>(...) (outcome=..., termBuffer=..., termOffset=0, handler=..., fragmentsLimit=2147483647, header=..., exceptionHandler=...) at ../thirdparty/aeron/source/aeron-client/src/main/cpp/concurrent/logbuffer/TermReader.h:85

#4  0x000000000040f969 in aeron::Image::poll<...>(...) (this=0x7ffff0001478, fragmentHandler=..., fragmentLimit=2147483647) at ../thirdparty/aeron/source/aeron-client/src/main/cpp/Image.h:398

[...]

PR #582, opened 2018-11-14
Add assertions and cleanup

denizevrenci:
Related to #579

### DIFF
--- a/aeron-client/src/main/cpp/concurrent/logbuffer/LogBufferDescriptor.h
+++ b/aeron-client/src/main/cpp/concurrent/logbuffer/LogBufferDescriptor.h
@@ -146,7 +146,7 @@ static const util::index_t LOG_PAGE_SIZE_OFFSET = (util::index_t)offsetof(LogMet
 static const util::index_t LOG_DEFAULT_FRAME_HEADER_OFFSET = (util::index_t)sizeof(LogMetaDataDefn);
 static const util::index_t LOG_META_DATA_LENGTH = 4 * 1024;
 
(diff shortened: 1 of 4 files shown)


aeron-client/src/main/cpp/concurrent/logbuffer/LogBufferDescriptor.h
```diff
@@ -354,7 +354,7 @@ inline static void rotateLog(AtomicBuffer& logMetaDataBuffer, std::int32_t curre
     LogBufferDescriptor::casActiveTermCount(logMetaDataBuffer, currentTermCount, nextTermCount);
 }
 
-inline static void initializeTailWithTermId(AtomicBuffer& logMetaDataBuffer, int partitionIndex, std::int32_t termId)
+inline void initializeTailWithTermId(AtomicBuffer& logMetaDataBuffer, int partitionIndex, std::int32_t termId)
```
mjpt777: Why remove the static declaration from methods? Would this not be a breaking API change?

denizevrenci: It may break the ABI in case someone wants to compile object files with the new header and link with the client library that was compiled with the old headers. With a simple test case, I confirmed that two object files still link if one of them is `inline static` and the other is only `inline`. The advantage is that the resulting binary is smaller as the linker is now free to eliminate repeating `inline` symbols as they are not internally linked.

tmontgomery:
Can this be brought up to date? Right now it has conflicts.

denizevrenci:
Rebased on master.

PR #583, opened 2018-11-15
Improve SessionSpecificSubscriptionTest with new case shouldNotSubscribeWithoutSpecificSession

QIvan:
Hi! Thank you very much for this test it makes so much sense for me and helps to understand how sessionId works. 
But i think with this patch makes the test a little bit more useful for people like me who tries to figure out what sessionId is.

but feel free to decline it if you think it's redurant.

### DIFF
--- a/aeron-system-tests/src/test/java/io/aeron/SessionSpecificSubscriptionTest.java
+++ b/aeron-system-tests/src/test/java/io/aeron/SessionSpecificSubscriptionTest.java
@@ -26,12 +26,15 @@ import org.junit.Test;
 
 import java.nio.ByteBuffer;
 
+import static io.aeron.CommonContext.UDP_MEDIA;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.is;
+import static org.junit.Assert.assertFalse;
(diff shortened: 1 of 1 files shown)


aeron-system-tests/src/test/java/io/aeron/SessionSpecificSubscriptionTest.java
```diff
@@ -74,15 +77,23 @@ public void shouldSubscribeToSpecificSessionIdsAndWildcard()
...
         {
             while (!publication1.isConnected() || !publication2.isConnected())
             {
                 SystemTest.checkInterruptedStatus();
                 Thread.yield();
             }
 
+            while (subscription1.imageCount() != 1 ||
+                    subscription2.imageCount() != 1 ||
+                    subscriptionWildcard.imageCount() != 2)
```
QIvan: we can't assert this conditionals here. Right? 
Because a publication is connected if there is at least one subscription but here it might be either `subscription1(2)` or `subscriptionWildcard`

mjpt777: We cannot assert because we must wait until they all reach that state.

PR #588, opened 2018-11-29
Use relative source path in exceptions

lukepalmer:
Have sourced exceptions contain text that looks like aeron-client/.../filename instead of /home/luke/aeron/aeron-client....

Note that #555 was a previous attempt at this.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -55,6 +55,9 @@ if(NOT CMAKE_BUILD_TYPE)
     set(CMAKE_BUILD_TYPE "Release" CACHE STRING "Choose the type of build" FORCE)
 endif(NOT CMAKE_BUILD_TYPE)
 
+# relative file paths for use in exceptions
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -55,6 +55,9 @@ if(NOT CMAKE_BUILD_TYPE)
     set(CMAKE_BUILD_TYPE "Release" CACHE STRING "Choose the type of build" FORCE)
 endif(NOT CMAKE_BUILD_TYPE)
 
+# relative file paths for use in exceptions
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -D__CMAKE_SOURCE_DIR__='\"${CMAKE_SOURCE_DIR}\"'")
```
denizevrenci: I know I suggested passing `CMAKE_SOURCE_DIR` but on second thought, it would be better to pass `PROJECT_SOURCE_DIR`.

If Aeron project is included by another project with `add_subdirectory`, `CMAKE_SOURCE_DIR` will point to the containing project's source directory while `PROJECT_SOURCE_DIR` will still point to Aeron's project root.

This will not affect standalone builds.

PR #594, opened 2018-12-11
truncateRecording doesn't effectively deletes the recording

QIvan:
Hi! 
I've added a test which shows what exactly doesn't work. The second `checkRecordingsCount` fails. 
Did I right understand 

> Truncating a recording to the start position effectively deletes the recording.

? My understanding It should remove a recording at all but now it leaves a record with `startPosition=0` and `endPosition=0`. 

(BTW is an issue report like this OK? )

### DIFF
--- a/aeron-system-tests/src/test/java/io/aeron/archive/BasicArchiveTest.java
+++ b/aeron-system-tests/src/test/java/io/aeron/archive/BasicArchiveTest.java
@@ -17,6 +17,7 @@ package io.aeron.archive;
 
 import io.aeron.*;
 import io.aeron.archive.client.AeronArchive;
+import io.aeron.archive.client.RecordingDescriptorConsumer;
 import io.aeron.archive.status.RecordingPos;
 import io.aeron.driver.MediaDriver;
 import io.aeron.driver.ThreadingMode;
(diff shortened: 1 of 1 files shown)


tmontgomery:
@mjpt777 can confirm, but I believe the desired effect is to delete all the segment files. This then allows ClusterTool to mark the entry as INVALID when run with `verify`.

mjpt777:
The behaviour is as expected. After a full truncate back to `startPosition` there are no segment files and the entry in the `catalog` reflects this. This commit show clarifies the state after a full truncate. https://github.com/real-logic/aeron/commit/0c0d0b54178521246d8c6208e0639d2eb5b40c6e

mjpt777:
@QIvan This is a reasonable way to raise an issue. A refinement would be to raise an issue and reference a PR. In the PR it is good to `@Ignore` the test so CI can pass the PR.

QIvan:
@mjpt777 Ok got. The good point about `@Ignore` thanks.

QIvan:
if we have 2 records: 
```
0 -> [0 ... 100]
1 -> [0 ... 200]
```
then we truncate recording 1 to the 0 position. Now we have 
```
0 -> [0 ... 100]
1 -> [0 ... 0]
```
Right?

Is it possible somehow to completely remove  a recording then?

mjpt777:
The catalog retains the entry because the recording id should not be reused.

PR #601, opened 2018-12-18
Macro guard against missing _PROJECT_SOURCE_DIR_

LordNacho:
As per discussion https://github.com/real-logic/aeron/pull/596


LordNacho:
@tmontgomery any chance I can get you to bump the tag to 1.13.1?

tmontgomery:
1.13.1 will be tagged when we do our next release. Unless we do 1.14 instead.

PR #613, opened 2019-01-17
[Java] Add a test to validate that client message delivery will continue over a leader failover

epickrram:


tmontgomery:
We've moved to a cleaned up set of classes for this type of thing. See `ClusterTest`, `TestCluster` and `TestNode`

mjpt777:
Is this still relevant?

epickrram:
Probably worth having at some point, but I'll close the PR while I get around to re-working it.

PR #616, opened 2019-01-22
Share egress publication between CM & CSC

JPWatson:
Use shared egress publication in ClusteredService. CM already ensures the Publication is connected as part of the session processing logic.

Sharing this publication reduces the likelyhood of having to block the service thread. 
e.g. if a message needs to be sent as a result of onSessionOpen.

### DIFF
--- a/aeron-cluster/src/main/java/io/aeron/cluster/ClusterSession.java
+++ b/aeron-cluster/src/main/java/io/aeron/cluster/ClusterSession.java
@@ -16,7 +16,6 @@
 package io.aeron.cluster;
 
 import io.aeron.Aeron;
-import io.aeron.ChannelUri;
 import io.aeron.Publication;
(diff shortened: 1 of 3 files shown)


mjpt777:
I'm not convinced this is the correct approach. It means the Consensus Module and services must use the same term length and MTU.  Given that the Consensus Module sends back almost no traffic then this puts a throughput limitation on the service that seems unnecessary, especially in burst scenarios.

mjpt777:
This also introduces head-of-line blocking when there are multiple services.

mjpt777:
@tmontgomery This makes the case of a service having to wait on the publication connecting go away which is nice to have. However I want to make sure we don't give up too much or miss other ways to do it.

mjpt777:
There are a number of benefits to this that probably outweigh the consequences.  We can always revisit it later.

PR #630, opened 2019-02-03
aeronmd: Change aeron_uri_parse to accept len.

mianos:
When the URI is received in a command over UDP it does not have a trailing null. 
The size of the URI is known in the command buffer and does not need to be null terminated.
In some cases aeron_uri_parse is called without a null terminated string so it fails to parse the address, even though the URI is correctly received.

### DIFF
--- a/aeron-driver/src/main/c/aeron_driver_conductor.c
+++ b/aeron-driver/src/main/c/aeron_driver_conductor.c
@@ -1726,7 +1726,7 @@ int aeron_driver_conductor_on_add_ipc_publication(
     aeron_uri_t aeron_uri_params;
     aeron_uri_publication_params_t params;
 
-    if (aeron_uri_parse(uri, &aeron_uri_params) < 0 ||
+    if (aeron_uri_parse(uri, (size_t)uri_length, &aeron_uri_params) < 0 ||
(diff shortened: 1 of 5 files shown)


mianos:
Needs some more work for the other unit tests. I'm onto this now.

mjpt777:
Thanks @mianos. I noticed a similar issue just recently and missed fixing this case too. I can pick this up.

mianos:
I seem to have broken the UDP channel parser but in debugging it I hit this line:
https://github.com/real-logic/aeron/blob/master/aeron-driver/src/test/c/aeron_udp_channel_test.cpp#L85
and I can't figure out why it subtracts one from the string. If you are not going to use this pull request and do it in your own manner go right ahead and I'll use my fixed version with the broken tests until you get yours in master.

mjpt777:
I've pushed a fix https://github.com/real-logic/aeron/commit/40fb737fff97a37487a1a5cacedaedf040f3647c

I've no idea why the -1 was there either. Maybe @tmontgomery knows?

tmontgomery:
The `uri_length` using `strlen - 1` is probably an error. It seems to only be used for doing the copy of the URI for storage as the original URI.

LordNacho:
Does this fix pass all the tests now?

tmontgomery:
@mjpt777 has pushed a fix instead.

PR #631, opened 2019-02-06
Various cleanup of C++ client

denizevrenci:
And two new features:
* Make Image movable without a copy.
* Add an accessor to Context for media driver timeout.

### DIFF
--- a/aeron-client/src/main/cpp/Aeron.h
+++ b/aeron-client/src/main/cpp/Aeron.h
@@ -64,7 +64,7 @@ public:
      * @param context for configuration of the client.
      */
     Aeron(Context& context);
-    virtual ~Aeron();
+    ~Aeron();
 
     /**
      * Create an Aeron instance and connect to the media driver.
--- a/aeron-client/src/main/cpp/BufferBuilder.h
+++ b/aeron-client/src/main/cpp/BufferBuilder.h
@@ -35,10 +35,6 @@ public:
     {
     }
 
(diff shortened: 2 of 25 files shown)


mjpt777:
A lot of this looks good. However removing the `Context::conclude()` changes the semantics for how contexts are used by convention. This will expand with archive and cluster being integrated.

tmontgomery:
Couple/Few things.

* not sure about asserting on clz value of 0. I know why. Just not comfortable with it instead of a guard.
* `Context.conclude` is needed in general. We could get away with it, but better to keep it.
* I've tried various things with Images. Copying vs. Moving, etc. This is fine. But I think in general, we will eventually need to move to be compatible with how Java does it instead of trying to be local with the array. In other words, Image should be a standalone object. And the array in Subscription should be an array of pointers to Images.

denizevrenci:
@mjpt777 I have put back `conclude()` and just changed the initializers to defaults instead of nulls.

@tmontgomery What kind of guard do you have in mind instead of assertions? To me, it looked like passing zero to `clz` and `ctz` is a programming error that cannot be handled at runtime.

I suppose the standalone Image can also be moved, so the addition of move semantics here would not have to be retracted later.

tmontgomery:
I think a move for a reference doesn't make as much sense. But, in general, we can take this and then deal with that when it is needed.

The guard I was thinking of would be a return of the bit size as Java does.

PR #633, opened 2019-02-08
[Java] Test showing recovery issue with timers.

ZachBray:
This test captures some behaviour we've seen at a client. 

Please let me know if you'd like me to clean it up at all. I considered injecting a different `ClusteredService` implementation rather than adding the switching in the existing one. I'm not sure what your preference is.

We see follower nodes stuck whilst catching up after performing the following steps:
1. Starting a 3 node cluster
2. Stopping a follower node
3. Sending a message to the cluster that triggers a timer to be
registered
4. Starting the follower node (again)

I've added two tests:
- `shouldCatchUpAfterFollowerMissesAMessage()` passes; but
- `shouldCatchUpAfterFollowerMissesTimerRegistration()` fails

The only difference between them is that the failing one registers a timer in the `onSessionMessage(...)` handler of `ClusteredService`.

The failure I see is:
```
Expected :is null
Actual   :<FOLLOWER_CATCHUP>
```

I'm running on JDK 1.8 (202 AMZN).

### DIFF
--- a/aeron-cluster/src/test/java/io/aeron/cluster/TestNode.java
+++ b/aeron-cluster/src/test/java/io/aeron/cluster/TestNode.java
@@ -41,6 +41,7 @@ import static org.agrona.BitUtil.SIZE_OF_INT;
 
 class TestNode implements AutoCloseable
 {
+
     private final ClusteredMediaDriver clusteredMediaDriver;
     private final ClusteredServiceContainer container;
     private final TestService service;
@@ -240,6 +241,12 @@ class TestNode implements AutoCloseable
             final int length,
             final Header header)
         {
+            final String message = buffer.getStringWithoutLengthAscii(offset, length);
+            if (message.equals(TestMessages.REGISTER_TIMER))
+            {
+                cluster.scheduleTimer(1, cluster.timeMs() + 1_000);
+            }
+
             while (session.offer(buffer, offset, length) < 0)
             {
                 cluster.idle();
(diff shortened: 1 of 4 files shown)


mjpt777:
@ZachBray This is failing the checkstyle build. Please do a full build and fix then I'll review and merge.

ZachBray:
Sorry about that. Should be passing the checkstyle build now.

aeron-cluster/src/test/java/io/aeron/cluster/TestNode.java
```diff
@@ -240,6 +241,12 @@ public void onSessionMessage(
             final int length,
             final Header header)
         {
+            final String message = buffer.getStringWithoutLengthAscii(offset, length);
+            if (message.equals(TestMessages.REGISTER_TIMER))
+            {
+                cluster.scheduleTimer(1, cluster.timeMs() + 1_000);
```
ZachBray: I've tried varying the expiry. At `10_000` milliseconds the test does not fail. It seems `TimerService.onTimerExpiry()` must be triggered on the leader for the test to fail.

ZachBray: This should probably be lower to reduce "intermittency"

ZachBray:
I may be wrong but it looks to me like the follower gets stuck in a while loop inside `ConsensusModuleAgent.catchupLogPoll()`.

```
while (!missedTimersSet.isEmpty()) ...
```

The follower adds timer event to its `missedTimersSet` when replaying.

The `ClusteredService` on that follower is never called with `onSessionMessage()` to register the timer.

PR #645, opened 2019-03-18
Use system defined iovec rather than a custom one.

mikeb01:
If this is not the correct approach I can create a scoped aeron_spsc_iovec (comments @tmontgomery ?).


tmontgomery:
It will fail on Windows (probably), but better to define for specific platforms that need it.

PR #646, opened 2019-03-19
[Java] Add Cluster and Archive events to logging agent.

epickrram:
This PR is starting to get chunky, so would be keen to merge before I start on the Archive response events.


mjpt777:
Thanks Mark. I'll make some changes to it tomorrow so that it can work if the archive or cluster is run separate from the driver.

tmontgomery:
C media driver agent could use this also. ;)

PR #647, opened 2019-03-21
[Java] Remove source of garbage when cancelling missed timers

epickrram:
Use of the `Predicate` was causing the `long` values to be auto-boxed to `Long`.


mjpt777:
Damn JIT not smart enough.

PR #651, opened 2019-03-29
[Work in progress + requires feedback] issue relating to followers catching up from scratch 

ZachBray:
## Background

We've noticed the following:

```
ERROR aeron - consensus module error
io.aeron.exceptions.TimeoutException: no heartbeat from service
```

after performing the following steps:

1. Running a three node cluster
2. Connecting a client and sending some commands to the cluster to create a bit of log
3. Stopping the cluster
4. Deleting the data from the follower nodes
5. Restarting all the nodes at the same time

## Changes in this PR

- I've added a test to demonstrate the error they were seeing: 31176f1
- I've added a fix to avoid the timeout due to the clustered service container not servicing the heartbeat: 4720a92 (this is not the underlying issue)
- I've added (+ removed) some debug logging (which I will rebase away entirely if any of this is heading in the right direction)
- I've added some much scarier changes to defer member status processing during leadership transition: 
56b3392

I'm not that familiar with the various pieces of Aeron Cluster yet so **I apologise if any of this is stupid!** :-)

### Suspected root cause

1. Three nodes were started at roughly the same time
2. Only one node had the "history" of log
3. This node with "history" was elected the leader
4. The leader entered the `LEADER_REPLAY` state and didn't immediately transition out
because `shouldReplay == true`
5. The followers published their catch-up positions
6. The leader polled for member status which dispatches catch-up
positions to the `onCatchupPosition(...)` handler
7. The leader dropped the catch-up positions because its leadership term
and role are not updated until `LEADER_TRANSITION` is reached.
Therefore, the log is never replayed to followers.

### Changes to polling for member status

I've tried to skip polling for member status whilst the consensus module
for a node is in `{ LEADER_REPLAY, LEADER_TRANSITION }`. This means we
don't process catch-up positions (amongst other messages) from followers
during this leadership transition.

### Big/scary assumptions

- `LEADER_REPLAY` always transitions to `LEADER_READY`. If not, then I've
introduced the possibility for a node to become "stuck" in a
transitioning state.

- It is safe to defer the processing of member status events

### Questions

1. Would it be better to signal `LEADER_READY` to followers before they
publish their catch-up positions?

2. Would it be better to transition the `role` and `leadershipTermId`
earlier rather than pausing member status processing?

3. Is the `LEADER_REPLAY` behaviour different on startup in order to
replay the log to the clustered service container? I.e., when it is
replaying is it replaying to "itself"?

### Known issues with this change

`ClusterTest.shouldCatchupFromEmptyLogThenSnapshotAfterShutdownAndFollowerCleanStart()`
now fails ~ 50% of the time. It fails because when processing an
acknowledgement of `Snapshot` from the clustered service container it
cannot find the entry for the previous leadership term in the recording
log. I believe this race was already present but I have made it worse;
however, I'm not 100% sure!

### Test trace

```
objenesis/2.6/639033469776fd37c08358c6b92a4761feb2af4b/objenesis-2.6.jar com.intellij.rt.execution.junit.JUnitStarter -ideVersion5 -junit4 io.aeron.cluster.ClusterTest,shouldCatchUpTwoFreshNodesAfterRestart
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by org.agrona.nio.TransportPoller (file:/home/dev/.gradle/caches/modules-2/files-2.1/org.agrona/agrona/0.9.35/487c1a2eeb85d3715b27487c3d9b1a8bbda42ca4/agrona-0.9.35.jar) to field sun.nio.ch.SelectorImpl.selectedKeys
WARNING: Please consider reporting this to the maintainers of org.agrona.nio.TransportPoller
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
memberId=0 nowMs=1553872798868 INIT -> CANVASS
memberId=1 nowMs=1553872798939 INIT -> CANVASS
memberId=2 nowMs=1553872799026 INIT -> CANVASS
memberId=2 nowMs=1553872799026 CANVASS -> NOMINATE
memberId=0 nowMs=1553872799028 CANVASS -> NOMINATE
memberId=1 nowMs=1553872799029 CANVASS -> NOMINATE
memberId=2 nowMs=1553872799073 NOMINATE -> CANDIDATE_BALLOT
memberId=0 nowMs=1553872799075 NOMINATE -> FOLLOWER_BALLOT
memberId=1 nowMs=1553872799079 NOMINATE -> FOLLOWER_BALLOT
memberId=2 nowMs=1553872799084 CANDIDATE_BALLOT -> LEADER_REPLAY
memberId=2 nowMs=1553872799084 LEADER_REPLAY -> LEADER_TRANSITION
becomeLeader: leadershipTermId=0
memberId=2 nowMs=1553872799093 LEADER_TRANSITION -> LEADER_READY
memberId=1 nowMs=1553872799231 FOLLOWER_BALLOT -> FOLLOWER_REPLAY
memberId=1 nowMs=1553872799231 FOLLOWER_REPLAY -> FOLLOWER_CATCHUP_TRANSITION
memberId=0 nowMs=1553872799233 FOLLOWER_BALLOT -> FOLLOWER_REPLAY
memberId=0 nowMs=1553872799233 FOLLOWER_REPLAY -> FOLLOWER_CATCHUP_TRANSITION
memberStatusPublisher.catchupPosition (Sending catchup position): leadershipTermId=0, logPosition=0, followerMemberId=1
memberId=1 nowMs=1553872799231 FOLLOWER_CATCHUP_TRANSITION -> FOLLOWER_CATCHUP
memberStatusPublisher.catchupPosition (Sending catchup position): leadershipTermId=0, logPosition=0, followerMemberId=0
memberId=0 nowMs=1553872799233 FOLLOWER_CATCHUP_TRANSITION -> FOLLOWER_CATCHUP
onCatchupPosition: termId=0, logPosition=0, followerMemberId=0
onCatchupPosition: startReplay
memberId=0 nowMs=1553872799276 FOLLOWER_CATCHUP -> FOLLOWER_TRANSITION
memberId=0 nowMs=1553872799284 FOLLOWER_TRANSITION -> FOLLOWER_READY
onCatchupPosition: termId=0, logPosition=0, followerMemberId=1
onCatchupPosition: startReplay
memberId=1 nowMs=1553872799292 FOLLOWER_CATCHUP -> FOLLOWER_TRANSITION
memberId=1 nowMs=1553872799303 FOLLOWER_TRANSITION -> FOLLOWER_READY
Restarting...
memberId=2 nowMs=1553872802092 INIT -> CANVASS
memberId=0 nowMs=1553872802133 INIT -> CANVASS
memberId=1 nowMs=1553872802173 INIT -> CANVASS
memberId=2 nowMs=1553872802175 CANVASS -> NOMINATE
memberId=2 nowMs=1553872802213 NOMINATE -> CANDIDATE_BALLOT
memberId=0 nowMs=1553872802215 CANVASS -> FOLLOWER_BALLOT
memberId=1 nowMs=1553872802215 CANVASS -> FOLLOWER_BALLOT
memberId=2 nowMs=1553872802215 CANDIDATE_BALLOT -> LEADER_REPLAY
memberId=0 nowMs=1553872802374 FOLLOWER_BALLOT -> FOLLOWER_REPLAY
memberId=0 nowMs=1553872802374 FOLLOWER_REPLAY -> FOLLOWER_CATCHUP_TRANSITION
memberId=1 nowMs=1553872802374 FOLLOWER_BALLOT -> FOLLOWER_REPLAY
memberId=1 nowMs=1553872802374 FOLLOWER_REPLAY -> FOLLOWER_CATCHUP_TRANSITION
memberStatusPublisher.catchupPosition (Sending catchup position): leadershipTermId=1, logPosition=0, followerMemberId=0
memberId=0 nowMs=1553872802374 FOLLOWER_CATCHUP_TRANSITION -> FOLLOWER_CATCHUP
memberStatusPublisher.catchupPosition (Sending catchup position): leadershipTermId=1, logPosition=0, followerMemberId=1
memberId=1 nowMs=1553872802374 FOLLOWER_CATCHUP_TRANSITION -> FOLLOWER_CATCHUP
dropping! onCatchupPosition: termId=1, logPosition=0, followerMemberId=0, expectedTermId=0
dropping! onCatchupPosition: termId=1, logPosition=0, followerMemberId=1, expectedTermId=0
memberId=2 nowMs=1553872802701 LEADER_REPLAY -> LEADER_TRANSITION
becomeLeader: leadershipTermId=1
memberId=2 nowMs=1553872802704 LEADER_TRANSITION -> LEADER_READY
timeout waiting for agent 'clustered-service' to close, retrying...
close failed for agent 'aeron-client-conductor' due to InterruptedException
timeout waiting for agent 'clustered-service' to close, retrying...
close failed for agent 'aeron-client-conductor' due to InterruptedException

java.lang.AssertionError: 
Expected: is null
     but: was <FOLLOWER_CATCHUP>
Expected :is null
Actual   :<FOLLOWER_CATCHUP>
```

The key bit being:

```
dropping! onCatchupPosition: termId=1, logPosition=0, followerMemberId=0, expectedTermId=0
dropping! onCatchupPosition: termId=1, logPosition=0, followerMemberId=1, expectedTermId=0
```

### DIFF
--- a/aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModuleAgent.java
+++ b/aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModuleAgent.java
@@ -527,8 +527,16 @@ class ConsensusModuleAgent implements Agent, MemberStatusListener
 
     public void onCatchupPosition(final long leadershipTermId, final long logPosition, final int followerMemberId)
     {
+        if (leadershipTermId != this.leadershipTermId) {
+            System.out.println("dropping! onCatchupPosition: termId=" + leadershipTermId + ", logPosition=" + logPosition +
+                ", followerMemberId=" + followerMemberId + ", expectedTermId=" + this.leadershipTermId);
+        }
+
         if (Cluster.Role.LEADER == role && leadershipTermId == this.leadershipTermId)
         {
+            System.out.println("onCatchupPosition: termId=" + leadershipTermId + ", logPosition=" + logPosition +
+                ", followerMemberId=" + followerMemberId);
+
             final ClusterMember follower = clusterMemberByIdMap.get(followerMemberId);
 
             if (null != follower)
@@ -543,6 +551,8 @@ class ConsensusModuleAgent implements Agent, MemberStatusListener
 
                 if (follower.catchupReplaySessionId() == Aeron.NULL_VALUE)
                 {
+                    System.out.println("onCatchupPosition: startReplay");
+
                     follower.catchupReplaySessionId(archive.startReplay(
                         logRecordingId(), logPosition, Long.MAX_VALUE, replayChannel, ctx.logStreamId()));
                 }
@@ -1161,6 +1171,7 @@ class ConsensusModuleAgent implements Agent, MemberStatusListener
 
     void becomeLeader(final long leadershipTermId, final long logPosition, final int logSessionId)
     {
+        System.out.println("becomeLeader: leadershipTermId=" + leadershipTermId);
         this.leadershipTermId = leadershipTermId;
 
         final ChannelUri channelUri = ChannelUri.parse(ctx.logChannel());
--- a/aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModuleAgent.java
+++ b/aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModuleAgent.java
@@ -527,16 +527,8 @@ class ConsensusModuleAgent implements Agent, MemberStatusListener
 
     public void onCatchupPosition(final long leadershipTermId, final long logPosition, final int followerMemberId)
     {
-        if (leadershipTermId != this.leadershipTermId) {
-            System.out.println("dropping! onCatchupPosition: termId=" + leadershipTermId + ", logPosition=" + logPosition +
-                ", followerMemberId=" + followerMemberId + ", expectedTermId=" + this.leadershipTermId);
-        }
-
         if (Cluster.Role.LEADER == role && leadershipTermId == this.leadershipTermId)
         {
-            System.out.println("onCatchupPosition: termId=" + leadershipTermId + ", logPosition=" + logPosition +
-                ", followerMemberId=" + followerMemberId);
-
             final ClusterMember follower = clusterMemberByIdMap.get(followerMemberId);
 
             if (null != follower)
@@ -551,8 +543,6 @@ class ConsensusModuleAgent implements Agent, MemberStatusListener
 
                 if (follower.catchupReplaySessionId() == Aeron.NULL_VALUE)
                 {
-                    System.out.println("onCatchupPosition: startReplay");
-
                     follower.catchupReplaySessionId(archive.startReplay(
                         logRecordingId(), logPosition, Long.MAX_VALUE, replayChannel, ctx.logStreamId()));
                 }
@@ -1171,7 +1161,6 @@ class ConsensusModuleAgent implements Agent, MemberStatusListener
 
     void becomeLeader(final long leadershipTermId, final long logPosition, final int logSessionId)
     {
-        System.out.println("becomeLeader: leadershipTermId=" + leadershipTermId);
         this.leadershipTermId = leadershipTermId;
 
         final ChannelUri channelUri = ChannelUri.parse(ctx.logChannel());
--- a/aeron-cluster/src/main/java/io/aeron/cluster/Election.java
+++ b/aeron-cluster/src/main/java/io/aeron/cluster/Election.java
@@ -207,57 +207,70 @@ public class Election implements AutoCloseable
     int doWork(final long nowMs)
     {
         int workCount = State.INIT == state ? init(nowMs) : 0;
-        workCount += memberStatusAdapter.poll();
 
         try
         {
             switch (state)
             {
                 case CANVASS:
+                    workCount += memberStatusAdapter.poll();
                     workCount += canvass(nowMs);
                     break;
 
                 case NOMINATE:
+                    workCount += memberStatusAdapter.poll();
                     workCount += nominate(nowMs);
                     break;
 
                 case CANDIDATE_BALLOT:
+                    workCount += memberStatusAdapter.poll();
                     workCount += candidateBallot(nowMs);
                     break;
 
                 case FOLLOWER_BALLOT:
+                    workCount += memberStatusAdapter.poll();
                     workCount += followerBallot(nowMs);
                     break;
 
                 case LEADER_REPLAY:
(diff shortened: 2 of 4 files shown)


tmontgomery:
Just a heads up. Martin has been making some changes in this area and I have a potential fix that we've been seeing with `ClusterTest.shouldCatchupFromEmptyLog` which is quite similar. Part of these changes handle changing the bounded position for services consuming replays as well as some replay EOS changes. Also, some additional checks added in various places.

mjpt777:
With the changes this week I'm seeing tests be very stable. Worth testing for yourself with the latest build.

mjpt777:
I found your test does not work on Windows as it is trying to delete files which are still open.

ZachBray:
Sorry about that, I've fixed the test on Windows.

The test still fails until I apply my fix on top.

I believe there is still a problem where the leader processes catch-up positions before `LEADER_READY`, i.e., whilst in `LEADER_CATCHUP`.

```
    public void onCatchupPosition(final long leadershipTermId, final long logPosition, final int followerMemberId)
    {
        if (Cluster.Role.LEADER == role && leadershipTermId == this.leadershipTermId)
        {
            // ...
        }
    }

```

The consensus module agent drops those catch-up requests, as during `LEADER_REPLAY` the role is not `LEADER` and the `leadershipTermId` has not yet been incremented.

This only seems to happen with "fresh" or "clean" follower nodes; otherwise, the time the followers take to replay their own logs seems to be enough time for the leader to have transitioned out of `LEADER_REPLAY` and `LEADER_TRANSITION` into `LEADER_READY`.

mjpt777:
Member status messages need to be processed in all states in order to detect new election and leaders, so your proposed changes are not suitable. The election will need to be modified to cope with the situation outlined.

ZachBray:
OK yep that invalidates my assumption above.

> It is safe to defer the processing of member status events

I had assumed that it would be safe to defer as I believed `LEADER_REPLAY` was a "short lived" state.

What I don't understand about the safety is that there is a race anyway. Deferring could happen naturally if the clustered service container were to stall during `LEADER_REPLAY`. If this isn't safe does this indicate another issue or am I misunderstanding something? 

I can raise a PR with just the test if that is useful.

tmontgomery:
I've made a change in ffab7bb48 that might help some here. Specifically, catchupPosition might be dropped for a couple reasons. One being role and timing. But it could have other reasons. So, to cover that, I've added a change to have the follower send catchupPosition periodically until catchup is done and transitioned. This is cleaner in general and needed anyway.

@ZachBray feel free to try out this change and let us know how it goes.

mjpt777:
After the change by @tmontgomery I've merged in the test which has been running in a loop without failing for me.

PR #658, opened 2019-04-19
Rescheduled timer test

zyulyaev:


mjpt777:
Thanks. Point of process, best to ignore a test so that it does not fail the build until fixed.

PR #664, opened 2019-05-03
Trigger scans when run on CI

aalmiray:
Configure build scans to be triggered when running on CI


aalmiray:
Thanks! Is it OK if I send you a similar PR for SBE?

mjpt777:
Sure

PR #666, opened 2019-05-12
Prevent creating multiple Clients or Media Drivers with the same cont…

mikeb01:
…ext.  Allowing it with clients cause a SEGV, instead now will throw IllegalStateException.  This has tripped me up twice now.

### DIFF
--- a/aeron-client/src/main/java/io/aeron/Aeron.java
+++ b/aeron-client/src/main/java/io/aeron/Aeron.java
@@ -123,6 +123,8 @@ public class Aeron implements AutoCloseable
      */
     public static Aeron connect(final Context ctx)
     {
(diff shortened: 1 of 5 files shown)


mikeb01:
I notice that you moved the implementation inside of conclude.  This is what I had originally, but I found that this implementation still segfaults as the the `Aeron.connect` constructs the Aeron instance inside of a try/catch block that closes the reused context on failure.

mjpt777:
Good point. Maybe a different exception type that does not result in close being called.

mikeb01:
I'll have a look today.

mjpt777:
I pushed a change that does this.

mikeb01:
Looks good.  Added another small PR that includes the unit tests.

PR #672, opened 2019-05-29
Treat back-pressure as a successful keep-alive signal

cowwoc:
We shouldn't disconnect from the cluster while we're actively sending it data.

### DIFF
--- a/aeron-cluster/src/main/java/io/aeron/cluster/client/AeronCluster.java
+++ b/aeron-cluster/src/main/java/io/aeron/cluster/client/AeronCluster.java
@@ -403,7 +403,7 @@ public final class AeronCluster implements AutoCloseable
         while (true)
         {
             final long result = publication.offer(keepaliveMsgBuffer, 0, keepaliveMsgBuffer.capacity(), null);
-            if (result > 0)
+            if (result > 0 || result == Publication.BACK_PRESSURED)
             {
                 return true;
             }


cowwoc:
To clarify, this PR is meant to solve a problem we're seeing in production where a client is so busy sending data to the cluster that it ends up starving the keep-alive mechanism. When the user sees the keep-alive fail, they disconnect the client thinking that network connectivity is down but in reality connection is alive and well.

I propose treating back-pressure as an indication that network data is actively being sent and equivalent to a successful keep alive.

cowwoc:
Not sure why Github thinks Travis tests are still running. You can see they have completed successfully if you click `Details`.

mjpt777:
This does not feel like an appropriate solution. Failure to send should not result in the client thinking the network connectivity is down. The client should check the ingress publication or egress subscription for connectivity.

tmontgomery:
Agree with @mjpt777

cowwoc:
Thanks for the feedback. Which methods do you recommend I look at? I saw `Publication.isConnected()` but it's not clear what this actually checks for. Will this detect if the line is idle?

mjpt777:
BTW the cluster treats data ingress as a valid keep alive so you don't need to set keep alives when data is flowing. The fact that you are getting back pressure is what needs investigated. If this is a blip then buffer sizing needs considered. If a more consistent state that indicates the service is not fast enough and needs to be profiled and improved.

mjpt777:
I'm closing this PR and further discussion can occur on the issue as appropriate.

cowwoc:
@mjpt777 Not sure whether you saw my question above ...?

tmontgomery:
`Publication.isConnected` checks that there is an active receiver. The progress of the receiver is a separate concern.

PR #674, opened 2019-06-06
ClusterException test: invalid service ACK state SNAPSHOT reproduction case

ratcashdev:
Two test cases: one that works (with a custom service) and one that doesn't.
The only difference is that in the second case I request an extra snapshot to be created in addition to the one created during shutdown.

The TC also confirms, that if a followers loses state, it does a replay from the start of times (not from a snapshot). Not sure if this is intended, though.


mjpt777:
If submitting failing tests please mark them with @Ignore so they can be picked up later.

ratcashdev:
OK, will do shortly.

mjpt777:
I've added the ability to provide a supplier for test services and awaiting the neutral control toggle. The tests here are more complicated than the issue you are trying to point out, e.g. I don't see why you are reconnecting the client. Can you reduce the test cases based on the changes and try create the minimum example to reproduce?

ratcashdev:
I have updated the test case to use the supplier.
As for the reconnect: indeed it was not necessary in the first case (and I stripped out as much as possible from the test cases while leaving the essence), but it is important in the second one because it shows NODE 2 as nonoperational (the exception thrown by the cluster is processed by the errorHandler and the context is not reachable to set a different error handler)

mjpt777:
I've fix the main point here and a few related issues.

On clarifying, replaying should not be from latest snapshot as our clients need to rebuild down-stream state and thus play forward. If you are prone to ops people deleting disks then use a traditional disk backup or dynamic membership. We may offer a tool to fast recover from the cluster much like dynamic membership as an extension of `ClusterTool` if our clients require it.

ratcashdev:
> I've fix the main point here and a few related issues.

That was quick - although I see it wasn't a simple one.

> On clarifying, replaying should not be from latest snapshot as our clients need to rebuild down-stream state and thus play forward. If you are prone to ops people deleting disks then use a traditional disk backup or dynamic membership. We may offer a tool to fast recover from the cluster much like dynamic membership as an extension of `ClusterTool` if our clients require it.

Got it. I perfectly understand the need for having an identical and complete journal/state on every single machine, even after a disk crash. Just wondering how sustainable that is, if, say, the journal is already covering several months or years? Though I guess once that becomes an issue for any of your clients, there will be a suitable solution in place.
That also means that the test case in question should be removed as invalid, shouldn't it? Or do you have other plans?

mjpt777:
> Got it. I perfectly understand the need for having an identical and complete journal/state on every single machine, even after a disk crash. Just wondering how sustainable that is, if, say, the journal is already covering several months or years? Though I guess once that becomes an issue for any of your clients, there will be a suitable solution in place. That also means that the test case in question should be removed as invalid, shouldn't it? Or do you have other plans?

To more quickly catch up after a significant history of logs people will backup from longer term storage. Financial regulations require 7 years of history and this will not be kept on active cluster machines as they are likely to have smaller fast storage such as NVMe SSDs whereas longer term storage would be some from of SAN/NAS or S3. Restore would happen from that. As mentioned above we will provide tooling for those clients which may or may not be open source.

PR #675, opened 2019-06-07
Set custom control address for a UDP multicast endpoint

khvorov:
This change allows to specify a custom control address for a multicast endpoint. For example, one can configure data & control groups by changing the third octet instead of the last one:

- Data group : 239.65.1.0/24
- Control group : 239.65.2.0/24

That could simplify setup of network and firewall rules for the multicast traffic.

The control channel is configured in the URI via the **control** parameter:

_aeron:udp?endpoint=224.10.9.9:40124|**control=224.10.10.9:40124**_


tmontgomery:
I am inclined against this. It really only adds more potential for misconfiguration. But, we will consider it.

tmontgomery:
We don't see the value here. And the additional problems introduced via ease of misconfiguration, etc. are a serious disadvantage.

Thanks, but no. This is not something we want to use.

khvorov:
Sure, no problems. Thanks for reviewing it. Though I don't quite understand what misconfiguration could be introduced here. The control parameter is used in the same way like it is used for multi-destination cast. It is merely a fine tuning, the default behaviour (control address = data address+1) remains unchanged

tmontgomery:
In the unicast MDC case, data will not flow if not correct. In this case, if 2 different control channels used, but a valid data channel, data will flow, but no NAKs or SMs. Which will cause asymmetry hard to track down.

PR #685, opened 2019-07-10
Retain timer on log append backpressure

zyulyaev:
Fix a bug in which timer correlation id is lost in case of log append backpressure.


mjpt777:
Thanks

PR #690, opened 2019-07-27
Add OSGi manifest headers

io7m:
This change moves to using the biz.aQute.bnd Gradle plugin to
produce OSGi-compatible manifests in jar files. Currently, only the
aeron-client, aeron-driver, and aeron-cluster modules have been updated
to provide OSGi metadata because the other modules are samples, tests,
or should never need to be visible inside an OSGi container. Care
has been taken to preserve the existing defined manifest fields
(Automatic-Module-Name, Implementation-Title, etc).

Note that until Agrona 1.0.4 is released (or at least a version containing
https://github.com/real-logic/agrona/commit/49e769e07c3ba3fa176d3b1079f2472184612e1d),
the generated package imports for Agrona will be unversioned. This will
self-correct when Agrona 1.0.4+ is released and Aeron is updated to
point to the new version.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -24,6 +24,7 @@ plugins {
     id 'com.gradle.build-scan' version '2.3'
     id "io.freefair.javadoc-links" version "3.8.1" apply false
     id 'com.github.johnrengelman.shadow' version '5.1.0' apply false
+    id 'biz.aQute.bnd.builder' version '4.2.0' apply false
 }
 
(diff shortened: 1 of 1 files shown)


io7m:
Please consider, if you have an Aeron release coming up, releasing Agrona first and then pointing Aeron at the new release. :smile:

mjpt777:
Thanks. Will do. We will release Agrona early this week. I need it for some other changes.

PR #701, opened 2019-08-22
Tests for potential issues in Aeron cluster

marion35700:
* 1st issue: with a 3-node cluster, we simulate a loss of traffic from the leader to a follower for some time, then when the traffic is back to normal, the election state of the follower remains stuck in FOLLOWER_CATCHUP_TRANSITION.

*2nd issue: with a 3-node cluster + a connected client, stop follower1, stop follower2, restart follower2, then the cluster never recovers (no leader)

@JPWatson @ZachBray


tmontgomery:
I've brought in the second test and made a change that, I think, addresses what I see from it. e8a7d5975

The first test has a number of oddities, but also has raised a number of things to look at. So, took a while to look through. First thing to mention is that the use of SendChannelEndpoint for loss creates some asymmetry in the loss behavior that causes a number of odd things to happen. One central one is that loss recovery and catchup can be extended due to losing the image due to timeout. The asymmetries are: log channel uses manual MDC and thus no loss is seen on the log channel, the SMs/NAKs coming into the leader are dropped as well as data going out. This creates a number of delays in recovering correctly. In testing this, it seemed that given a little bit more time, the catchup finishes. It just might go through a couple rounds to finally do it.

I'm going to look a little bit more into this, though. Some of the states I've seen the catchup node get into are worthy of investigation. But, as of right now, I can't see anything definitely incorrect.

PR #711, opened 2019-09-25
Re-election issue test example

zyulyaev:
This is a test to reproduce an issue during leader re-election.
The test passes, but logs an error.
The PR is not intended to be merged, but is created only to highlight the problem.

@tmontgomery please take a look.


tmontgomery:
Adjusting the clock like this test does have some odd behaviors. But, it is not the main cause of the issue. I've made a change that might be good to test with. e8a7d5975

zyulyaev:
Adjusting the clock seemed to be the easiest way to reproduce the problem. 
The change helps, thank you!

PR #712, opened 2019-10-06
[C] Add a termination validator to `aeronmd`

bspeice:
Had some confusion during local testing; while a termination hook is registered in `aeronmd`, the C++/Java `DriverTool` programs aren't actually able to terminate it. Adding a validator that just returns `true` allows proper shutdown.


tmontgomery:
This is not how this is done. By default the driver should not allow termination.

There is a function to set the validator to via configuration. Setting the env var AERON_DRIVER_TERMINATION_VALIDATOR to aeron_driver_termination_validator_default_allow

bspeice:
Got it, thanks for the clarification!

PR #718, opened 2019-10-20
Narrow Aeron constructor to private visibility

TisonKun:
We don't use it outside the file and generally speaking we use static factory methods for instancing Aeron object.

Correct me if there is any concern we intend to make it package-private.

### DIFF
--- a/aeron-client/src/main/java/io/aeron/Aeron.java
+++ b/aeron-client/src/main/java/io/aeron/Aeron.java
@@ -78,7 +78,7 @@ public class Aeron implements AutoCloseable
     private final AgentRunner conductorRunner;
     private final Context ctx;
 
-    Aeron(final Context ctx)
+    private Aeron(final Context ctx)
     {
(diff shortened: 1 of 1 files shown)


mjpt777:
I don't see the point to this change and it fails the build.

TisonKun:
@mjpt777 build fails because

```
> Task :aeron-client:checkstyleMain
[ant:checkstyle] [ERROR] /home/travis/build/real-logic/aeron/aeron-client/src/main/java/io/aeron/Aeron.java:61: Class Aeron should be declared as final. [FinalClass]
```

currently this class is not declared as final also. I think it is a checkstyle issue we can address.

For the point for this change, if we don't instance `Aeron` outside the file and it is already a non-public constructor, we can make it a private constructor.

mjpt777:
If the class is made final then it is more difficult to mock.

TisonKun:
Thanks for your explanation.

PR #719, opened 2019-10-21
[C++] Remove original registration ID from ExclusivePublication.

denizevrenci:
As exclusive publications are guaranteed to be original.

### DIFF
--- a/aeron-client/src/main/cpp/ClientConductor.cpp
+++ b/aeron-client/src/main/cpp/ClientConductor.cpp
@@ -16,6 +16,8 @@
 
 #include "ClientConductor.h"
 
+#include <cassert>
+
 namespace aeron {
 
 template<typename T, typename... U>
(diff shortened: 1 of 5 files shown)


mjpt777:
It is best leaving this in place as the publication API can be used interchangeable. What is the rationale for removing it?

denizevrenci:
Just code cleanup and maybe very tiny perf gain. I kept the public API fully compatible. Though, I am not sure if conductor's `onNewExclusivePublication` callback is used by someone.

denizevrenci:
Also wanted to confirm this statement.
> As exclusive publications are guaranteed to be original.

mjpt777:
Exclusive publications will always be original. 

The change I'm OK with except for the change to the `onNewExclusivePublication` callback which is breaking. Maybe @tmontgomery has a view on the value of the callbacks for publications?

tmontgomery:
Is it possible to keep the same callback and just not save the original registration ID?

denizevrenci:
I've reverted the change on `onNewExclusivePublication` leaving the constructor of `ExclusivePublication` unchanged from the initial commit.

mjpt777:
The test does not compile.

denizevrenci:
I had reverted the call to the `ExclusivePublisher` constructor in the test. That has been fixed.

PR #726, opened 2019-11-04
[C++] Fix narrowing conversion in client

StephanDollberg:
MSCV is correctly complaining about the narrowing conversion from
`std::size_t` to `int` in `AtomicArrayUpdater::removeElement`.

This patch makes us return a `std::size_t` instead. Using `-1` to
signify an invalid length is not needed as the length value is not used
anyway if the pointer is null. Hence we use `0` as the invalid value and
avoid the narrowing conversion.


tmontgomery:
I restarted the failed build. Once it comes back OK, will merge.

PR #727, opened 2019-11-05
[Java] Test cluster restart with tombstoned snapshot

zyulyaev:
There is an issue with new tombstone latest snapshot command. The following steps leads to corrupted cluster state:
1) Send some messages to a cluster;
2) Take a snapshot;
3) Send some more messages;
4) Stop the cluster;
5) Invoke tombstone latest snapshot command;
6) Restart the cluster;
7) Send some messages to the cluster;
8) Restart the cluster.

This PR adds a test with these steps.


mjpt777:
I pushed a fix for this and merged a reworked version of the test. In future it is easier to merge a PR with the test ignored so the build does not fail.

zyulyaev:
Okay, thank you.

PR #730, opened 2019-11-08
[Java] Test on truncate recording issue

zyulyaev:
Recording cannot be truncated to the start position if it is greater than term-length.
This PR adds a test reproducing this issue.


mjpt777:
This fails checkstyle so will close. I fixed the calculation.

PR #732, opened 2019-11-12
[C++] add explicit cast to prevent warnings.

mikeb01:


aeron-client/src/main/cpp/util/Index.h
```diff
@@ -29,7 +29,7 @@ typedef std::int32_t index_t;
 
 inline static index_t convertSizeToIndex(size_t size)
 {
-    if (size > std::numeric_limits<index_t>::max())
+    if (size > (size_t) std::numeric_limits<index_t>::max())
```
mjpt777: Would be better with a C++ rather than C style cast.

mikeb01: Would it also make sense to have something like:
```c++
static_assert (sizeof(std::int32_t) <= sizeof(size_t));
```
In the header, in case someone built it on some platform where an unsigned long int was less than 32 bit?

PR #738, opened 2019-11-15
Remove commented out println statement

jamesasefa:
Small change removing commented out println from `ReplayMerge`


mjpt777:
We like to use these for debugging so will not be removed.

PR #741, opened 2019-11-18
Error Counter Type IDs

JPWatson:
Because the 'id' field is not set for these, they clash with "Bytes Sent" when read (id=0).
Simplest strategy seemed to be to give them their own types.


tmontgomery:
I think we might want to consider consolidating the Aeron counter types into fewer places. Maybe 1 for Aeron base counter types beyond System Counters. 1 for Archive and 1 for Cluster. That way it is easier to track IDs as well as easier to find.

PR #742, opened 2019-11-18
Support specifying session-id on publications in the C driver

mikeb01:
Implements:
- session-id parsing on the URI
- reserved range for publication session-ids
- validation of mtu and term lengths for matching publications
- prevention of re-using specified session-ids when using system generated ones


aeron-driver/src/main/c/aeron_driver_conductor.c
```diff
@@ -315,6 +322,88 @@ bool aeron_client_has_reached_end_of_life(aeron_driver_conductor_t *conductor, a
...
 }
 
+int32_t aeron_driver_conductor_next_session_id(aeron_driver_conductor_t* conductor)
+{
+    const int32_t low = conductor->publication_reserved_session_id_low;
+    const int32_t high = conductor->publication_reserved_session_id_high;
+
+    return
+        low <= conductor->next_session_id && conductor->next_session_id <= high
+            ? high + 1
```
mjpt777: Operator at the end of the line for existing format match.

aeron-driver/src/main/c/aeron_driver_conductor.c
```diff
@@ -315,6 +322,88 @@ bool aeron_client_has_reached_end_of_life(aeron_driver_conductor_t *conductor, a
...
+int aeron_confirm_publication_match(
+    const char *uri, int32_t stream_id, aeron_uri_publication_params_t *params, int32_t existing_session_id,
+    aeron_logbuffer_metadata_t *logbuffer_metadata)
+{
+    if (params->has_session_id && params->session_id != existing_session_id)
+    {
+        aeron_set_err(
+            EINVAL,
+            "Existing session-id [%d] doesn't match specified channel: %s, stream-id: %" PRId32,
+            existing_session_id, uri, stream_id);
```
mjpt777: channel uri from client may not be null terminated, length must also be considered.

aeron-driver/src/main/c/aeron_driver_conductor.c
```diff
@@ -739,17 +871,19 @@ aeron_ipc_publication_t *aeron_driver_conductor_get_or_add_ipc_publication(
...
+                int32_t session_id = params->has_session_id
+                    ? params->session_id
+                    : aeron_driver_conductor_update_next_session_id(conductor, speculated_session_id);
                 int32_t initial_term_id = params->has_position ? params->initial_term_id : aeron_randomised_int32();
                 aeron_position_t pub_pos_position;
                 aeron_position_t pub_lmt_position;
 
                 pub_pos_position.counter_id = aeron_counter_publisher_position_allocate(
-                    &conductor->counters_manager, registration_id, session_id, stream_id, channel_length, channel);
+                    &conductor->counters_manager, registration_id, session_id, stream_id, channel_length, uri);
```
mjpt777: channel => uri rename but not consistent with channel_length.

aeron-driver/src/main/c/collections/aeron_bit_set.h
```diff
@@ -0,0 +1,189 @@
...
+inline int aeron_bit_set_stack_alloc(
+    size_t bit_count,
+    uint64_t* static_array,
+    size_t static_array_len,
+    aeron_bit_set_t* bit_set)
+{
+    (bit_set)->bit_count = bit_count;
+    (bit_set)->static_array = static_array;
+
+    const size_t u64_len = ((bit_count + 63) / 64);
```
mjpt777: should name be bitset_length?

aeron-driver/src/test/c/collections/aeron_bit_set_test.cpp
```diff
@@ -0,0 +1,248 @@
+/*
+ * Copyright 2019 Real Logic Ltd.
```
mjpt777: Copyright is 2014-2019 on project.

aeron-driver/src/main/c/aeron_driver_conductor.c
```diff
@@ -315,6 +322,88 @@ bool aeron_client_has_reached_end_of_life(aeron_driver_conductor_t *conductor, a
...
+    {
+        return -1;
+    }
+
+    *session_id = aeron_driver_conductor_next_session_id(conductor) + (int) index;
+    return 0;
+}
+
+int aeron_confirm_publication_match(
+    const char *uri, int32_t stream_id, aeron_uri_publication_params_t *params, int32_t existing_session_id,
```
mjpt777: Formatting.

aeron-driver/src/main/c/collections/aeron_bit_set.c
```diff
@@ -0,0 +1,44 @@
...
+ * limitations under the License.
+ */
+
+#include <string.h>
+#include <errno.h>
+#include "aeron_bit_set.h"
+
+extern int aeron_bit_set_stack_alloc(
+    size_t bit_count,
+    uint64_t* static_array,
```
mjpt777: Inconsistent use of where the star is.

mjpt777:
Looking good so far Mike. Would be useful to provide an overview of how the bitset is used to track sessions in use.

aeron-driver/src/main/c/aeron_driver_conductor.c
mikeb01: Remove uri from the method and made the error strings match those in the Java driver.

aeron-driver/src/main/c/collections/aeron_bit_set.h
mikeb01: Went with bit_set_length to match the naming of the struct.

mikeb01:
Similar to the Java driver the aeron_driver_conductor tracks the next session id to be used.  When we add a publication we may need to add some delta to that value to ensure there is no collision with an existing session.  Typically this delta will be 0, and we can be certain that the delta will not be more than the total number of publications.

A bit set of length `publications.length + 1` is constructed (and initialised with all false values) and while iterating over the publications in order to find any matching active publications (an existing loop), the difference between the active publication's session id and the conductor's next session id is calculated.  This value is flagged as true in the bit set.  Once the search over the publications is complete, the bit set is scanned for the first false value.

We store this as the `speculated_session_id` as we search for it regardless of whether is is required or not.  Iff a new publication is required and a session id is not specified in the uri, the `speculated_session_id` is used and the conductor's next session id is updated to the `speculated_session_id + 1`.

Prior to the scan through the publications, when the next session id is fetched from the conductor, any reserved ranged will be skipped over.

I went with the naming of `speculated_session_id` as it is always calculated but only used if required.  That simplified the logic and reduced the amount of branching required.

aeron-driver/src/main/c/aeron_driver_conductor.c
```diff
@@ -315,6 +323,90 @@ bool aeron_client_has_reached_end_of_life(aeron_driver_conductor_t *conductor, a
...
 }
 
+int32_t aeron_driver_conductor_next_session_id(aeron_driver_conductor_t* conductor)
+{
+    const int32_t low = conductor->publication_reserved_session_id_low;
+    const int32_t high = conductor->publication_reserved_session_id_high;
+
+    return
+        low <= conductor->next_session_id && conductor->next_session_id <= high ?
+        aeron_add_wrap_i32(high, 1) :
```
mjpt777: Indent the wrap.

aeron-driver/src/main/c/aeron_driver_conductor.c
```diff
@@ -705,25 +797,67 @@ aeron_ipc_publication_t *aeron_driver_conductor_get_or_add_ipc_publication(
...
+        aeron_set_err(EINVAL, "(BUG) Unable to allocate session-id");
+        return NULL;
+    }
+
+    if (is_session_id_in_use && (is_exclusive || NULL == publication))
+    {
+        aeron_set_err(
+            EINVAL,
+            "Specified session-id is already in exclusive use for channel: %s, stream-id: %" PRId32,
+            uri, stream_id);
```
mjpt777: Need to account for channel length. Example is https://github.com/real-logic/aeron/blob/master/aeron-driver/src/main/c/aeron_position.c#L63

aeron-driver/src/main/c/collections/aeron_bit_set.h
```diff
@@ -0,0 +1,187 @@
...
+    aeron_bit_set_t *bit_set)
+{
+    (bit_set)->bit_set_length = bit_set_length;
+    (bit_set)->static_array = static_array;
+
+    const size_t u64_len = ((bit_set_length + 63) / 64);
+    if (NULL != static_array && u64_len <= static_array_len)
+    {
+        (bit_set)->bits = static_array;
+    } else
```
mjpt777: else on new line.

PR #745, opened 2019-11-18
[C++] Tidy up ClientConductor.

denizevrenci:


mjpt777:
This fails the build due to usage in archive client.

denizevrenci:
Will push a fix.

PR #746, opened 2019-11-18
[Java] Fixed dynamic node joins single static node

artem-v:
**Given:**
Start single static node `S`. Start dynamic node `D`, and point it to `S`. See discontinued exceptions in the attached logs. Looks like dynamic node decided that it is LEADER. **NOTE: this is not reproduced when there're more than 1 static node when dynamic node is joining a cluster.

**Version**
`1.23.1`

**Attached Logs**

[Static_Node_Output.txt](https://github.com/real-logic/aeron/files/3859616/Static_Node_Output.txt)
[Dynamic_Node_Output.txt](https://github.com/real-logic/aeron/files/3859617/Dynamic_Node_Output.txt)


mjpt777:
The issue with this was that the election was not expecting a dynamic member entering when the cluster size is 1. Dynamic members need to enter an election to catchup and then they can join as full members.

PR #748, opened 2019-11-20
[Java] Set a default error handler in Archive context when it is not supplied.

denizevrenci:
While it is intended to run the Archive module independently of the MediaDriver, in its current implementation, it fails due to no default error handler being supplied.

I tried to come up with a simple default here which just prints the stack trace based on the one in the client minus the special handling of timeout exceptions. I think it is possible to pipe stdout from Archive module to a file to provide simple logging.

### DIFF
--- a/aeron-archive/src/main/java/io/aeron/archive/Archive.java
+++ b/aeron-archive/src/main/java/io/aeron/archive/Archive.java
@@ -348,6 +348,19 @@ public class Archive implements AutoCloseable
         public static final int ARCHIVE_ERROR_COUNT_TYPE_ID = 101;
 
         /**
+         * The Default handler for Aeron runtime exceptions.
+         * <p>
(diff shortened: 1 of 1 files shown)


mjpt777:
This approach is a bit simplistic. The expectation is a distinct error log would be used which is then owned by the archive when it is not paired with a media driver. I'll look at a more complete solution.

denizevrenci:
Before settling on this, I took a look around the codebase to implement a `LoggingErrorHandler` as the default but it seems like it needs a buffer backed by a file. So I stopped short of trying to implement it before getting some opinions.

mjpt777:
I pushed an approach which allows the error log to be in the archive mark file like it is in the cluster components.

denizevrenci:
Thank you for the quick resolution, this is what I initially intended to implement but gave up due to a version change required in the archive catalog.

PR #749, opened 2019-11-20
[Java] Add RESOLVE_HOSTNAMES command to resolve hostnames into ip when needed and tests

lightZebra:
There is an issue when working with aeron in cloud(kubernetes for example).
Node wouldn't reconnect to cluster after went down and up with different IP but same hostname:port.
Because hostnames resolved into IP once on startup(creation) and reused most of the time.

I added new command RESOLVE_HOSTNAMES which we could send to media driver to resolve hostnames and apply only new IP.

Scenario:
1. Start 3 static nodes
2. For some reason one node went down and up with different IP but same hostname:port and data written to disk(still can't join cluster).
3. Using DriverTool CLI(or programmatically) execute "resolveHostname" command for 2 other nodes(also we could execute command for each node, result will be the same)
4. Hostnames with new IP will be updated, old IP stay as it is
5. Cluster working again with 3 nodes.

Added tests and tested using docker-compose and kubernetes.

### DIFF
--- a/aeron-agent/src/main/java/io/aeron/agent/DriverEventEncoder.java
+++ b/aeron-agent/src/main/java/io/aeron/agent/DriverEventEncoder.java
@@ -18,6 +18,7 @@ package io.aeron.agent;
 import org.agrona.DirectBuffer;
 import org.agrona.MutableDirectBuffer;
 
+import java.net.InetAddress;
 import java.net.InetSocketAddress;
 import java.nio.ByteBuffer;
 
@@ -33,6 +34,7 @@ public class DriverEventEncoder
 {
     private static final int LOG_HEADER_LENGTH = 16;
     private static final int SOCKET_ADDRESS_MAX_LENGTH = 24;
+    private static final byte[] EMPTY_BYTES = {};
 
     public static int encode(
         final MutableDirectBuffer encodingBuffer, final DirectBuffer buffer, final int offset, final int length)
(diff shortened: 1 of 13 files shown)


mjpt777:
This is a good start but it is incomplete. When IPs change then the canonical form for URIs will change. If they are not updated then latter adds for pubs and subs will not match with the ones registered before names are re-resolved.

lightZebra:
Thank you for the feedback. I'm glad that I'm on the right track and will continue working on this.

mjpt777:
This fails the build on Java 11.

https://travis-ci.org/real-logic/aeron/jobs/615559707?utm_medium=notification&utm_source=github_status

lightZebra:
I have fixed the issue with canonical form and added tests for send and receive channels.
Unfortunately I haven't found better way to change hostname resolution on a fly than reflection.

Also, it seems like `onResolveHostnames` method should update `PublicationImage.imageConnections` because we have `InetSocketAddress` inside which will be used as a send address.
But cluster seems working even with current changes.
Am I missing something, or we don't need to update `PublicationImage.imageConnections` right now?

lightZebra:
Did someone get a chance to look at latest changes? Do we need something more to merge this?

tmontgomery:
This is not the direction we want to go with supporting Kubernetes and cloud for re-resolution.

We have a more holistic approach in mind that we've been looking at that is automatic and not necessarily tied to the underlying name resolution.

However, if OK, let's keep this open for now as some of this might be very useful.

lightZebra:
Last update, might be useful
Added automatic hostname resolution when `timeNs > statusMessageDeadlineNs` with delay between resolutions. Removed resolve command.

I hope this version somehow correlate with your approach. I see latest solution as good enough, at least good point to start from, so people could start using it in near time.

Let me know is it good enough for you? 
Anyway you could leave it open if you need, it's OK for me.
If you don't have any solution right now and don't have plans to add it soon then maybe we could use this one for now, is it possible?

mjpt777:
A better approach has been implemented in Aeron for re-resolution.

PR #751, opened 2019-11-21
[Java] Add ClusterTool#addMember

segabriel:
I have a use case when some static nodes drop down and it never will start again on the same IP address, at that time its data volume is still accessible so I’d like to use the kept volume to launch a node on another IP address. And I need to launch the node in static mode due to the node publishes outcome into archive recordings, it's kinda stateful service. The issue is how to notice other cluster members about it and successfully join them. 
I looked at the `ClusterTool` and it has already the ability to remove a normal static member and its IP address from the cluster members, so I tried to add an opportunity to add a new cluster member as a static member. It seems working well. And after that, I can remove the old member, which will never come back, and then add a static member with a new IP address, and finally start node with the old volume on added member IP address.

So could it be an acceptable solution?

### DIFF
--- a/aeron-cluster/src/main/java/io/aeron/cluster/ClusterTool.java
+++ b/aeron-cluster/src/main/java/io/aeron/cluster/ClusterTool.java
@@ -720,6 +720,7 @@ public class ClusterTool
         out.println(
             "  tombstone-latest-snapshot: Mark the latest snapshot as a tombstone so previous is loaded.");
         out.println(
-            "                 add-member: [memberId] [memberEndpoints] requests adding a member by its memberId and endpoints.");
+            "                 add-member: [memberId] [memberEndpoints] requests adding a member by its memberId and " +
+            "endpoints.");
     }
 }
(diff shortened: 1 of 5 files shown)


mjpt777:
Please ensure the build passes before asking for review.

segabriel:
@mjpt777
My apologies, I corrected checkstyle.

tmontgomery:
The concept of this breaks many requirements of the cluster for safety. While the usage may be safe, the means is not safe for general usage.

segabriel:
@tmontgomery 
Yes, it is, and remove-member invoking is not safe as well. Currently, I don't see a way how to replace the IP address for a forever dyed node and launch another one with old data volume.

tmontgomery:
Remove member is safe as it goes through the required steps for members being removed. The requirements have to do with positions of append and commit and cluster operation.

We are aware of the name resolution issues and we are looking for a complete solution.

segabriel:
OK, thanks.

PR #753, opened 2019-11-23
Use StopPosition as terminating condition for snapshot transfer

JPWatson:
(Currently using synchronous call to get stop position, looking at making that async)

### DIFF
--- a/aeron-cluster/src/main/java/io/aeron/cluster/ClusterBackupAgent.java
+++ b/aeron-cluster/src/main/java/io/aeron/cluster/ClusterBackupAgent.java
@@ -25,10 +25,7 @@ import io.aeron.cluster.client.AeronCluster;
 import io.aeron.cluster.client.ClusterException;
 import io.aeron.cluster.codecs.BackupResponseDecoder;
 import io.aeron.cluster.codecs.MessageHeaderDecoder;
-import io.aeron.cluster.codecs.SnapshotMarkerDecoder;
 import io.aeron.cluster.service.ClusterMarkFile;
-import io.aeron.cluster.service.ClusteredServiceContainer;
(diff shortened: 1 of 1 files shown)


tmontgomery:
What about putting the snapshot length in the backup response message?

JPWatson:
Think that would probably require a breaking change to the recording.log to include the stop position.

mjpt777:
Also looking this up would cause a pause to the consensus module. Use replicate could be a good compromise.

tmontgomery:
The snapshot length is a consequence of loading or recording a snapshot. So, doesn't need to be in the recording log. Nor does it require it also be queried when responding.... as it can be saved when taking or loading a snapshot.

The only incompatibility is in the Query response messages. And as if you are running ClusterBackup then you need to be on the same version. No reason not to be current.

tmontgomery:
BTW, I've now confirmed this also happens with Dynamic Join as the same logic error is there.

tmontgomery:
It is hard to tell from the diff as so much is moved around what is actually added here.

A blocking archive query will not play well with the activity checking that has been added.

JPWatson:
In that case, there's also a blocking call to stopRecording.

tmontgomery:
It's a local call to backupArchive and not a network call. We tend to not treat them as non-blocking.... even though we should.

JPWatson:
Fair enough.

tmontgomery:
I have an async version stashed. Don't have time to get it in today and port to DynamicJoin and also to add a test with a larger snapshot before release.

But, will do so tonight and/or tomorrow.

tmontgomery:
With 6d4d064f43485660f276d51a202f32910606a075 , this is no longer needed.

PR #768, opened 2019-12-04
Recreate cluster member's publications by hostname when it is in the unconnected state

segabriel:
It is the next iteration of https://github.com/real-logic/aeron/pull/751 and I need still a solution for it.
I have seen https://github.com/real-logic/aeron/pull/749 and it partially resolves the problem. And this PR is an attempt to make the hostname resolving automatically. 
So on each slow tick, the `ConsensusModuleAgent` verifies status cluster member publications and if it is `NOT_CONNECTED` and if the cluster member IP address was resolved and was changed the agent will recreate necessary publications. I spotted that `io.aeron.cluster.MemberStatusPublisher#checkResult` doesn't check `NOT_CONNECTED` and it means that if a node was gone, everything, that was sent via such publication, will be skipped. And no need to take care of such publications in that status.
It seems like a working solution and I can specify hostnames and it will be resolved automatically when IP address.

What do you think guys?

### DIFF
--- a/aeron-cluster/src/main/java/io/aeron/cluster/ClusterMember.java
+++ b/aeron-cluster/src/main/java/io/aeron/cluster/ClusterMember.java
@@ -23,8 +23,12 @@ import org.agrona.CloseHelper;
 import org.agrona.collections.ArrayUtil;
 import org.agrona.collections.Int2ObjectHashMap;
 
+import java.net.InetAddress;
+import java.net.UnknownHostException;
+
 import static io.aeron.Aeron.NULL_VALUE;
 import static io.aeron.CommonContext.ENDPOINT_PARAM_NAME;
+import static io.aeron.CommonContext.UDP_MEDIA;
 import static io.aeron.archive.client.AeronArchive.NULL_POSITION;
 
 /**
@@ -54,6 +58,7 @@ public final class ClusterMember
     private final String archiveEndpoint;
     private final String endpointsDetail;
     private Publication publication;
+    private String resolvedAddress;
     private Boolean vote = null;
 
     /**
@@ -487,6 +492,7 @@ public final class ClusterMember
     {
         CloseHelper.close(publication);
         publication = null;
+        resolvedAddress = null;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


tmontgomery:
This approach has a number of issues, unfortunately.

One is that there is no backoff or rate limit to account for the fact that the connectivity may be down because the node is down and not simply moving. Clusters often operate in situations where nodes are down for extended periods of time. Especially in 5 node clusters. Slow tick is not so slow as it is several times a second.

Another is that resolution can block for a bit. Blocking in the consensus module loop is a bad thing. That combined with the above means at best spikes in latency. At worst, it may trigger an actual failure of a node.

segabriel:
I am aware of hostname resolving is not fast but it has networkaddress.cache. And yes slowTick is not a good place for it. Even adding a big timeout (5-10 sec) will not avoid spikes during resolving hostnames. And as I see creating a publication via `aeron.addExclusivePublication` will suffer from hostname resolving as well. Thus cluster layer is not an appropriate place for resolving IP address changes at all. So it's not the cluster issue. Am I right?

And probably the best place where to do it is `io.aeron.driver.DriverConductor` agent. Also no need to resolve hostnames for all publications, it should be set explicitly via channel params like:

```
aeron:udp?endpoint=my-hostname:5001|enableAutoResolving=true
```

And when something happens with publication (maybe check on losing SM, `statusMessageDeadlineNs`) start resolving hostnames for current publication.
Сould it be like that?

tmontgomery:
@segabriel correct. The DriverConductor is the best place to attempt resolves. And it only needs to be done for specific publications and subscriptions based on types and conditions. We don't need to tag channels. We automatically know the conditions that should trigger a resolve based on UDP operation. It is SM based for publications. It is traffic based for subscriptions with dynamic destinations. But has some caveats to do well.

We have plans and will be getting to this soon. It also needs to account for some of the ef_vi, DPDK, and io_uring operation we have long term plans for. There is a good bit to consider.

segabriel:
@tmontgomery Thank you for your reply.

PR #771, opened 2019-12-06
Update to Gradle 5.6.4

vyazelenko:


mjpt777:
We should look to upgrade to Gradle 6.x after the next release.

PR #775, opened 2019-12-10
RecordingLog should find service snapshots at index 0

dunkymole:


mjpt777:
This PR fails checkstyle.

mjpt777:
You also included binary files which should not be in the repo.

dunkymole:
Noted on both counts. I am going to close anyway since there should always be a preceding term entry.

PR #776, opened 2019-12-10
Fix compilation of C Media driver under visual studio

farnyser:
C version used in visual studio (C++ compiler) does not support VLA


mjpt777:
I've pushed a change which is more in keeping with the Aeron codebase.

PR #782, opened 2019-12-12
Cached clocks

mikeb01:
This is mostly refactoring/restructuring to support cached clocks.  Adds aeron_clock_t to hold state and changes the aeron_clock_func_t to take the clock pointer as an argument.  It pulls all of the clock functionality out of aeronmd.h and aeron_driver.c and puts it into a separate compilation unit (aeron_clock.{c,h}).

I've also made it an opaque pointer for safety.  Originally the use pattern was `thing->clock->now(thing->clock)`, the more I used that the more I realised that it wasn't very safe.  E.g. if I switched from a nano_clock to a cached_clock, but accidentally forgot to update the parameter then the results would not be valid.  Therefore I've hidden that behind `aeron_clock_now(aeron_clock_t *)`.  The opaque pointer makes it more difficult to use to use the clock incorrectly.

The cached clocks are allocated but not yet used, but that is a simpler exercise of using them in the appropriate places.


mjpt777:
The general approach looks good. The one part I'm uncomfortable about is passing the `NULL` argument to `aeron_clock_now`. This is not intuitive for what what unit is applicable for the return type.

mikeb01:
I'm going back to the drawing board on this one.

PR #785, opened 2019-12-15
Unit test for multiple services with a single consensus module.

mikeb01:


aeron-cluster/src/test/java/io/aeron/cluster/MultipleClusteredServicesTest.java
```diff
@@ -0,0 +1,121 @@
...
+        final AeronCluster client = AeronCluster.connect(new AeronCluster.Context()
+            .aeronDirectoryName(aeronDirName)
+            .ingressChannel("aeron:udp")
+            .clusterMemberEndpoints(TestCluster.clientMemberEndpoints(3)));
+
+        try
+        {
+            while (client.offer(new ExpandableArrayBuffer(100), 0, 100) < 0)
+            {
+                Thread.yield();
```
mjpt777: Should check for test interrupt in loops.

aeron-cluster/src/test/java/io/aeron/cluster/TestCluster.java
```diff
@@ -54,7 +59,7 @@
     private static final String ARCHIVE_CONTROL_RESPONSE_CHANNEL =
         "aeron:udp?term-length=64k|endpoint=localhost:8020";
     private static final String ARCHIVE_RECORDING_EVENTS_CHANNEL =
-        "aeron:udp?control-mode=dynamic|control=localhost:8030";
+        "aeron:udp?endpoint=235.0.0.1:8030";
```
mjpt777: Not good to use multicast in tests as issues can occur on Windows if publication is setup before subscriptions. I've disabled recording events for cluster tests.

PR #789, opened 2019-12-18
[C] Fixes for Windows

ltrzesniewski:
This fixes various things on Windows, and makes the CMake build and tests pass with VS 2019.

Every commit message explains the introduced changes in more detail.


aeron-driver/src/main/c/aeron_windows.c
```diff
@@ -31,62 +31,6 @@
 #define __builtin_popcount __popcnt
 #define __builtin_popcountll __popcnt64
 
-static DWORD dwTlsIndex; // address of shared memory
-
-// DllMain() is the entry-point function for this DLL.
-
-BOOL WINAPI DllMain(HINSTANCE hinstDLL, DWORD fdwReason, LPVOID lpvReserved)
```
mjpt777: Why remove DllMain?

aeron-driver/src/test/c/aeron_strutil_test.cpp
```diff
@@ -42,7 +46,7 @@ TEST_F(StrUtilTest, shouldHandleSingleValue)
     EXPECT_EQ(num_tokens, 1);
     EXPECT_STREQ(tokens[0], "single_token");
 
-    aeron_free(input);
+    free(input);
```
mjpt777: Why not keep the `aeron_free` version and have it do the right thing depending on platform?

aeron-driver/src/main/c/aeron_windows.c
ltrzesniewski: It does nothing useful, and wastes a TLS slot.

This code is basically taken from [the docs example](https://docs.microsoft.com/en-us/windows/win32/dlls/using-thread-local-storage-in-a-dynamic-link-library) but unlike the example it does not use `dwTlsIndex` anywhere outside of `DllMain`. So basically `DllMain` manages a TLS slot and some per-thread memory just for itself. The other TLS code does not need this.

aeron-driver/src/test/c/aeron_strutil_test.cpp
ltrzesniewski: `aeron_free` already does the right thing and calls `free`. I'm not sure why, but this additional indirection causes an assertion in the debug build. It looks like the two different modules use two different heaps.

In any case, I think `free` is more correct here, as the `strdup` docs tell you should call `free` on what it returns. Something like `aeron_free` could potentially switch to a different allocator in the future, breaking this code even for other platforms.

ltrzesniewski:
BTW would you accept a PR which adds a Windows CI, using something like GitHub Actions for instance (so it won't require you to setup any external account)?

If yes, would you like to also migrate the Linux CI from Travis to GitHub Actions for consistency?

mjpt777:
Migrating to GitHub Actions is on my list of things to try. Travis has been very frustrating.

aeron-driver/src/test/c/aeron_strutil_test.cpp
tmontgomery: `aeron_free` should be used by all memory calls. The debug build is probably mixing new and malloc.

aeron-driver/src/main/c/aeron_windows.c
tmontgomery: Actually, I think DllMain will be necessary. Currently, the error handling is broken in tests due to the TLS state not working. And that is partially due to it not being wired in correctly with DllMain or handling static libs correctly.

mjpt777:
I've reverted so it can done in parts which are easily agreed.

aeron-driver/src/main/c/aeron_windows.c
tmontgomery: A correct DllMain should use the existing slot and make the checks work. The changes to the TLS handling may work for static libs. Should write a test to make sure error handling is actually correct for threading, etc.

tmontgomery:
Everything but the DllMain and the removal of aeron_free from the test are OK.

aeron-driver/src/main/c/aeron_windows.c
ltrzesniewski: I actually fixed the error handling in this PR, it was previously causing one of the unit tests to fail because `aeron_thread_once` was not working properly.

While we *could* make the TLS slot be managed by `DllMain`, it would break TLS if the media driver is used as a static library, since those don't use `DllMain`, and it would create the need to maintain different code paths for Windows and Linux.

I think the TLS code from this PR is fine, and it doesn't need a `DllMain` function to work properly, unless I'm misunderstanding something.

aeron-driver/src/test/c/aeron_strutil_test.cpp
ltrzesniewski: The code uses `strdup` so we can't really control how it allocates the memory. The docs simply say to use `free` to release it.
Under a MSVC debug build, going through `aeron_free` causes an assertion failure, while using `free` works fine.

aeron-driver/src/main/c/aeron_windows.c
tmontgomery: You don't need different code paths. The DllMain can just pre-populate the TLS key/slot. Same code in what is there now but with DllMain but thread once will need to see if the slot is pre-populate or not. And as you point out it is needed for static libs. 

A DllMain is needed as we have plans that do require it soon. Also, while it is small, it is an advantage to pre-populate on DLL attach when possible.

aeron-driver/src/test/c/aeron_strutil_test.cpp
tmontgomery: As `aeron_free` uses free under the covers.... it indicates a different problem is at play. So, just moving to `free` does not fix that. Also, as I mentioned, all memory free should use `aeron_free`. Probably shouldn't be using `strdup` either. @mikeb01 can these not use `strdup` and be on the stack instead so that we can get rid of both `free` and `aeron_free` in this test?

tmontgomery: I've removed the strdup and free from the test as of 22401b35b

aeron-driver/src/main/c/aeron_windows.c
ltrzesniewski: > And as you point out it is needed for static libs.

Not exactly, I just said that `DllMain` is not called in static libs, which would require a different initialization path if you want to support them.

But since you have plans to require `DllMain` soon, I suppose that using the media driver as a static library will not be possible anymore in the future, right?

I could move the TLS slot handling to `DllMain` but my understanding is that it will make retrieving the error messages impossible in a static library on Windows as a consequence, unless we add the currently existing code in an `#if STATIC_LIBRARY` or something similar.

aeron-driver/src/test/c/aeron_strutil_test.cpp
ltrzesniewski: > can these not use `strdup` and be on the stack instead so that we can get rid of both `free` and `aeron_free` in this test?

I like that solution, it should be easy with a `strcpy` to the stack.

aeron-driver/src/main/c/aeron_windows.c
tmontgomery: This can be done, I think, quite easily. A static set to indicate that `TlsAlloc` has been called by DllMain means that `aeron_thread_key_create` doesn't need to call it. If not, then it can call `TlsAlloc`. All of this can be hidden in there. So, this shouldn't require anything weird. The use of `TlsSetValue` can still use the existing paths and created on demand.

If you are not comfortable about this or do not quite understand what I am suggesting. Just keep the DllMain and remove the TLS calls from it and keep the rest as you fixed it. I'll fix DllMain up when I have time.

In terms of requiring DllMain. Only the use of specific options may (or may not). Eventually, we would like to do DPDK, RIO, and other specific support. Some of that will be easier to integrate with DllMain than without.

tmontgomery: BTW, in a static lib, won't the error state be a leak when the thread goes away? Meaning to avoid the leak, DLL would be a better idea? (I've not looked for a hook to catch thread death for a while)...

aeron-driver/src/test/c/aeron_strutil_test.cpp
ltrzesniewski: @tmontgomery OK I'm pretty sure I understand the problem now.

CMake generated different configurations regarding the CRT library for the media driver and the unit test: the media driver uses *"Multi-threaded Debug DLL (/MDd)"* while the test uses *"Multi-threaded Debug (/MTd)"*.

This basically means they use a different instance of the C runtime, so memory allocated by one cannot be freed by the other.

Anyway, your fix makes the issue go away.

tmontgomery: AH. Makes perfect sense. Glad you figured it out and mentioned it. And glad the change works. Thanks!

aeron-driver/src/main/c/aeron_windows.c
ltrzesniewski: Ok I think I understand what you're suggesting now. I'll implement that tomorrow.

> Eventually, we would like to do DPDK, RIO, and other specific support.

That's great! 👍

> Some of that will be easier to integrate with DllMain than without.

I figured an empty `DllMain` wouldn't be useful and you could add it back at a later time when needed. Anyway, I'll add it back with specific support for error messages.

> BTW, in a static lib, won't the error state be a leak when the thread goes away?

Yes, it will leak. That can be mitigated from `DllMain`. But if I understand correctly, it also means there's currently a leak on Linux.

tmontgomery: On Linux it doesn't leak because `pthread_key_create` takes `free` as the destructor arg.

PR #794, opened 2019-12-19
[C] Fixes for Windows - Take 2

ltrzesniewski:
This is the same as #789 (which got reverted) but with two commits removed. It fixes the build and unit tests on Windows.

I'll implement changes discussed [here](https://github.com/real-logic/aeron/pull/789#discussion_r360052814) (`DllMain`) as a separate PR, as I feel the implementation may require more discussion, so it won't be cluttered by these changes.


tmontgomery:
Thanks! Will merge after checks done.

PR #795, opened 2019-12-20
[C] Handle error TLS from DllMain

ltrzesniewski:
This is the change discussed [here](https://github.com/real-logic/aeron/pull/789#discussion_r360052814). When the C media driver is used as a DLL, the TLS slot used to handle the error state is now managed by `DllMain`, and the per-thread memory is correctly freed when a thread exits.


tmontgomery:
Thanks!

PR #798, opened 2019-12-24
[C] Separate clocks in to separate compilcation unit (aeron_clock.{c,…

mikeb01:
…h}).  Remove all function pointer indirection to aeron_clock_func_t (and remove typedef).  Add aeron_stub_clock.{c,h} and use it within tests via a link time ordering to override the production clock implementation.

This is initially here for comment and will follow up this with changes to use the cached clocks in various places.

### DIFF
--- a/aeron-driver/src/main/c/util/aeron_clock.c
+++ b/aeron-driver/src/main/c/util/aeron_clock.c
@@ -67,10 +67,10 @@ int64_t aeron_clock_nano_time()
 
 int64_t aeron_clock_cached_epoch_time(aeron_clock_cache_t *cache)
 {
-    return cache->cached_ms;
+    return cache->cached_epoch_time;
 }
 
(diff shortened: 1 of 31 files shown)


aeron-driver/src/main/c/util/aeron_clock.c
```diff
@@ -0,0 +1,76 @@
...
+        return -1;
+    }
+#endif
+
+    return (ts.tv_sec * 1000000000) + ts.tv_nsec;
+}
+
+int64_t aeron_clock_cached_epoch_time(aeron_clock_cache_t *cache)
+{
+    return cache->cached_epoch_time;
```
mjpt777: This needs to be a volatile load as it can be concurrent. Same for update.

mikeb01:
Doesn't quite work with MacOS.  May come back to this implementation later.

PR #800, opened 2019-12-29
Cached clocks 3

mikeb01:
Simplified version:
- Moves clocks to a separate compilation unit.
- Retains existing nano and epoch clocks.
- Add cached clock contain cached epoch and nano times.
- Add an instance onto the media driver context.
- Utilise the cached clock in the appropriate places.


aeron-driver/src/main/c/aeronmd.h
```diff
@@ -738,25 +738,6 @@ int aeron_driver_close(aeron_driver_t *driver);
...
- * Clock function used by aeron.
- */
-typedef int64_t (*aeron_clock_func_t)();
-
-/**
- * Return time in nanoseconds for machine. Is not wall clock time.
- *
- * @return nanoseconds since epoch for machine.
- */
-int64_t aeron_nano_clock();
```
tmontgomery: These are part of the API and thus should not be removed from the header.

tmontgomery: aeronmd.h is the single header file for the API. So, the functions should be available with just this header file.

mikeb01: There's a couple of places where being able to include the aeron_clock.h instead of aeronmd.h is useful.  Would splitting aeron_clock.h into aeron_clock_api.h (for aeron_nano_clock and aeron_epoch_clock) and aeron_clock.h (all the rest) and have aeronmd.h and aeron_clock.h include aeron_clock_api.h work?

mikeb01: Does the cached clock need to become part of the API?

tmontgomery: I don't think the cached clock needs to be part of the API. If needed, then it could be included via pulling in the other headers. And also, the cached clock needs to be updated, etc. While the ms and ns clocks are simply single functions.

In terms of the header, I would like to keep the single include file if possible.

PR #803, opened 2019-12-31
[Java] Increase timeouts when running in debug.

mikeb01:
Basic implementation, intended for discussion.

Set aeron.debug.timeout.ns to some value.  For timeouts that need to be overridden, it calls `CommonContext::checkDebugTimeout` and replaces them with this value (if the debugger is attached).  It prints out a message to `stdout` when this occurs.  I didn't cover all of the timeouts, but enough that it allowed me to debug a couple of the cluster tests.

### DIFF
--- a/aeron-client/src/main/java/io/aeron/Aeron.java
+++ b/aeron-client/src/main/java/io/aeron/Aeron.java
@@ -1214,7 +1214,7 @@ public class Aeron implements AutoCloseable
          */
         public long interServiceTimeoutNs()
         {
-            return interServiceTimeoutNs;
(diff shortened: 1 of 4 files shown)


mjpt777:
The general approach looks sound. The challenge will be finding all the applicable timeouts and making sure other things progress as required, e.g. sending of status messages and heartbeats.

mikeb01:
I wonder if the simplest is to merge this as is so we can start using it and finding any places where additional timeouts may need to be dealt with.

tmontgomery:
I'm OK with merging and adjusting as needed.

PR #807, opened 2020-01-06
Logging and system tests

mikeb01:
- Increase the gradle log level when execution within the cppbuild to try and catch the errors that are occurring.
- Execute the Java system tests against the C media driver as part of the cppbuild.
   - This disables the C media drivers single system test (doesn't do much) as it causes breakages in the Java system tests.


mikeb01:
BTW, the aeron_system_test.cpp has been restored (contrary to the initial description).

PR #810, opened 2020-01-07
[Java] Remove 'compile' and 'testCompile' scope dependencies from Gradle build

vyazelenko:
In Gradle `compile` and `testCompile` dependency scopes were deprecated and will be removed in Gradle 7.0. They need to migrated to `implementation` or `api` for `compile` and `testImplementation` for `testCompile`.

**_Note: changing from `compile` to `implementation` makes a dependency non-transitive._**

See [API and implementation separation](https://docs.gradle.org/current/userguide/java_library_plugin.html#sec:java_library_separation).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -310,7 +310,7 @@ project(':aeron-driver') {
     mainClassName = 'io.aeron.driver.MediaDriver'
 
     dependencies {
-        compile project(':aeron-client')
+        api project(':aeron-client')
         testImplementation "org.mockito:mockito-inline:${mockitoVersion}"
     }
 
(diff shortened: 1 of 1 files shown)


mjpt777:
I tried to get this working on Aeron before but found limitations in Shadow not including the dependencies correctly. Worth confirming this works as expected.

vyazelenko:
@mjpt777 Can you elaborate?
I just tested master and my branch and see that the shadow jar for `aeron-all` are exactly the same.

mjpt777:
Maybe 5.2.0 fixed this. It was a problem the last time I checked earlier last year. Does all have the dependencies including Agrona and HdrHistogram?

vyazelenko:
Yes, both Agrona and HdrHistogram is there. Note, that I declared dependency twice:
- once for `aeron-samples` as `implementation`
- second time for `aeron-all` also as `implementation`

This was necessary since `implementation` dependencies are non-transitive.

PR #816, opened 2020-01-09
[Java] Use `Assertions.assertTimeoutPreemptively` to ensure that long running tests are aborted

vyazelenko:
JUnit 5 provides two sets of APIs for timeout handing:
- `Assertions.assertTimeout` - runs the code on the same thread and fails the test if the execution took longer than the specified timeout.
- Assertions.assertTimeoutPreemptively - run the code on a different thread and aborts running test if timeout reached. This is equivalent to the JUnit 4 `@Test(timeout=...)` and `org.junit.rules.Timeout` rule.

### DIFF
--- a/aeron-agent/src/test/java/io/aeron/agent/ArchiveLoggingAgentTest.java
+++ b/aeron-agent/src/test/java/io/aeron/agent/ArchiveLoggingAgentTest.java
@@ -22,7 +22,7 @@ import java.util.concurrent.CountDownLatch;
 import static io.aeron.agent.EventConfiguration.EVENT_READER_FRAME_LIMIT;
 import static io.aeron.agent.EventConfiguration.EVENT_RING_BUFFER;
 import static java.time.Duration.ofSeconds;
-import static org.junit.jupiter.api.Assertions.assertTimeout;
+import static org.junit.jupiter.api.Assertions.assertTimeoutPreemptively;
 
 public class ArchiveLoggingAgentTest
 {
@@ -56,7 +56,7 @@ public class ArchiveLoggingAgentTest
     @Test
     public void shouldLogMessages() throws Exception
     {
-        assertTimeout(ofSeconds(10), () ->
+        assertTimeoutPreemptively(ofSeconds(10), () ->
         {
             testDirName = Paths.get(IoUtil.tmpDirName(), "archive-test").toString();
             final File testDir = new File(testDirName);
--- a/aeron-agent/src/test/java/io/aeron/agent/ClusterLoggingAgentTest.java
+++ b/aeron-agent/src/test/java/io/aeron/agent/ClusterLoggingAgentTest.java
@@ -42,7 +42,7 @@ import java.util.concurrent.CountDownLatch;
 import static io.aeron.agent.EventConfiguration.EVENT_READER_FRAME_LIMIT;
 import static io.aeron.agent.EventConfiguration.EVENT_RING_BUFFER;
 import static org.agrona.BitUtil.SIZE_OF_INT;
-import static org.junit.jupiter.api.Assertions.assertTimeout;
+import static org.junit.jupiter.api.Assertions.assertTimeoutPreemptively;
 import static org.mockito.Mockito.mock;
 
 public class ClusterLoggingAgentTest
@@ -77,7 +77,7 @@ public class ClusterLoggingAgentTest
     @Test
     public void shouldLogMessages() throws Exception
     {
-        assertTimeout(Duration.ofSeconds(10), () ->
+        assertTimeoutPreemptively(Duration.ofSeconds(10), () ->
         {
             testDirName = Paths.get(IoUtil.tmpDirName(), "cluster-test").toString();
             final File testDir = new File(testDirName);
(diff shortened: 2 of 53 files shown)


QIvan:
i'm sorry, but I disagree with this PR a bit... 
After a [looooong discussion ](https://github.com/junit-team/junit5/issues/80) JUnit 5 team introduced a `@Timeout` annotation https://junit.org/junit5/docs/current/api/org/junit/jupiter/api/Timeout.html
From my point of view it looks better and more similar to the old tests. See examples in the doc:
https://junit.org/junit5/docs/current/user-guide/#writing-tests-declarative-timeouts

@vyazelenko Did you consider this API? If so, could you share your reasons why it's not sutable here? 

Thank you.

mjpt777:
@QIvan `@Timeout` is currently marked as an experimental feature but it does look useful.  What thread runs the after or clean up methods on timeout and successful operation?

vyazelenko:
@QIvan Put it simply: I missed `@Timeout` annotation. Thanks for the tip.

It seems that threading model is different: with `@Timeout` tests are always executed on the main thread which gets interrupted from another observer thread, whereas `assertTimeoutPreemptively` is executed on another thread (never on main).

As always there are pros and cons to any solution. The `assertTimeoutPreemptively` API works for all kinds of tests (including dynamic tests) and lets precisely control the scope on which timeout is applied. `@Timeout` on the other hand can be declared on either class or method level which means that one could avoid lots of repetition, also since the test is executed on the main thread timeout handling works in more cases (e.g. with `@Transactional` annotation in Spring tests).

I'll experiment with `@Timeout` to see if it works for our use cases.

vyazelenko:
@QIvan @mjpt777 Found an issue with `@Timeout`. It assumes that the main thread can be interrupted. However if thread is stuck on socket call or doing something that swallows interrupts such test will never terminate.

Here is an example:
```java
public class TimeoutTests
{
    @Test
    @Timeout(1)
    void timeout()
    {
        while (true)
        {
            try
            {
                System.out.println("Sleeping...");
                Thread.sleep(500);
            }
            catch (final InterruptedException e)
            {
                System.out.println("Ignore interrupt...");
            }
        }
    }

    @Test
    void assertTimeoutPreemptively()
    {
        Assertions.assertTimeoutPreemptively(Duration.ofSeconds(1), () ->
        {
            while (true)
            {
                try
                {
                    System.out.println("Sleeping...");
                    Thread.sleep(500);
                }
                catch (final InterruptedException e)
                {
                    System.out.println("Ignore interrupt...");
                }
            }
        });
    }
}
```

If you run `timeout` it will never terminate and will keep printing `Sleeping...`, e.g.:
```
Sleeping...
Sleeping...
Ignore interrupt...
Sleeping...
Sleeping...
Sleeping...
Sleeping...
Sleeping...
Sleeping...
```
Whereas if you run `assertTimeoutPreemptively` you see that it will terminate, e.g.:
```
Sleeping...
Sleeping...
Sleeping...
Ignore interrupt...
Sleeping...

execution timed out after 1000 ms
org.opentest4j.AssertionFailedError: execution timed out after 1000 ms
	at org.junit.jupiter.api.AssertTimeout.assertTimeoutPreemptively(AssertTimeout.java:144)
	at org.junit.jupiter.api.AssertTimeout.assertTimeoutPreemptively(AssertTimeout.java:111)
	at org.junit.jupiter.api.Assertions.assertTimeoutPreemptively(Assertions.java:3341)
	at org.agrona.TimeoutTests.assertTimeoutPreemptively(TimeoutTests.java:47)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.platform.commons.util.ReflectionUtils.invokeMethod(ReflectionUtils.java:686)
	at org.junit.jupiter.engine.execution.MethodInvocation.proceed(MethodInvocation.java:60)
	at org.junit.jupiter.engine.execution.InvocationInterceptorChain$ValidatingInvocation.proceed(InvocationInterceptorChain.java:131)
	at org.junit.jupiter.engine.extension.TimeoutExtension.intercept(TimeoutExtension.java:149)
	at org.junit.jupiter.engine.extension.TimeoutExtension.interceptTestableMethod(TimeoutExtension.java:140)
	at org.junit.jupiter.engine.extension.TimeoutExtension.interceptTestMethod(TimeoutExtension.java:84)
...
```

mikeb01:
I think that was the case in JUnit 4 as well.  In a number of the unit tests we'd call `TestUtil.checkInterruptedStatus()` within any busy wait loops.

calvertdw:
I see that this PR was merged, but it seems with JUnit 5.7.0 tests can still run past the timeout.

Simple case:
```
@Test
@Timeout(1)
public void testTimeout()
{
   Random random = new Random();
   double result = 0.0;
   for (int i = 0; i < 1e8; i++)
   {
      result += random.nextDouble();
   }
   System.out.println("Result: " + result);
}
```

Results in 
image

vyazelenko:
@calvertdw JUnit does not guarantee that the test will stop after timeout. Instead that the test will be marked as failed if the timeout expires.

PR #821, opened 2020-01-15
[Java] Add @SlowClusterTest annotation to allow excluding of clusteri…

mikeb01:
…ng tests without disabling them.

### DIFF
--- a/aeron-client/src/main/java/io/aeron/command/CounterMessageFlyweight.java
+++ b/aeron-client/src/main/java/io/aeron/command/CounterMessageFlyweight.java
@@ -79,13 +79,13 @@ public class CounterMessageFlyweight extends CorrelatedMessageFlyweight
     }
 
     /**
-     * Absolute offset of the key buffer
+     * Relative offset of the key buffer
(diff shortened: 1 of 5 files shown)


mjpt777:
Slow tests are more than cluster. Annotation would be better in `aeron-client`.

mikeb01:
I'd want to avoid putting the test annotation in aeron-client/main and sharing a dependency on another modules test is awkward (but possible).  Would it make more sense to have an aeron-test-util (or similar) module that could hold this and some of the other TestUtil classes that is only imported as a `testImplementation` dependency.

The much simpler implementation would be to not define our own annotation and just put `@Tag("slow")` directly on the test.

vyazelenko:
@mikeb01 I like the idea of having `aeron-test-common` module. We shouldn't duplicate test code in each module.

mjpt777:
We could have a test package in Agrona and this may be easier to share.

tmontgomery:
Test package in Agrona sounds good to me. Have needed that for a while.

PR #825, opened 2020-01-17
Add aeron-test-support module and @SlowTest annotation.

mikeb01:
Went with an aeron-test-support module for now, to save lots of back and forth with gradle and maven central.  We can pull up to agrona anything that may be valuable for SBE as that happens.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -859,13 +859,13 @@ project(':aeron-system-tests') {
 
 project(':aeron-test-support') {
     dependencies {
-        implementation "org.junit.jupiter:junit-jupiter-api:${junitVersion}"
+        api "org.junit.jupiter:junit-jupiter-api:${junitVersion}"
     }
 }
 
 project(':aeron-all') {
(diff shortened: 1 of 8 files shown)


build.gradle
```diff
@@ -819,6 +857,12 @@ project(':aeron-system-tests') {
     }
 }
 
+project(':aeron-test-support') {
+    dependencies {
+        implementation "org.junit.jupiter:junit-jupiter-api:${junitVersion}"
```
vyazelenko: The dependency on JUnit should be `api` instead of `implementation` since this module depends in it's public API on the JUnit classes.

vyazelenko:
@mikeb01 Should we have a separate CI build that triggers slow tests? Maybe we can trigger them nightly and on demand. /cc @mjpt777

mikeb01:
Yes, ran out of time to do that today. Can do that next week once it is merged.

mjpt777:
I'm good with this. @tmontgomery ?

tmontgomery:
I'm good with this.

PR #829, opened 2020-01-19
[Java] Add slow test Github actions target (linux, jdk11 only).

mikeb01:
Runs in about 7 minutes in CI.  Runs any test tagged with @SlowTest.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,6 +42,36 @@ jobs:
       - name: Build with Gradle
         run: ./gradlew
 
+  java-slow-test-build:
+    name: Java ${{ matrix.java }} (${{ matrix.os }})
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        java: [ '11' ]
+        os: ['ubuntu-latest']
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Cache Gradle dependencies
+        uses: actions/cache@v1
+        with:
+          path: ~/.gradle/caches
+          key: ${{ runner.os }}-gradle-caches-${{ hashFiles('**/*.gradle') }}
+          restore-keys: |
+            ${{ runner.os }}-gradle-caches-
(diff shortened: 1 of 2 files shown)


vyazelenko:
I thought the idea was to have `slow tests` run on a different cadence then the rest of the build, e.g. nightly build and maybe via a manual trigger. As it currently stands the `slow tests` build will run on every commit and PR.

If we are to separate them from the main build we'll need to create a separate yaml file just for that build and specify trigger conditions there, i.e. schedule and manual trigger. See [Events that trigger workflows](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/events-that-trigger-workflows).

On the other hand having `slow tests` for every build will catch errors faster but then why separate them into a separate build and not just run all tests in the normal build?

.github/workflows/ci.yml
```diff
@@ -42,6 +42,36 @@ jobs:
       - name: Build with Gradle
         run: ./gradlew
 
+  java-slow-test-build:
+    name: Java ${{ matrix.java }} (${{ matrix.os }})
```
vyazelenko: Please use different name so we can distinguish from the rest of the builds.

.github/workflows/ci.yml
```diff
@@ -42,6 +42,36 @@ jobs:
...
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        java: [ '11' ]
```
vyazelenko: I would go with Java 8 since it is faster for Aeron builds than 11.

mjpt777:
I agree that the cadence should be about once per day.

mikeb01:
I've moved it to a 12 hour cycle and put the full matrix of build targets in place (JDK, OS).

mikeb01:
I've verified it works on a 20 minute cycle.  Moved it back to 12 hours and will wait until tomorrow to see.  I've dropped Windows out of the matrix for the time being as it fails.  Will have to dig into the reasons for that later (probably file handles).

PR #833, opened 2020-01-23
Stream id in uri

mikeb01:
This adds support in the Java media driver and Java client for specifying the stream id in the URI.  There is a common message used for adding subscriptions, publications and exclusive publications (ChannelMessage), which just contains the URI.  There are new control protocol event codes for each of this new events.  I've gone with a naming scheme which appends the version to the name of the event (e.g. `ADD_PUBLICATION_V1`), with the assumption that the existing one is `_V0`, but didn't change the existing names in case users have referenced them.  This did require an additional conductor event SUBSCRIPTION_READY_V1 as the existing event didn't return the streamId (happy to use an alternate naming scheme here if anyone has a better approach). 

When using the new version of the subscription we delay the construction of the `Subscription` object until we get the response.  We stash the channel and availability handlers in the same manner that the publication does.

I've upped the CnC version from 0.0.16 to 0.1.0.

I've included support for the new messages in the event log dissector.

C media driver and C++ client haven't been updated yet, I was going to see if we're happy with this approach before tackling those.

I've also updated the ChannelUriStringBuilder to take in a uri string or constructed ChannelUri which is useful for amending existing URIs.

### DIFF
--- a/aeron-agent/src/main/java/io/aeron/agent/CmdInterceptor.java
+++ b/aeron-agent/src/main/java/io/aeron/agent/CmdInterceptor.java
@@ -53,8 +53,13 @@ class CmdInterceptor
         CMD_IN_ADD_RCV_DESTINATION,
         CMD_IN_REMOVE_RCV_DESTINATION,
         CMD_OUT_ON_CLIENT_TIMEOUT,
-        CMD_IN_TERMINATE_DRIVER);
+        CMD_IN_TERMINATE_DRIVER,
+        CMD_IN_ADD_PUBLICATION_V1,
+        CMD_IN_ADD_EXCLUSIVE_PUBLICATION_V1,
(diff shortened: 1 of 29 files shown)


aeron-agent/src/main/java/io/aeron/agent/CmdInterceptor.java
```diff
@@ -53,8 +53,13 @@
         CMD_IN_ADD_RCV_DESTINATION,
         CMD_IN_REMOVE_RCV_DESTINATION,
         CMD_OUT_ON_CLIENT_TIMEOUT,
-        CMD_IN_TERMINATE_DRIVER);
+        CMD_IN_TERMINATE_DRIVER,
+        CMD_IN_ADD_PUBLICATION_V1,
```
mjpt777: Not sure on the `_V1` naming. Need to mull it.

PR #836, opened 2020-01-29
[C] Add client shared library

kdoherty2:


aeron-client/src/main/cpp/util/aeron_error_client.c
```diff
@@ -0,0 +1,84 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "util/aeron_error_client.h"
+
+#if defined(AERON_COMPILER_MSVC)
+static pthread_key_t error_key = TLS_OUT_OF_INDEXES;
+
+BOOL aeron_error_dll_process_attach()
```
tmontgomery: What prevents these functions from name clashing with the existing functions in the C native driver? If the C native driver were brought in, this seems like it would break.

tmontgomery:
What does the TLS error handling do? The API at this point does not use this style of error handling. So, why is it added? 

Also, the naming conflicts with C native driver naming making bringing in an embedded native driver problematic.

Thirdly, what is the motivation for this PR? The client is quite small. What does making it a shared object help with?

Fourth, why export all symbols? This is less than ideal on Windows. Why do this?

aeron-client/src/main/cpp/util/aeron_error_client.c
kdoherty2: This will be addressed

skurkchan:
re: TLS error handling
I can remove DLLMain then. Is that OK ?
Or keep the function and remove error handling ?
Thanks.

tmontgomery:
The DllMain and associated error functions are not needed in the client. And as I point out, could conflict if the C driver is used with the client.

mjpt777:
How where the set of classes decided that `CLIENT_EXPORT` got added to?

PR #837, opened 2020-01-30
[Java] Re-implement replay using new `offerBlock` API

vyazelenko:
In order to apply some batching for the replay a new method `ExclusivePublication#offerBlock` was added. It is a low-level API intended for offering pre-formatted block of messages directly into publication. This method then copies provided block verbatim into the underlying term.
Preliminary performance tests showed a 50% increase in message throughput for the replay.

### DIFF
--- a/aeron-client/src/main/java/io/aeron/ExclusivePublication.java
+++ b/aeron-client/src/main/java/io/aeron/ExclusivePublication.java
@@ -19,10 +19,13 @@ import io.aeron.logbuffer.BufferClaim;
 import io.aeron.logbuffer.ExclusiveTermAppender;
 import io.aeron.logbuffer.LogBufferDescriptor;
 import org.agrona.DirectBuffer;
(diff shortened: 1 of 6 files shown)


aeron-client/src/main/java/io/aeron/ExclusivePublication.java
```diff
@@ -400,6 +403,77 @@ public long appendPadding(final int length)
...
+                ", available space in the term buffer=" + availableSpace);
+        }
+    }
+
+    private void checkBlockStart(final MutableDirectBuffer buffer, final int offset)
+    {
+        final int sessionId = this.sessionId;
+        final int streamId = this.streamId;
+        final int frameType = HDR_TYPE_DATA;
+        final int blockSessionId = buffer.getInt(offset + SESSION_ID_FIELD_OFFSET, LITTLE_ENDIAN);
```
mjpt777: Should also check `term-offset`.

PR #840, opened 2020-02-07
[Java] Use `AeronCloseHelper` with the `ErrorHandler` to close resources

### LINKED ISSUE #818: Ensure clean shutdown upon abrupt termination
Make sure that Aeron, Archive and Cluster can be cleanly shutdown and close all resources upon abrupt termination.

vyazelenko:
Ensure proper closure of the resources by using new helper API that uses `ErrorHandler` which consumes
any error generated by the resource being closed. In cases where there is no `ErrorHandler` around or if
it was already closed use variants of the `CloseHelper#closeAll` methods and if not possible
`CloseHelper#quiteClose` as the final resort.

Because not all of the resource being closed implement the `AutoCloseable` interface we are forced to use
lambda expressions in several places. This might be a potential performance issues, however given the fact
that we are dealing with `close` methods that should never be on the fast path.

Closes #818

### DIFF
--- a/aeron-client/src/main/java/io/aeron/Aeron.java
+++ b/aeron-client/src/main/java/io/aeron/Aeron.java
@@ -153,7 +153,7 @@ public class Aeron implements AutoCloseable
     /**
      * Print out the values from {@link #countersReader()} which can be useful for debugging.
      *
-     *  @param out to where the counters get printed.
+     * @param out to where the counters get printed.
      */
     public void printCounters(final PrintStream out)
     {
@@ -225,13 +225,14 @@ public class Aeron implements AutoCloseable
     {
         if (IS_CLOSED_UPDATER.compareAndSet(this, 0, 1))
         {
+            final ErrorHandler errorHandler = ctx.errorHandler();
             if (null != conductorRunner)
             {
-                conductorRunner.close();
+                AeronCloseHelper.close(errorHandler, conductorRunner);
             }
             else
             {
-                conductorInvoker.close();
+                AeronCloseHelper.close(errorHandler, conductorInvoker);
             }
         }
     }
@@ -1128,7 +1129,7 @@ public class Aeron implements AutoCloseable
 
         /**
          * Set a {@link Runnable} that is called when the client is closed by timeout or normal means.
-         *
+         * <p>
          * It is not safe to call any API functions from any threads after this hook is called. In addition any
          * in flight calls may still cause faults. It is thus recommended to treat this as a hard error and
          * terminate the process in this hook as soon as possible.
--- a/aeron-cluster/src/test/java/io/aeron/cluster/ClusterBackupMediaDriverTest.java
+++ b/aeron-cluster/src/test/java/io/aeron/cluster/ClusterBackupMediaDriverTest.java
@@ -1,0 +1,63 @@
+/*
+ * Copyright 2014-2020 Real Logic Limited.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.aeron.cluster;
+
+import io.aeron.archive.Archive;
+import io.aeron.driver.MediaDriver;
+import org.agrona.ErrorHandler;
+import org.agrona.concurrent.CountedErrorHandler;
+import org.junit.jupiter.api.Test;
+import org.mockito.InOrder;
+
+import static io.aeron.test.Tests.throwOnClose;
+import static org.junit.jupiter.api.Assertions.assertSame;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.mockito.Mockito.*;
+
+class ClusterBackupMediaDriverTest
+{
+    @Test
+    void close() throws Exception
+    {
+        final ErrorHandler errorHandler = mock(ErrorHandler.class);
+        final CountedErrorHandler countedErrorHandler = mock(CountedErrorHandler.class);
+
+        final MediaDriver driver = mock(MediaDriver.class);
+        when(driver.context()).thenReturn(new MediaDriver.Context().errorHandler(errorHandler));
+        final AssertionError driverException = new AssertionError("driver");
+        throwOnClose(driver, driverException);
+
+        final Archive archive = mock(Archive.class);
+        when(archive.context()).thenReturn(new Archive.Context().countedErrorHandler(countedErrorHandler));
+        final ArithmeticException archiveException = new ArithmeticException("archive");
+        throwOnClose(archive, archiveException);
+
+        final ClusterBackup clusterBackup = mock(ClusterBackup.class);
+        final IllegalStateException clusterBackupException = new IllegalStateException();
+        throwOnClose(clusterBackup, clusterBackupException);
+
+        final ClusterBackupMediaDriver clusterBackupMediaDriver =
+            new ClusterBackupMediaDriver(driver, archive, clusterBackup);
+
+        final AssertionError ex = assertThrows(AssertionError.class, clusterBackupMediaDriver::close);
+
+        assertSame(driverException, ex);
+        final InOrder inOrder = inOrder(errorHandler, countedErrorHandler);
+        inOrder.verify(countedErrorHandler).onError(clusterBackupException);
+        inOrder.verify(errorHandler).onError(archiveException);
+        inOrder.verifyNoMoreInteractions();
+    }
+}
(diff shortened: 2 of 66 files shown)


aeron-archive/src/main/java/io/aeron/archive/client/ReplayMerge.java
```diff
@@ -175,18 +176,23 @@ public void close()
...
-            if (!archive.context().aeron().isClosed())
+            final AeronArchive.Context context = archive.context();
+            final Aeron aeron = context.aeron();
+            if (!aeron.isClosed())
             {
+                final ErrorHandler errorHandler = context.errorHandler();
                 if (State.MERGED != state && State.STOP_REPLAY != state)
                 {
-                    subscription.removeDestination(replayDestination);
+                    AeronCloseHelper.close(errorHandler, () -> subscription.removeDestination(replayDestination));
```
mjpt777: This should be an asyncRemoveDestination and not need a close helper.

aeron-archive/src/main/java/io/aeron/archive/client/ReplayMerge.java
```diff
@@ -175,18 +176,23 @@ public void close()
...
 
                 if (isReplayActive)
                 {
                     isReplayActive = false;
-                    final long correlationId = archive.context().aeron().nextCorrelationId();
-                    archive.archiveProxy().stopReplay(replaySessionId, correlationId, archive.controlSessionId());
+                    final long correlationId = aeron.nextCorrelationId();
+                    AeronCloseHelper.close(errorHandler,
+                        () -> archive.archiveProxy()
+                        .stopReplay(replaySessionId, correlationId, archive.controlSessionId()));
```
mjpt777: This should be dealt with at a higher level in the doWork duty cycle.

aeron-client/src/main/java/io/aeron/Aeron.java
```diff
@@ -1313,8 +1314,8 @@ public void close()
         {
             final MappedByteBuffer cncByteBuffer = this.cncByteBuffer;
             this.cncByteBuffer = null;
-            IoUtil.unmap(cncByteBuffer);
-            super.close();
+            AeronCloseHelper.close(errorHandler, () -> IoUtil.unmap(cncByteBuffer));
```
mjpt777: Unmap should not generate an error.

aeron-cluster/src/test/java/io/aeron/cluster/ClusterBackupMediaDriverTest.java
```diff
@@ -0,0 +1,66 @@
...
+        final AssertionError driverException = new AssertionError("driver");
+        throwOnClose(driver, driverException);
+
+        final Archive archive = mock(Archive.class);
+        when(archive.context()).thenReturn(new Archive.Context().countedErrorHandler(countedErrorHandler));
+        final ArithmeticException archiveException = new ArithmeticException("archive");
+        throwOnClose(archive, archiveException);
+
+        final ClusterBackup clusterBackup = mock(ClusterBackup.class);
+        final BatchUpdateException clusterBackupException = new BatchUpdateException();
```
mjpt777: Is the BatchUpdateException type correct?

aeron-cluster/src/test/java/io/aeron/cluster/ConsensusModuleContextCloseTests.java
```diff
@@ -0,0 +1,148 @@
...
+    @Test
+    void doesNotOwnAeronClientAndClientIsNotClosed() throws Exception
+    {
+        context.ownsAeronClient(false);
+        when(aeron.isClosed()).thenReturn(false);
+
+        final IllegalStateException ex = assertThrows(IllegalStateException.class, context::close);
+
+        assertSame(moduleStateException, ex);
+        assertArrayEquals(new Throwable[]{
```
mjpt777: Match braces.

aeron-samples/src/main/java/io/aeron/samples/archive/EmbeddedRecordingThroughput.java
```diff
@@ -99,12 +101,12 @@ public EmbeddedRecordingThroughput()
...
-        CloseHelper.close(aeron);
-        CloseHelper.close(archivingMediaDriver);
-
-        archivingMediaDriver.archive().context().deleteArchiveDirectory();
-        archivingMediaDriver.mediaDriver().context().deleteAeronDirectory();
+        CloseHelper.closeAll(
+            aeronArchive,
+            aeron,
+            archivingMediaDriver,
+            () -> archivingMediaDriver.archive().context().deleteArchiveDirectory(),
```
mjpt777: We should add a delete specific function that takes an error handler. This would avoid quite a few lambdas.

aeron-archive/src/main/java/io/aeron/archive/client/ReplayMerge.java
vyazelenko: @mjpt777 There is already a `stopReplay()` which does that and it is called from the `doWork`. Does it mean that the entire ` if (isReplayActive)` block can be removed from the `close()` method?

aeron-samples/src/main/java/io/aeron/samples/archive/EmbeddedRecordingThroughput.java
vyazelenko: @mjpt777 Having a delete with an `ErrorHandler` won't help in this case since we are deleting the aeron directories which contain the error log to which `ErrorHandler` will normally write. So the only thing we can do if don't want to use lambdas when deleting directories is to leverage the fact that the `org.agrona.IoUtil#delete` method accepts a boolean which allows ignoring errors.
So the `close()` might become:
```java
        CloseHelper.closeAll(
            aeronArchive,
            aeron,
            archivingMediaDriver);
       IOUtil.delete(new File(archivingMediaDriver.archive().context().deleteArchiveDirectory()), true);
       IOUtil.delete(new File(archivingMediaDriver.mediaDriver().context().deleteAeronDirectory()), true);
``` 
However the deletes will not be performed if the `closeAll` throws. Which means we will also need to use `quiteCloseAll` instead.

Provided all of the above and the fact that this is all in samples I'm keen on keeping the current implementation.

aeron-archive/src/main/java/io/aeron/archive/client/ReplayMerge.java
vyazelenko: Done

aeron-client/src/main/java/io/aeron/Aeron.java
vyazelenko: I change all of the unmap calls to be naked.

aeron-cluster/src/test/java/io/aeron/cluster/ClusterBackupMediaDriverTest.java
vyazelenko: Changed

aeron-cluster/src/test/java/io/aeron/cluster/ConsensusModuleContextCloseTests.java
vyazelenko: Done

aeron-driver/src/main/java/io/aeron/driver/MediaDriver.java
```diff
@@ -522,23 +525,23 @@ public void close()
...
+                    CloseHelper.quietClose((AutoCloseable)errorHandler); // Ignore error to ensure the rest is closed
                 }
 
                 IoUtil.unmap(cncByteBuffer);
                 cncByteBuffer = null;
 
                 if (dirDeleteOnShutdown && null != aeronDirectory())
                 {
-                    deleteAeronDirectory();
+                    CloseHelper.quietClose(this::deleteAeronDirectory);
```
mjpt777: Delete can be better handled with this. https://github.com/real-logic/agrona/commit/43cdebd5dfc0d1daa2684da632de951f4fecae55

vyazelenko: @mjpt777 Have you seen my [previous comment](https://github.com/real-logic/aeron/pull/840#discussion_r376727260)? We won't be able to use new API here, because we are deleting Aeron directory which contains the error log thus we cannot use `ErrorHandler` here.
Plus it would require bypassing `deleteAeronDirectory` and calling the `IoUtils#delete` method directly.

In this particular case we can probably just call `deleteAeronDirectory` without `quietClose` since the only other action in the close sequence is to call `super.close` which is an empty method as of now.

mjpt777: You are right that we have an ordering issue here. We cannot `quietClose` this otherwise we miss cases on Windows that should be notified.

vyazelenko: @mjpt777 I changed code to call `deleteAeronDirectory()` directly.

aeron-archive/src/main/java/io/aeron/archive/client/ReplayMerge.java
mjpt777: I've pushed a different way to stop the replay which may fail to merge so this should be better.

aeron-samples/src/main/java/io/aeron/samples/archive/EmbeddedRecordingThroughput.java
mjpt777: We need to make sure the delete is attempted and any errors reported. Lambdas may be required here.

vyazelenko: @mjpt777 OK, so I'll keep current logic.

aeron-archive/src/main/java/io/aeron/archive/client/ReplayMerge.java
vyazelenko: Merged. The close remains as-is from the HEAD.

PR #853, opened 2020-02-13
[Java] Add an option to free allocated direct ByteBuffers

vyazelenko:
Free allocated direct `ByteBuffers` when a `Context#close` is called and the context was configured with the `shouldFreeBuffersOnClose(true)`.

There is at least one open question, i.e. are we allowed to free `ByteBuffers` for the `DataTransportPoller` and `ControlTransportPoller`?

### DIFF
--- a/aeron-archive/src/main/java/io/aeron/archive/Archive.java
+++ b/aeron-archive/src/main/java/io/aeron/archive/Archive.java
@@ -44,7 +44,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
 import java.util.function.Supplier;
 
-import static io.aeron.archive.Archive.Configuration.MAX_BLOCK_LENGTH;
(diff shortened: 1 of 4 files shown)


aeron-archive/src/main/java/io/aeron/archive/Archive.java
```diff
@@ -1963,34 +1994,49 @@ UnsafeBuffer dataBuffer()
...
+                dataBuffer = allocateBuffer();
             }
-
             return dataBuffer;
         }
 
         UnsafeBuffer replayBuffer()
         {
-            if (null == replayBuffer)
+            if (DEDICATED != threadingMode)
```
mjpt777: Why the extra checks on every access. Existing code pays price once.

vyazelenko: I want to void aliasing the same buffer from different variables so that close/free will be done only once. Of course one could do it by checking the threading mode in the `close`.

PR #854, opened 2020-02-13
[C] Fix aeron_randomised_int32 on Windows

ltrzesniewski:
The media driver always produced the same sequence of session IDs.

 - `rand()` was apparently not seeded properly (yet it should have been, I need to check why)
 - `rand() * INT_MAX` would always cause an overflow
 - This uses `rand_s` instead of `rand` for better randomness and to avoid the dependency on a seed (it better aligns to what is done on other platforms)


tmontgomery:
Will merge once checks pass.

PR #855, opened 2020-02-13
[Java] Set field pointing to MappedByteBuffer to null before calling IoUtil#unmap

vyazelenko:
For the symmetry with other places where we free buffers we should null-out the corresponding field reference before freeing/un-mapping the underlying buffer.


mjpt777:
Some bits of this are useful but others should not be done. Classes like Catalog which are not creating and destroyed more than once in normal operation don't need this and it loses the opportunity to optimise the final fields.

vyazelenko:
I've reverted change to the `Catalog`. Anything else you'd like to exclude?

PR #859, opened 2020-02-17
Group size on min fc strategy

mikeb01:
- Implement group size tracking for min flow control strategy (C & Java Drivers)
- Separate flow control system tests by strategy.
- Remove some noise from debug timeout overrides.


aeron-driver/src/main/java/io/aeron/driver/MinMulticastFlowControl.java
```diff
@@ -42,6 +43,7 @@
     static final Receiver[] EMPTY_RECEIVERS = new Receiver[0];
     private Receiver[] receivers = EMPTY_RECEIVERS;
```
mjpt777: receivers needs to be volatile.

aeron-driver/src/main/java/io/aeron/driver/MinMulticastFlowControl.java
```diff
@@ -140,7 +154,12 @@ public long onIdle(final long timeNs, final long senderLimit, final long senderP
             receivers = truncateReceivers(receivers, removed);
         }
 
-        return receivers.length > 0 ? minLimitPosition : senderLimit;
+        return receivers.length < groupMinSize || receivers.length == 0 ? senderLimit : minLimitPosition;
+    }
+
+    public boolean hasRequiredReceivers()
```
mjpt777: receivers has to be volatile because hasRequiredReceivers is called from the conductor thread whereas the other methods are called from the sender thread.

aeron-system-tests/src/test/java/io/aeron/TaggedFlowControlSystemTest.java
```diff
@@ -344,23 +344,13 @@ void shouldPreventConnectionUntilRequiredGroupSizeMatchTagIsMet()
...
                 subscription2 = clientA.addSubscription(uriPlain, STREAM_ID);
-
-                Thread.sleep(100);
-                assertFalse(publication.isConnected());
-
                 subscription3 = clientB.addSubscription(uriWithReceiverTag, STREAM_ID);
 
-                Thread.sleep(100);
+                // Sleep for a bit to ensure that we haven't become connected.
+                Thread.sleep(500);
```
mjpt777: Need to find a better way than this.

mikeb01: It's tricky testing to see if something hasn't happened.  Perhaps I can look at some of the counters to infer that the NetworkPublication has see status messages from 2 endpoints.

aeron-driver/src/main/java/io/aeron/driver/MinMulticastFlowControl.java
mikeb01: I think groupMinSize also needs to be volatile.

aeron-system-tests/src/test/java/io/aeron/TaggedFlowControlSystemTest.java
mikeb01: I've got rid of all of the long timeouts in the flow control tests.  Mostly I'm waiting for the subscription to connect and seeing the status messages increment directly after that.

PR #860, opened 2020-02-18
Refactor tagged and min fc strategy

mikeb01:
- For both Java and C driver, treat the min flow control strategy as a special case of tagged where is_tagged is always true.


aeron-client/src/main/java/io/aeron/ChannelUriStringBuilder.java
```diff
@@ -1228,6 +1228,36 @@ public ChannelUriStringBuilder taggedFlowControl(
...
+
+    /**
+     * Set the min flow control settings to be used on channel.  All specified values may be null and the default
+     * specified in the MediaDriver.Context will be used instead.
+     *
+     * @param groupSize group size required to allow publications for this channel to be move to connected status.
+     * @param timeout timeout receivers, default is ns, but allows suffixing of time units (e.g. 100ms).
+     * @return this for fluent API.
+     */
+    public ChannelUriStringBuilder minFlowControl(
```
mjpt777: Should this also be done for tagged?

aeron-driver/src/main/java/io/aeron/driver/MinMulticastFlowControl.java
```diff
@@ -32,40 +27,13 @@
  * Tracking of receivers is done as long as they continue to send Status Messages. Once SMs stop, the receiver tracking
  * for that receiver will timeout after a given number of nanoseconds.
  */
-public class MinMulticastFlowControl implements FlowControl
+public class MinMulticastFlowControl extends TaggedMulticastFlowControlBase
```
mjpt777: Tagged should extend Min as it added new features.

aeron-client/src/main/java/io/aeron/ChannelUriStringBuilder.java
mikeb01: Yes :-), https://github.com/real-logic/aeron/blob/master/aeron-client/src/main/java/io/aeron/ChannelUriStringBuilder.java#L1201

aeron-driver/src/main/java/io/aeron/driver/MinMulticastFlowControl.java
mikeb01: I think the problem is with the naming of the base class.  I'm going to rework is slightly.  I see Tagged and Min as peers.  Just with different approaches to as to whether the receiver is in the group.  Min - always, Tagged - if receiverTag matches.

mikeb01:
This moves some things around.  The base class is now AbstractGroupMinFlowControl and is agnostic of tagging, but acknowledges that receivers can be in or out of a nominated group.  In the min case all receivers are in the group and for tagged only the ones that where the receiverTag matches are in the group.

mjpt777:
This branch has conflicts which need addressed before it can be merged.

aeron-driver/src/main/java/io/aeron/driver/TaggedMulticastFlowControlBase.java
```diff
@@ -0,0 +1,220 @@
...
+
+import io.aeron.CommonContext;
+import io.aeron.driver.media.UdpChannel;
+import io.aeron.protocol.StatusMessageFlyweight;
+import org.agrona.AsciiEncoding;
+import org.agrona.SystemUtil;
+
+import static io.aeron.logbuffer.LogBufferDescriptor.computePosition;
+
+public abstract class TaggedMulticastFlowControlBase implements FlowControl
```
mjpt777: I think `AbstractMinMulticastFlowControl` is a better name.

mikeb01:
This has diverged too far.  I'm going recreate the PR from a fresh branch.

PR #861, opened 2020-02-18
[C] Windows fixes

ltrzesniewski:
Here's another round of Windows fixes for the media driver. See individual commit messages for details. The cleanup commit mostly gets rid of some warnings and makes the IDE happier.

There are still other things I'll need to fix though, I'll send the fixes later.

### DIFF
--- a/aeron-driver/src/main/c/concurrent/aeron_atomic64_gcc_x86_64.h
+++ b/aeron-driver/src/main/c/concurrent/aeron_atomic64_gcc_x86_64.h
@@ -64,16 +64,6 @@ do \
 } \
 while (false)
 
-#define AERON_CMPXCHG64(original, dst, expected, desired) \
-do \
-{ \
-    asm volatile( \
-        "lock; cmpxchgq %2, %1" \
-        : "=a"(original), "+m"(dst) \
-        : "q"(desired), "0"(expected)); \
-} \
-while (false)
-
 inline bool aeron_cmpxchg64(volatile int64_t* destination, int64_t expected, int64_t desired)
 {
     int64_t original;
--- a/aeron-driver/src/main/c/media/aeron_udp_channel_transport.c
+++ b/aeron-driver/src/main/c/media/aeron_udp_channel_transport.c
@@ -20,9 +20,9 @@
 #endif
 
 #include "util/aeron_platform.h"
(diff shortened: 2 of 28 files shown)


aeron-driver/src/main/c/util/aeron_error.c
```diff
@@ -168,8 +168,14 @@ void aeron_set_windows_error()
         0,
         NULL);
 
-    aeron_set_err(errorId, messageBuffer);
-    free(messageBuffer);
+	if (result == 0)
```
mjpt777: Formatting is a bit screwed here. Is there something missing?

aeron-driver/src/main/c/concurrent/aeron_atomic64_gcc_x86_64.h
```diff
@@ -64,16 +64,6 @@ do \
 } \
 while (false)
 
-#define AERON_CMPXCHG64(original, dst, expected, desired) \
```
mjpt777: Why remove this?

aeron-driver/src/main/c/util/aeron_error.c
ltrzesniewski: Sorry, I inserted tabs here instead of spaces, will fix.

aeron-driver/src/main/c/concurrent/aeron_atomic64_gcc_x86_64.h
ltrzesniewski: - It's not used anywhere, `aeron_cmpxchg64` is used instead
 - The MSVC definition used GCC syntax so wouldn't compile if used

aeron-driver/src/main/c/media/aeron_udp_channel_transport.c
```diff
@@ -226,7 +226,7 @@ int aeron_udp_channel_transport_init(
     if(set_socket_non_blocking(transport->fd) < 0)
```
mjpt777: Minor point on formatting here. No space after if.

ltrzesniewski: Ok, I'll fix that in the next PR

PR #863, opened 2020-02-20
[C] Windows fixes, next part

ltrzesniewski:
These are larger fixes than in my last PR, I recommend reviewing this commit-by-commit.

The remaining warnings that MSVC outputs are:
 - Various assignments of `size_t` values to `int` (I'm surprised GCC doesn't warn about those)
 - `getsockopt` and `setsockopt` have a `char*` type for the value instead of `void*`. It would be nice to have an `aeron_*` version of those to deal with those warnings if you're OK with that.

Also, errors aren't currently logged properly on Windows ([`strerror` is kind of useless](https://docs.microsoft.com/en-us/cpp/c-language/strerror-function?view=vs-2019) and `errno` isn't used by winsock). Fixing that would require replacing calls to `aeron_set_err` that use `strerror` with something similar to what `aeron_set_windows_error` does. I'm not yet sure about the best way to do it, but would you agree to such a change in principle?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -195,6 +195,7 @@ elseif (CYGWIN)
 elseif (MSVC)
     add_definitions(-DWIN32)
     add_definitions(-D_CRT_SECURE_NO_WARNINGS)
+    add_definitions(-D_CRT_NONSTDC_NO_WARNINGS)
     add_definitions(-DNOMINMAX)
 
     if (${MSVC_VERSION} GREATER_EQUAL 1915)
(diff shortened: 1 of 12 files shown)


mjpt777:
It would be better to limit the scope of a PR. For example separate out addressing warnings from behavioural change. It is easier to discuss behaviour change in the context of each change.

ltrzesniewski:
Yes, I understand. I didn't want to flood you with PRs, but I can split this into 6 or 7 PRs if you prefer.

PR #867, opened 2020-02-20
[C] Fix mutexes on Windows

ltrzesniewski:
- Lock/unlock did not work as an indirection was missing
 - Made the function prototypes match the Linux version
 - Added `aeron_mutex_destroy`

Part of #863


ltrzesniewski:
There were some additional whitespace changes in `aeron_thread.h` in the original PR (I removed the empty lines between `aeron_thread_*` functions) which are not included here to avoid a merge conflict with #864.

PR #871, opened 2020-02-21
[C] Refactor error handling to improve error messages on Windows

ltrzesniewski:
The issue I'm trying to solve:

 - WinAPI functions return error codes through `GetLastError()`, not through `errno`
 - `strerror` is [not very useful](https://docs.microsoft.com/en-us/cpp/c-language/strerror-function?view=vs-2019) on Windows, so mapping error codes wouldn't work

I came up with this refactor which I hope you'll find acceptable. It ends up simplifying the `aeron_set_err` calls.

I've added an `aeron_set_err_from_last_err_code` function which will act the same way as `aeron_set_err`, except it will automatically append an error message based on the last error code. This will use `GetLastError` and `FormatMessage` on Windows, and `strerror(errno)` on other platforms.

 - This will still work on Windows with the C library file I/O calls, since they call WinAPI
 - WinSock returns errors through `WSAGetLastError`, but this is an alias to `GetLastError`
 - I added a couple `SetLastError` calls in a few places which set `errno` themselves
 - Some places like the `strto*` calls continue to use `errno`

### DIFF
--- a/aeron-driver/src/main/c/aeron_agent.c
+++ b/aeron-driver/src/main/c/aeron_agent.c
@@ -47,9 +47,7 @@ int aeron_idle_strategy_sleeping_init_args(void **state, const char *env_var, co
 {
     if (aeron_alloc((void **)state, sizeof(uint64_t)) < 0)
     {
-        int err_code = errno;
-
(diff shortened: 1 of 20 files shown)


tmontgomery:
Excellent PR. Thanks! Will merge once green

PR #875, opened 2020-02-25
[Java] Change the use of the isStartup flag for closing existing sessions.

mikeb01:
This causes a bug where a node that that dies and leaves that cluster and is then restarted and rejoins the cluster may end up having differing state from the leader as it's own isStartup flag gets used to determine if any sessions should be closed.  This records the isStartup flag from the leader in the leadership term event and uses that value to determine is session should be closed.

This can cause some strange behaviour around sessions and visibility of events from the cluster.  It was causing `ClusterTest::shouldRecoverWhenLastSnapshotIsInvalidAndWasBetweenTwoElections` to be intermittent.


mjpt777:
Good catch @mikeb01. However I don't think the change should be to the log. A better message to change would be `NewLeadershipTerm` id=53. This can carry the `isStartup` flag which can be used to update the late starting node. The decision needs to be made by the leader. As this message is transient we don't need to bump the version.

PR #876, opened 2020-02-25
[Java] Clear the cluster client decoders after use

mikeb01:
This helps prevent segfaults when using the debugger - the debugger calls toString on the decoder which will try to decode whatever data they happen to be pointing at and can happen at a point where that data is not valid.


mjpt777:
I've pushed a change to SBE to better support this. https://github.com/real-logic/simple-binary-encoding/commit/177e21df1fce779e2a9c521e8ce059c9bd9559ed

mjpt777:
Using the large "empty" buffer is wasteful and will cause GC write barriers to be invoked on update, null path is optimised for GC reference update on G1.

PR #877, opened 2020-02-25
[C] Consolidate min and tagged flow control in the C media driver.

mikeb01:
For parity with the Java media driver.

### DIFF
--- a/aeron-driver/src/main/c/aeron_driver_context.h
+++ b/aeron-driver/src/main/c/aeron_driver_context.h
@@ -125,7 +125,7 @@ typedef struct aeron_driver_context_stct
     sm_receiver_tag;
     struct
     {
-        int32_t receiver_group_min_size;                     /* aeron.flow.control.receiver.group.min.size = 0 */
+        uint32_t receiver_group_min_size;                    /* aeron.flow.control.receiver.group.min.size = 0 */
         uint64_t receiver_timeout_ns;                        /* aeron.flow.control.receiver.timeout */
         int64_t group_receiver_tag;                          /* aeron.flow.control.group.rtag = -1 */
     }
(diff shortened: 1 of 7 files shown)


aeron-driver/src/main/c/aeron_flow_control.c
```diff
@@ -166,7 +166,7 @@ void aeron_flow_control_extract_strategy_name_length(
     size_t *strategy_length)
 {
     const char *next_option = (const char *)memchr(options, ',', options_length);
-    *strategy_length = NULL == next_option ? options_length : labs((long)(next_option - options));
+    *strategy_length = NULL == next_option ? options_length : (size_t)labs((next_option - options));
```
mjpt777: This change reverts a previous fix. The long cast is correct.

aeron-driver/src/main/c/aeron_driver_context.h
```diff
@@ -125,7 +125,7 @@ typedef struct aeron_driver_context_stct
     sm_receiver_tag;
     struct
     {
-        int32_t receiver_group_min_size;                     /* aeron.flow.control.receiver.group.min.size = 0 */
+        uint32_t receiver_group_min_size;                    /* aeron.flow.control.receiver.group.min.size = 0 */
```
mjpt777: Why change the type to be inconsistent with Java types? Elsewhere we try to remain consistent.

mikeb01: Mostly because it removes a number of casts that changed signedness at point of use.

aeron-driver/src/main/c/aeron_flow_control.c
mikeb01: Reverted.

aeron-driver/src/main/c/aeron_driver_context.h
mikeb01: Changed it back to int32_t

PR #883, opened 2020-03-02
Name re-resolution in C media driver

mikeb01:
Name re-resolution:
- Introduce aeron_name_resolver.
- Add CSV table for doing name look ups (mostly as a stub for testing).
- Hook name_resolver into the appropriate places during URI parsing
- Update canonical form to match Java driver (using names instead of hex ips where appropriate).
- Re-resolve endpoint name for simple and manual MDC publications.
- Re-resolve control name for dynamic MDC publications.
- Add resolution change counters.


aeron-driver/src/main/c/aeron_driver_conductor.c
```diff
@@ -3050,6 +3071,50 @@ void aeron_driver_conductor_on_linger_buffer(void *clientd, void *item)
...
+{
+    aeron_driver_conductor_t *conductor = clientd;
+    aeron_command_re_resolve_t *cmd = item;
+    struct sockaddr_storage resolved_addr;
+    memset(&resolved_addr, 0, sizeof(resolved_addr));
+
+    if (aeron_name_resolver_resolve_host_and_port(
+        &conductor->name_resolver, cmd->endpoint_name, AERON_UDP_CHANNEL_ENDPOINT_KEY, true, &resolved_addr) < 0)
+    {
+        // TODO: What is the best error handling here, error counter for re-resolution failures???
```
mjpt777: Since it is a re-resolve it feels like a counter to me.

aeron-driver/src/main/c/aeron_driver_context.c
```diff
@@ -464,6 +465,12 @@ int aeron_driver_context_init(aeron_driver_context_t **context)
...
     }
 
+    if ((_context->name_resolver_supplier_func = aeron_name_resolver_supplier_load(
+        AERON_NAME_RESOLVER_SUPPLIER_DEFAULT)) == NULL)
+    {
+        return -1;
+    }
+
 #if defined(__linux__)
     snprintf(_context->aeron_dir, AERON_MAX_PATH - 1, "/dev/shm/aeron-%s", username());
```
mjpt777: Is username restricted in length?

aeron-driver/src/main/c/aeron_driver_context.c
```diff
@@ -526,6 +533,8 @@ int aeron_driver_context_init(aeron_driver_context_t **context)
     _context->nak_unicast_delay_ns = AERON_NAK_UNICAST_DELAY_NS_DEFAULT;
     _context->publication_reserved_session_id_low = AERON_PUBLICATION_RESERVED_SESSION_ID_LOW_DEFAULT;
     _context->publication_reserved_session_id_high = AERON_PUBLICATION_RESERVED_SESSION_ID_HIGH_DEFAULT;
+    _context->name_resolver_init_args = NULL;
+    _context->re_resolution_check_interval_ns = AERON_DRIVER_RERESOLUTION_CHECK_INTERVAL_DEFAULT;
```
mjpt777: Add NS to constant name to indicate units.

mjpt777:
LGTM.

aeron-driver/src/main/c/aeron_driver_conductor.c
mikeb01: @tmontgomery and I were thinking that the distinct error log based on name would be useful as it would quickly highlight a bad name.

aeron-driver/src/main/c/aeron_driver_context.c
mikeb01: `man useradd` limits to 32 characters.  Not sure about windows.

aeron-driver/src/main/c/aeron_driver_conductor.c
tmontgomery: need to know the name that failed, so, yes. Error log feels more useful.

tmontgomery: I've now added the error logging to the Java side.

aeron-driver/src/main/c/aeron_driver_context.c
tmontgomery: the length param of the snprintf will stop trashing aeron_dir buffer even if username is too long.

aeron-driver/src/main/c/media/aeron_receive_channel_endpoint.c
```diff
@@ -546,6 +558,24 @@ int aeron_receiver_channel_endpoint_validate_sender_mtu_length(
...
+    aeron_receive_channel_endpoint_t *endpoint,
+    int64_t now_ns,
+    aeron_driver_conductor_proxy_t *conductor_proxy)
+{
+    // MDS is not yet supported in the C media driver,
+    if (endpoint->conductor_fields.udp_channel->has_explicit_control &&
+        now_ns > endpoint->time_of_last_activity_ns + AERON_UDP_DESTINATION_TRACKER_DESTINATION_TIMEOUT_NS)
+    {
+        const char *endpoint_name = endpoint->conductor_fields.udp_channel->uri.params.udp.control;
+        struct sockaddr_storage *addr = &endpoint->conductor_fields.udp_channel->local_control;
```
tmontgomery: Changed this slightly in the Java version to track the current control address and update on re-resolution change. Same approach can be done here by storing the current sockaddr_storage for the control address in the endpoint.

aeron-driver/src/main/c/util/aeron_env.h
```diff
@@ -0,0 +1,40 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef AERON_AERON_ENV_H
+#define AERON_AERON_ENV_H
+
+#include <stdlib.h>
+
+inline int aeron_env_set(const char *key, const char *val)
```
tmontgomery: We already have this in aeron_properties_util.c as aeron_properties_setenv

tmontgomery: Ah. I see. That the properties were changed. nvm.

aeron-driver/src/main/c/aeron_driver_conductor.c
mjpt777: Given many different names could be used then maybe we might want to keep a resolution report similar to the loss report to aid debugging?

tmontgomery: that's an interesting idea. We could. I'm very interesting in getting some feedback from users and seeing what evolves.

mikeb01:
Brought up to date with the Java changes.

aeron-driver/src/main/c/aeron_driver_conductor.c
mikeb01: I'm inclined to leave it in the error log.  A resolution report could be useful, but it would be in addition to the error log.  Mostly because it reduces the number of places an operator needs to look when a system is not working.

PR #885, opened 2020-03-05
System test updates for the C media driver implementation of name re-resolution

mikeb01:
Test updates for PR #883.


mjpt777:
LGTM.

mikeb01:
I've added a test for the errors, which does a couple of funky things to support testing the error logs for both the C and Java drivers and still dump out the errors on failure.  I've moved around where the error is generated so that more context is captured on failure.

PR #890, opened 2020-03-17
C driver name resolver (#886)

mikeb01:
- Adds driver name resolver.
- Support sending/receiving resolution frames to support gossip protocol.
- Fixes race condition also seen in Java driver.
- Implemented log dissection for resolution frames.
- Some additional unit tests.


aeron-driver/src/main/c/aeron_name_resolver_driver.h
```diff
@@ -0,0 +1,44 @@
...
+#ifndef AERON_NAME_RESOLVER_DRIVER_H
+#define AERON_NAME_RESOLVER_DRIVER_H
+
+#define AERON_NAME_RESOLVER_DRIVER_SELF_RESOLUTION_INTERVAL_MS (1 * INT64_C(1000))
+#define AERON_NAME_RESOLVER_DRIVER_NEIGHBOUR_RESOLUTION_INTERVAL_MS (2 * INT64_C(2000))
+#define AERON_NAME_RESOLVER_DRIVER_TIMEOUT_MS (10 * INT64_C(1000))
+#define AERON_COUNTER_NAME_RESOLVER_NEIGHBORS_COUNTER_TYPE_ID (14)
+#define AERON_COUNTER_NAME_RESOLVER_CACHE_ENTRIES_COUNTER_TYPE_ID (15)
+
+int aeron_name_resolver_driver_set_resolution_header(
```
tmontgomery: shouldn't the name be "aeron_driver_name_resolver".... seems odd to have "driver" at the end.

tmontgomery:
This looks good to me. But do have 1 comment on naming.

aeron-driver/src/main/c/aeron_name_resolver_driver.h
mikeb01: I did tie myself in a few knots with naming conventions.  There are a couple of knock effects with changing that.  I have `aeron_name_resolver_default` and `aeron_name_resolver_csv_table`, which should also change to `aeron_default_name_resolver` and `aeron_csv_table_name_resolver` respectively.

mikeb01: But I agree, will make the change and update the PR.

mikeb01: I'm also tempted to rename the `aeron_name_resolver_driver_cache` to `aeron_name_resolver_cache`.  It's decoupled from the driver gossip and could potentially be a cache for any name resolver.

tmontgomery: Kinda what I went for in the Java version. Wanted to bring it out separately.... so, the naming would make sense to me.

Naming is hard.

PR #906, opened 2020-04-30
Append tags or unique id to canoncial form and support Wildcard ports

mikeb01:
- Appends unique index or tag to canonical form for udp channels that have wildcard fields.  Specifically this is MDS URIs (`aeron:udp?control-mode=manual`) and URIs that use a wildcard port (i.e. 0) e.g. `aeron:udp?endpoint:192.168.10.2:0`.
- Adds counters for each address/port binding on the sender and receiver side.  The counter will include the string representation in it's key, space is set up such that it won't be truncated.
- Add `bindAddressAndPorts()` method to Subscription and Publication so that clients can retrieve the bind addresses and ports once the channel and destination is active.
- Updates the format of the address/port pair to surround the address in square brackets `[`, `]` when using IPv6 addresses.

Probably some further discussion on naming required.


mjpt777:
Rather than depend on Agrona snapshot you could edit the metadata buffer directly and then we can patch up later.

mikeb01:
I can do that, but it would also require reflecting to get access to some of the private fields of the AtomicCounter.  As long as that is okay I can do that tomorrow.

mjpt777:
No need for reflection. Use the counter id to calculate the offset in the metadata then use constants.

mikeb01:
How do you get access to the metadata buffer from within the ReceiveChannelEndpoint?

mjpt777:
`context.countersMetaDataBuffer()`

mikeb01:
Spotted that just as I posted.  Will do.

mikeb01:
I've updated the code to no longer depend on agrona 1.5 and updating the name to refer to the addresses as local socket addresses.

mjpt777:
Thanks Mike. We'll need to remember the C++ client needs updated too.

PR #916, opened 2020-05-07
Improve cmake building files

lygstate:
Convert HdrHistogram_c googletest zlib to use git submodule instead zip files.
Getting cppbuild-driver-vs.cmd and cppbuild-vs.cmd to be able running with double click.


.github/workflows/ci-low-cadence.yml
```diff
@@ -65,6 +65,8 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v2
+        with:
```
tmontgomery, line 68: we do this? We don't want submodules.

lygstate: Because unzip the files are hard to maintain

tmontgomery:
The LGTM alerts need to be fixed.

We do not want submodules.
We do want to keep zlib and the deps as archives so we use specific versions we have vetted and make the build self contained.
We do not want to change defaults and default config for CMake nor the build scripts in this way.

In general, this is the wrong direction.

lygstate:
Is that binary files contained in source tree are good idea? That's wierd.

PR #918, opened 2020-05-07
Do not review and do not close it, just used to trigger the CI

lygstate:


mjpt777:
To be clear we do not agree with this approach. Happy to discuss what you would like to achieve in the Gitter channel.

PR #919, opened 2020-05-09
Enable C Media Driver  system test on Win32

lygstate:


lygstate:
I need to know how to make sure the driver are started?
Does `isActive = context.isDriverActive(context.driverTimeoutMs(), logger);`
is what I wanted?

mjpt777:
The client waits for the CnC file to be ready. I've pushed a change to the Java client which should make it more reliable on Windows.

mjpt777:
There is still too many things changed in one PR with this. Please make things more discrete.

PR #921, opened 2020-05-10
Do not pollute the global state by BUILD_SHARED_LIBS and CMAKE_WINDOW…

lygstate:
…S_EXPORT_ALL_SYMBOLS

These two global variable can be set on the target properties.


tmontgomery:
These are changes requested by a active supported users. And we want to have them enabled by default.

lygstate:
This change didn't change the default behavour of aeron, just getting aeron be easier interating into other project

PR #923, opened 2020-05-10
[C/CPP] Improve cppbuild/cppbuild

lygstate:
Set BUILD_CONFIG/BUILD_DIR for Debug build.
-DAERON_SYSTEM_TESTS=OFF not necessary for slow-system-tests, we can use --no-system-tests instead

### DIFF
--- a/cppbuild/cppbuild
+++ b/cppbuild/cppbuild
@@ -1,7 +1,8 @@
 #!/usr/bin/env bash
 
 SOURCE_DIR="`pwd`"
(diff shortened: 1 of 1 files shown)


tmontgomery:
Checks were not successful.

Also, do not change cppbuild defaults by removing setting of AERON_SYSTEM_TESTS.

lygstate:
As I said in gitter, the checks not caused by me, you may need have a look at the failure

PR #927, opened 2020-05-10
[CPP] fix std::array not initialized bug

lygstate:
std:array<> not zero-initialized without calling the constructor directly.

Refer to https://stackoverflow.com/questions/18295302/default-initialization-of-stdarray


tmontgomery:
While this is fine, the commit from a8f418bde78573cf83f218e6aa74283e61ad2437 explicitly set the buffers to 0 for clarity and so it is done for each setup. So, this is unnecessary. But, will include for correctness.

PR #928, opened 2020-05-10
[C] Trying to extract zip source file directly, improve cmake build procedure

lygstate:
[C] Trying to extract zip source file directly, so we can simplify the build procedure on Win32. And also reduce dependencies on other platform.


tmontgomery:
This is the wrong direction. The way it is done now is a better option for us at the moment.

PR #929, opened 2020-05-10
[C] '+' and ':' are not valid win32 path symbol. replace them with '-'

lygstate:


lygstate:
Updated with exist style

tmontgomery:
This needs to be defined for Windows specifically.

lygstate:
I didn't do it for consistence across platform, and also the Java Version use the same schema.

tmontgomery:
Ah. Very good. I had missed that. Thanks!

PR #930, opened 2020-05-10
[C] make sure the cnc.dat have valid file length before init mpsc

lygstate:

### DIFF
--- a/aeron-client/src/main/c/aeron_context.c
+++ b/aeron-client/src/main/c/aeron_context.c
@@ -35,6 +35,8 @@
     #pragma clang diagnostic ignored "-Wunused-function"
 #endif
 
+#define AERON_FILE_PAGE_SIZE_DEFAULT (4 * 1024)
+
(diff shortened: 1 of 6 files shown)


tmontgomery:
This is a dupe of #926 . Close one or the other.

mjpt777:
@lygstate  Do you have evidence, or some failing test, that illustrates how a mapped file is not either zero or the correct length?

lygstate:
@mjpt777 Easy to reproduce for the Java Client with C Media driver, just enable the Windows System  C Media Driver test, then the error would happen. That's way I spent so much time to fix it. I need keep the CI clean.

tmontgomery:
Again, C windows driver is not supported and may not work. Does this happen on Windows with Java and Java client?

lygstate:
The C media driver are working fine for me now, after fixing those issues, especially #932 and #933 (These two issue not only happens on Windows, but on all platform) . Java media driver with Java client didn't have any problem.

PR #931, opened 2020-05-10
[C] Use custom aeron_uuid

lygstate:
Use custom uuid geneartor and remove the system uuid deps.
Using the true random number generator as much as possible.
And this also make uuid works under Win32

### DIFF
--- a/aeron-client/src/main/c/util/aeron_bitutil.h
+++ b/aeron-client/src/main/c/util/aeron_bitutil.h
@@ -134,6 +134,4 @@ inline int32_t aeron_find_next_power_of_two(int32_t value)
     return value + 1;
 }
 
-int32_t aeron_randomised_int32();
-
 #endif //AERON_BITUTIL_H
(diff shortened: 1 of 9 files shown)


tmontgomery:
A few initial issues and questions before a more thorough review.

* The coding style does not match the existing style and function/variable/etc. naming is not respected.
* The code looks like it comes from somewhere else without attribution. Is this code you wrote or code you copied from somewhere else?
* Why is a dependency on uuid bad for those platforms that have it? Especially when it gives us the uniqueness we desire.
* Why remove arc4random?

lygstate:
The randombytes are using https://github.com/dsprenkels/randombytes for true random value generator.
The the style are not changed.

UUID is for generating a unique identifier for the computer, so that's important using a true random-generator across platform.
Remove the dependencies to libuuid make aeron to be a more portable message queue.
Other message system like nanomsg are also have zero dependencies.

tmontgomery:
We do not want to bring in other code with the MIT license. In addition, this is not a problem except for Windows C driver at this time and Windows C driver is not supported at this time.

PR #932, opened 2020-05-11
&stream_interest->image_by_session_id_map, are hash header->session_id,

lygstate:


lygstate:
This is broken in the right way. We may need better unittest to protect it.

PR #935, opened 2020-05-12
[C/CPP] Improve cppbuild/cppbuild for support both Debug and Release …

lygstate:
…build


tmontgomery:
I'm ok with this. If @mjpt777 is.

PR #936, opened 2020-05-12
Udp transport abstraction improvement

lygstate:
This patch removes a lot of duplicated code

The ExternalProject_Add are easy to cause build failure across platform.
And also the the Debug/Release build are not working at the same time on Windows at least

### DIFF
--- a/aeron-driver/src/main/c/media/aeron_udp_channel_transport.h
+++ b/aeron-driver/src/main/c/media/aeron_udp_channel_transport.h
@@ -32,8 +32,6 @@ typedef struct aeron_udp_channel_transport_stct
 }
 aeron_udp_channel_transport_t;
 
-struct mmsghdr;
-
 int aeron_udp_channel_transport_init(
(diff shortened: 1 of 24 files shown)


aeron-driver/src/main/c/media/aeron_udp_channel_transport.h
```diff
@@ -49,22 +47,15 @@ int aeron_udp_channel_transport_close(aeron_udp_channel_transport_t *transport);
 
 int aeron_udp_channel_transport_recvmmsg(
     aeron_udp_channel_transport_t *transport,
-    struct mmsghdr *msgvec,
-    size_t vlen,
+    aeron_udp_channel_recv_buffers_t *recv_buffers,
```
tmontgomery, line 50: changing bindings is not something we want to do right now.

tmontgomery:
changing the transport bindings is not something we want. Also, this is unnecessary and does not make things cleaner nor better. Thanks, but no thanks.

PR #922, opened 2020-05-10
[C] Improve cppbuild/cppbuild-vs.cmd to match cppbuild/cppbuild

lygstate:
Getting cppbuild-vs.cmd to be consistence with cppbuild-vs.
Getting java_system_tests_c_media_driver and java_slow_system_tests_c_media_driver works under win32
Enable system-tests for Win/MSVC CI

Signed-off-by: Yonggang Luo <luoyonggang@gmail.com>

PR #937, opened 2020-05-12
[C] Trying to extract zip source file directly, so we can simplify th…

lygstate:
The CI random failing on MSVC target
refer to 
https://github.com/real-logic/aeron/pull/922/checks?check_run_id=665696447


tmontgomery:
This is not necessary or something we want to do.

PR #938, opened 2020-05-12
reduce msvc warning

lygstate:


tmontgomery:
We will be redoing the C++ API to use the C API. So, we don't want to silence this warning at the moment because we want to see it.

PR #941, opened 2020-05-18
Add backoff period after exception in ClusterBackupAgent - 2nd attempt

JPWatson:
Reapplies #939 after test failure and adds an additional parameter to control cool off period.

30 seconds ensures any lingers resources will be closed.

### DIFF
--- a/aeron-cluster/src/main/java/io/aeron/cluster/ClusterBackup.java
+++ b/aeron-cluster/src/main/java/io/aeron/cluster/ClusterBackup.java
@@ -252,6 +252,16 @@ public final class ClusterBackup implements AutoCloseable
         public static final String CLUSTER_BACKUP_PROGRESS_TIMEOUT_PROP_NAME = "aeron.cluster.backup.progress.timeout";
 
         /**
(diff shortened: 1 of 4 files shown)


aeron-cluster/src/main/java/io/aeron/cluster/ClusterBackup.java
```diff
@@ -251,6 +251,16 @@ public void close()
          */
         public static final String CLUSTER_BACKUP_PROGRESS_TIMEOUT_PROP_NAME = "aeron.cluster.backup.progress.timeout";
 
+        /**
+         * Interval at which the cluster backup is re-initialised after an exception has been thrown.
+         */
+        public static final String CLUSTER_BACKUP_COOLOFF_INTERVAL_PROP_NAME = "aeron.cluster.backup.cooloff.interval";
```
mjpt777, line 257: COOLDOWN is already used in Aeron so would be a better fit.

JPWatson: 👍 will rename.

PR #948, opened 2020-05-28
log is not replayed if cluster was shut down and the snapshot is invalid

vdaniloff:
If cluster was terminated with the shapshot-and-shutdown operation, and the snapshot is invalidated, then even though the snapshot is invalidated, replay is checked with last commited position, but since snapshot with that position isn't loaded service is restored in incorrect state.

this may be not the optimal fix, but it highlights the problem. happy to apply any review notes.

### DIFF
--- a/aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModuleAgent.java
+++ b/aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModuleAgent.java
@@ -1454,17 +1454,21 @@ class ConsensusModuleAgent implements Agent
 
     LogReplay newLogReplay(final long logPosition, final long appendPosition)
     {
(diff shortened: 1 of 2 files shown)


vdaniloff:
wrong test version was commited

mjpt777:
Is there an issue here?

vdaniloff:
I believe it is, but I am unable to easily fix this. I only left test for now with some comments. Unfortunately it isn't enough to only restore logic of replay log.

this test passes in 1.27.0

mjpt777:
Thanks for the test. I've fixed and merged the test with this commit. https://github.com/real-logic/aeron/commit/e089ea9c322b548511f98ecaf85fc785bfe2420e

PR #953, opened 2020-06-07
[C] replacing getopt with a posix compatible version from https://github.com/skeeto/getopt/blob/master/getopt.h

lygstate:
/* Taken from https://github.com/skeeto/getopt/blob/master/getopt.h */
/* A minimal POSIX getopt() implementation in ANSI C
 *
 * This is free and unencumbered software released into the public domain.
 *
 * This implementation supports the convention of resetting the option
 * parser by assigning optind to 0. This resets the internal state
 * appropriately.
 *
 * Ref: http://pubs.opengroup.org/onlinepubs/9699919799/functions/getopt.html
 */

### DIFF
--- a/aeron-client/src/main/c/util/aeron_strutil.c
+++ b/aeron-client/src/main/c/util/aeron_strutil.c
@@ -19,6 +19,7 @@
 #define _GNU_SOURCE
 #endif
 
+#include <ctype.h>
 #include <time.h>
 #include <stdio.h>
 #include <inttypes.h>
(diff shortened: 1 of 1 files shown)


lygstate:
The current getopt are not posix compatible and can not parse -Dxxx properly.

tmontgomery:
We've talked and we are not comfortable with including source code that contains The Unlicense. We would be comfortable with an Apache or BSD licensed version or maybe if the original author is comfortable with using CC0 instead of The Unlicense.

PR #957, opened 2020-06-11
[C] Add logging for the untethered subscription state change

vyazelenko:
- Logging of the untethered subscription state change
- Add missing options to the `cppbuild`'s help output
- Fix logging constants to be power of two

For the logging changes I'm not sure if `aeron_driver_context.c` is the right place for the implementation of the `aeron_untethered_subscription_state_change` function or if it would be better to define it as `inline` in the `aeron_driver_common.h`.

### DIFF
--- a/aeron-driver/src/main/c/aeron_driver_common.h
+++ b/aeron-driver/src/main/c/aeron_driver_common.h
@@ -67,6 +67,20 @@ typedef struct aeron_tetherable_position_stct
 }
 aeron_tetherable_position_t;
 
+typedef void (*aeron_untethered_subscription_state_change_func_t)(
+        aeron_tetherable_position_t *,
+        int64_t now_ns,
+        aeron_subscription_tether_state_t new_state,
+        int32_t stream_id,
+        int32_t session_id);
+
+void aeron_untethered_subscription_state_change(
+        aeron_tetherable_position_t *tetherable_position,
+        int64_t now_ns,
+        aeron_subscription_tether_state_t new_state,
+        int32_t stream_id,
+        int32_t session_id);
+
 typedef struct aeron_subscribable_stct
 {
     size_t length;
--- a/aeron-driver/src/main/c/aeron_driver_common.h
+++ b/aeron-driver/src/main/c/aeron_driver_common.h
@@ -74,12 +74,16 @@ typedef void (*aeron_untethered_subscription_state_change_func_t)(
         int32_t stream_id,
         int32_t session_id);
 
-void aeron_untethered_subscription_state_change(
+inline static void aeron_untethered_subscription_state_change(
         aeron_tetherable_position_t *tetherable_position,
         int64_t now_ns,
         aeron_subscription_tether_state_t new_state,
         int32_t stream_id,
-        int32_t session_id);
+        int32_t session_id)
+{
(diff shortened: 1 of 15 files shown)


aeron-driver/src/main/c/aeron_driver_common.h
```diff
@@ -67,6 +67,20 @@ typedef struct aeron_tetherable_position_stct
...
+        int32_t session_id);
+
+void aeron_untethered_subscription_state_change(
+        aeron_tetherable_position_t *tetherable_position,
+        int64_t now_ns,
+        aeron_subscription_tether_state_t new_state,
+        int32_t stream_id,
+        int32_t session_id);
+
 typedef struct aeron_subscribable_stct
```
mikeb01, line 84: This is more a general design direction question (perhaps for @tmontgomery and @mjpt777).  Would there be a way to add the agent logging without needed function pointers at each place we need to log?  I tend to find the function pointers can be hard to follow.  The benefit of agent based logging in Java is that the code doesn't change (i.e. doesn't get cluttered with log statements) and the performance impact is negligible.  However in C introducing a function pointer just for logging adds both cognitive and performance overhead.

tmontgomery: I agree. However, for hooking, we can't LD_PRELOAD on functions that are not entry points. Or we could do that as was done previously with sendmsg/recvmsg/sendmmsg/recvmmsg.

aeron-driver/src/main/c/agent/aeron_driver_agent.c
```diff
@@ -316,15 +335,40 @@ void aeron_driver_agent_incoming_msg(
...
+            sizeof(aeron_driver_agent_untethered_subscription_state_change_log_header_t));
+}
+
 int aeron_driver_agent_interceptor_init(void **interceptor_state, aeron_udp_channel_transport_affinity_t affinity)
 {
     return 0;
 }
 
-int aeron_driver_agent_context_init(aeron_driver_context_t *context)
+static int init_logging_events_interceptors(aeron_driver_context_t *context)
```
tmontgomery, line 370: even static, this should be prepended with aeron_ for namespace cleanliness.

tmontgomery:
`aeron_untethered_subscription_state_change` can probably be inlined and placed in a header. But rest looks pretty good.

vyazelenko:
@tmontgomery I've declared function `aeron_untethered_subscription_state_change` as `inline static` in the `aeron_driver_common.h` and it did the trick.

After the changes I get two warnings:
```
aeron-driver/src/main/c/agent/aeron_driver_agent.c:121:13: warning: unused function 'aeron_free_logging_ring_buffer' [-Wunused-function]
static void aeron_free_logging_ring_buffer()
            ^
aeron-driver/src/main/c/agent/aeron_driver_agent.c:130:13: warning: unused function 'aeron_set_logging_mask' [-Wunused-function]
static void aeron_set_logging_mask(uint64_t new_mask)
            ^
[ 48%] Building C object aeron-driver/src/main/c/CMakeFiles/aeron_driver.dir/agent/aeron_driver_agent.c.o
```
The functions are being used only in tests. Not sure if there is a better way to do it.

PR #964, opened 2020-06-26
[C++] Refactor includes and add option to play nice with CMake and large projects to avoid header name conflicts

indiosmo:
Some of the aeron-client headers have very common names like `Subscription.h` and `Context.h` which could lead to include conflicts in large projects.

A common practice for C++ libs is to use a stuttering directory structure (e.g. boost, folly, fmt, date, tl_expected, etc) so that the user can `#include "library/header.h"` instead of `#include "header.h"` directly.

This would be a breaking change though so the alternative I came up with was to add the **AERON_CLIENT_STUTTERING_HEADERS** option to CMakeLists.txt. This option installs the headers under an `aeron-client` directory during the build so the user can `#include "aeron-client/Aeron.h"` for example.

To make this work without breaking existing code I had to make all the local includes relative to the file path.
I'm not sure how you feel about this stylistically speaking but it's a compromise that allows existing code to continue as is while allowing new code to use the stuttering option.

Includes were also inconsistent between <> and "" so I changed all local includes to "".

This solution also plays nice with CMake FetchContent.

### Example usage:

CMakeLists.txt:
```
# aeron
FetchContent_Declare(
  aeron
  GIT_REPOSITORY git@github.com:e2m-investimentos/aeron.git
  GIT_TAG master
)

set(AERON_CLIENT_STUTTERING_HEADERS ON)

FetchContent_MakeAvailable(aeron)

install(
  TARGETS aeron aeronmd aeron_client_shared aeron_driver
  RUNTIME DESTINATION ${MY_BIN_INSTALL_DIR}/aeron
  LIBRARY DESTINATION ${MY_LIB_INSTALL_DIR}
  ARCHIVE DESTINATION ${MY_LIB_INSTALL_DIR}
)
```

MyClass.h:
```
#include "aeron-client/Aeron.h"
....
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -45,6 +45,7 @@ option(AERON_INSTALL_TARGETS "Enable installation step" ${STANDALONE_BUILD})
 if (UNIX)
     option(AERON_ENABLE_NONSTANDARD_OPTIMIZATIONS "Enable Ofast for release builds" ${STANDALONE_BUILD})
 endif ()
+option(AERON_CLIENT_STUTTERING_HEADERS "Follow boost stuttering directory structure to avoid header name conflicts" OFF)
 
 unset(STANDALONE_BUILD)
 
--- a/aeron-client/src/main/cpp/Aeron.h
+++ b/aeron-client/src/main/cpp/Aeron.h
@@ -17,13 +17,13 @@
 #ifndef INCLUDED_AERON_H
 #define INCLUDED_AERON_H
 
-#include <util/Exceptions.h>
+#include "util/Exceptions.h"
 #include <iostream>
 #include <thread>
 #include <random>
-#include <concurrent/logbuffer/TermReader.h>
-#include <util/MemoryMappedFile.h>
-#include <concurrent/broadcast/CopyBroadcastReceiver.h>
+#include "concurrent/logbuffer/TermReader.h"
+#include "util/MemoryMappedFile.h"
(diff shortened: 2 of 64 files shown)


tmontgomery:
The install can be configured in a number of ways. So, changing the current structure is not necessary. If you want a specific type of setup, then do it in your own project install.

The `<>` and `""` are done for system vs. local headers. And is how we want it.

Also, the use of relative paths in includes is a bad idea all around.

Thanks, though. But this is not changes we want to make.

indiosmo:
Understood. Just a comment on the includes, when I mentioned the inconsistency it was for local headers specifically, e.g.

AtomicArrayUpdater.h:
`#include "concurrent/Atomic64.h"`

"HeaderFlyweight.h":
`#include <concurrent/AtomicBuffer.h>`

tmontgomery:
Sometimes CLion does things automatically. So, not ideal. :) I try to fix them as I find them. Thanks for pointing that one out!

indiosmo:
Also for reference in case anyone looks at this issue in the future, I don't think this is solvable from outside the project without poluting the include path.

I could certainly install the headers somewhere else for the sake of my own includes but the transitive includes would then fail, i.e. if I have `aeron-client/` in my include path I can `#include "aeron-client/Aeron.h"` but then Aeron.h would fail at `#include "ClientConductor.h"` for example.

I suppose if you happen to have another dependency with the same header names you'd have to choose one of the conflicting dependencies to fork.

PR #965, opened 2020-06-29
C Media Driver Unit Test Refactoring

mikeb01:
- Use a common base for receiver tests (e.g. data_packet_dispatcher and aeron_publication_image tests).
- Use GMock for callback for more flexible matching.
- Parameterise common tests between IPC and Network conductor tests.
- Fix bug around session id tracking for IPC publications.
- Decouple C media driver from the C++ API and remove linkage in build


aeron-driver/src/test/c/aeron_system_test.cpp
```diff
@@ -42,15 +45,39 @@ class SystemTest : public testing::Test
 
 TEST_F(SystemTest, shouldReclaimSubscriptionWhenOutOfScopeAndNotFound)
```
tmontgomery, line 46: This test was testing something specific to the C++ API involved with NOT doing a find on the Subscription. This test does not act that way and C API does not work the same. So, it might be better to just remove the test at this point.

mikeb01: Okay, makes sense.  I'll move it over to the C++ Tests.

mikeb01: I've moved the test across.  This has meant duplication of the `EmbededMediaDriver.h`.

tmontgomery:
This looks good to me

PR #968, opened 2020-07-05
[C] Don't allow session ids on network publications to be re-used until the publication is completely closed.

mikeb01:
Add system test to show external behaviour as well.

### DIFF
--- a/aeron-system-tests/src/test/java/io/aeron/SessionSpecificPublicationTest.java
+++ b/aeron-system-tests/src/test/java/io/aeron/SessionSpecificPublicationTest.java
@@ -29,7 +29,6 @@ import org.agrona.ErrorHandler;
 import org.agrona.concurrent.UnsafeBuffer;
 import org.agrona.concurrent.status.CountersReader;
 import org.junit.jupiter.api.AfterEach;
-import org.junit.jupiter.api.Test;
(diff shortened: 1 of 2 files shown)


aeron-system-tests/src/test/java/io/aeron/SessionSpecificPublicationTest.java
```diff
@@ -155,4 +158,53 @@ public void shouldNotCreateNonExclusivePublicationsWithDifferentSessionIdsForThe
...
+            .sessionId(SESSION_ID_1)
+            .build();
+
+        final Publication publication1 = aeron.addPublication(channel, STREAM_ID);
+        final int positionLimitId = publication1.positionLimitId();
+        assertEquals(CountersReader.RECORD_ALLOCATED, aeron.countersReader().getCounterState(positionLimitId));
+
+        publication1.close();
+
+        assertThrows(RegistrationException.class, () ->
```
mjpt777, line 179: This is a race so likely to be non-deterministic. Better to create a subscription and wait for it to connect but don't drain it.

mjpt777: To clarify that would be a spy subscription and spies simulate connection enabled on the driver with one offer. The publication would then go into draining after close.

PR #978, opened 2020-07-13
C api additional features

mikeb01:
- Metadata access functions for counters reader.
- Implement aeron_exclusive_publication_is_connected.
- Have add and remove destination return correlationId for compatibility with C API.
- Add a number of accessor fields for aeron_header_t.


aeron-client/src/main/c/aeron_exclusive_publication.c
```diff
@@ -567,30 +580,30 @@ int64_t aeron_exclusive_publication_position_limit(aeron_exclusive_publication_t
     return aeron_counter_get_volatile(publication->position_limit);
 }
 
-int aeron_exclusive_publication_add_destination(aeron_exclusive_publication_t *publication, const char *uri)
+int64_t aeron_exclusive_publication_add_destination(aeron_exclusive_publication_t *exclusive, const char *uri)
```
mjpt777, line 583: `exclusive` here is not a good change.

aeron-client/src/main/c/aeronc.h
```diff
@@ -34,6 +34,20 @@ typedef struct aeron_buffer_claim_stct aeron_buffer_claim_t;
 typedef struct aeron_publication_stct aeron_publication_t;
 typedef struct aeron_exclusive_publication_stct aeron_exclusive_publication_t;
 typedef struct aeron_header_stct aeron_header_t;
+typedef struct aeron_header_values_stct
```
mjpt777, line 37: Should be specifying the packing details here if it is an overlay?

aeron-client/src/main/c/aeron_client_conductor.c
```diff
@@ -2085,15 +2085,16 @@ int aeron_client_conductor_offer_remove_command(
     return 0;
 }
 
-int aeron_client_conductor_offer_destination_command(
+int64_t aeron_client_conductor_offer_destination_command(
```
mjpt777, line 2088: Style wise should we always return `int` for success/failure code or dual purpose with the correlation id? We could take an arg that is a `uint64_t *` for the correlation id.  Be good to discuss.

tmontgomery: I would say `int` return and take a `int64_t *` to return the id. This is an internal matter. But could be easier.

mikeb01: I've gone with the return int for success/failure and using a pointer for the correlation_id.  I've also surfaced that to the API methods that call down from there e.g. `aeron_exclusive_publication_add_destination` so the style is consistent.

aeron-client/src/main/c/aeronc.h
mikeb01: Done.

PR #979, opened 2020-07-15
interceptor changes

tmontgomery:
To pass back and forth to see if comprehensive enough.

@mikeb01 see if this seems logical.

Interceptor changes to include transport in the incoming path and various notifications of transport/publication/image events, and addition of per interceptor state to transport.

### DIFF
--- a/aeron-driver/src/main/c/media/aeron_udp_channel_transport.h
+++ b/aeron-driver/src/main/c/media/aeron_udp_channel_transport.h
@@ -29,6 +29,7 @@ typedef struct aeron_udp_channel_transport_stct
     void *dispatch_clientd;
     void *bindings_clientd;
     void *destination_clientd;
+    void *interceptor_clientds[AERON_UDP_CHANNEL_TRANSPORT_MAX_INTERCEPTORS];
 }
 aeron_udp_channel_transport_t;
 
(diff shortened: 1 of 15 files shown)


aeron-driver/src/main/c/media/aeron_udp_channel_transport.h
```diff
@@ -29,6 +29,7 @@ typedef struct aeron_udp_channel_transport_stct
     void *dispatch_clientd;
     void *bindings_clientd;
     void *destination_clientd;
+    void *interceptor_clientds[AERON_UDP_CHANNEL_TRANSPORT_MAX_INTERCEPTORS];
```
mikeb01, line 32: I can't seem to find an linkage between the interceptor and this clientds array.  I'm guessing each interceptor would need to be assigned an index?  Would this get built dynamically and assigned to each interceptor during `aeron_udp_channel_interceptor_bindings_load`?

tmontgomery: This array would only be called by interceptors themselves to get/set the pointers. If they need to actually use it. Logging and loss don't need to use it. So, no usage with them. But other interceptors will.

mikeb01: I guess the question was more around how would an interceptor know which slow is theirs?  I.e. how does a given interceptor know what its index is?

tmontgomery: That has to be handled by the interceptors. I picked max 2 for now so that we can use 0 and 1 ourselves for new features. We will give them IDs. Anyone else will have to assign there own. It's not meant to be foolproof. I've debated on adding a sanity check (i.e. IDs fit into range and are unique. We can add that on the paths init if we need to.

mikeb01: I had envisaged a similar solution, but was going to make the id assignment something that happens during the interceptor loading process.  We build an array with length equal to the number of interceptors configured then assign a index to each interceptor during the load process.

tmontgomery: Considered that, but felt like it would just be simpler to have a static compile time offset that can be used. And faster as well. If we need to make it variable, we can, though.

mikeb01: Would it make sense to setup a registry of the ids, similar to `aeron_system_counter_enum_stct`?

PR #982, opened 2020-07-21
Use the shared aeron_clock_gettime_monotonic, aeron_clock_gettime_realtime and aeron_epoch_clock

lygstate:
int aeron_clock_gettime_monotonic(struct timespec *tp);
int aeron_clock_gettime_realtime(struct timespec *tp);
Convert aeron_agent_epoch_clock to aeron_epoch_clock

### DIFF
--- a/aeron-client/src/main/c/util/aeron_clock.c
+++ b/aeron-client/src/main/c/util/aeron_clock.c
@@ -23,52 +23,106 @@
 #include <time.h>
 #include "aeron_alloc.h"
 #include "util/aeron_bitutil.h"
+#include "util/aeron_clock.h"
 #include "concurrent/aeron_atomic.h"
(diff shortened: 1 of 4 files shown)


aeron-client/src/main/c/util/aeron_clock.c
```diff
@@ -29,46 +29,21 @@
 int64_t aeron_nano_clock()
 {
     struct timespec ts;
-#if defined(__CYGWIN__) || defined(__linux__)
```
tmontgomery, line 32: These should not call into aeron_windows.c for there clock. And should be put back so that they directly call the system calls.

lygstate: Agreed, updated.

PR #984, opened 2020-07-21
Move place of the following function:

lygstate:
They should belongs to utils and can be access by both driver and client
+/**
+ * Return time in nanoseconds for machine. Is not wall clock time.
+ *
+ * @return nanoseconds since epoch for machine.
+ */
+int64_t aeron_nano_clock();
+
+/**
+ * Return time in milliseconds since epoch. Is wall clock time.
+ *
+ * @return milliseconds since epoch.
+ */
+int64_t aeron_epoch_clock();
+


tmontgomery:
Do not change the C API in aeronmd.h without prior discussion and agreement.

PR #992, opened 2020-07-24
[C/C++] C++ wrapper of the C API.

mikeb01:
An initial implementation of the wrapper.  It is still quite rough (lots of TODOs and UnsupportedOperationExceptions).  I've attempted to preserve compatibility with the existing C++ API for add(Publication|Subscription|Counter), but there is a subtle behaviour difference in that the int64_t that is returned from those methods is reintepreted pointer rather than the actual registration_id.  Testing so far has included ensure that Basic(Publisher|Subscriber), StreamingPublisher, RateSubscriber, Ping and Pong all work.  The build will create additional binaries for each of these suffixed with a `W`, e.g. `PingW`. So that performance can be compared with the pure C++ API.


lygstate:
Would this replace the C++ client?

mikeb01:
> Would this replace the C++ client?

Eventually, yes.

lygstate:
How about directly modify the C++ client, I think that would be easier to review, and not easy to prone.

mikeb01:
Having a side by side implementation allows a smoother transition for users, who may not want to migrate immediately.  The C API and wrapper will take some time to stabilise.  Parallel implementations make it easier comparative testing and benchmarking.

lygstate:
I see some C++ files in pull request, how about remove all .cpp files in C++ wrapper? getting C++ wrapper to be header only library, so we doesn't need to handling the dynamic library hell, and only need link to aeronc.dll

mikeb01:
> getting C++ wrapper to be header only library,

It is worth considering.  At the moment not all of the functionality required is available via the `aeronc.h`, so I'm adding as I go, but I want to be very careful that I don't cause a transient include of anything from the C API other than `aeronc.h`.  I will need to see what is left over once it is feature complete and I've removed the redundant code.

mjpt777:
There are now a significant number of changes and fixes to the existing C++ client that need to migrated across and we need to be careful not to miss them.

aeron-archive/src/test/cpp/ArchiveTestUtil.cpp
```diff
@@ -13,31 +13,26 @@
...
-
-    static std::int64_t getFileSize(const char *filename)
-    {
-        return aeron_file_length(filename);
-    }
-};
-
-}}
+    struct stat statInfo{};
+    return ::stat(path, &statInfo) == 0;
```
mjpt777, line 34: Is there a Windows implementation?

mikeb01: Is now :-).

aeron-client/src/main/cpp_wrapper/Aeron.h
```diff
@@ -0,0 +1,985 @@
...
+
+    /**
+     * Indicate if the instance is closed and can not longer be used.
+     *
+     * @return true is the instance is closed and can no longer be used, otherwise false.
+     */
+    inline bool isClosed()
+    {
+        return false;
+//        throw UnsupportedOperationException("Need client close on C API", SOURCEINFO);
```
mjpt777, line 105: Is something missing here?

aeron-client/src/main/cpp_wrapper/Aeron.h
```diff
@@ -0,0 +1,985 @@
...
+        on_unavailable_image_t & callback = *reinterpret_cast<on_unavailable_image_t *>(clientd);
+        Image imageWrapper(subscription, image);
+        callback(imageWrapper);
+    }
+
+    static void onAvailableCounterCallback(
+        void *clientd, aeron_counters_reader_t *counters_reader, int64_t registration_id, int32_t counter_id)
+    {
+        CountersReader reader = CountersReader(counters_reader);
+        on_available_counter_t& callback = *reinterpret_cast<on_available_counter_t *>(clientd);
```
mjpt777, line 965: Reference on type inconsistent with style.

aeron-client/src/main/cpp_wrapper/ClientConductor.h
```diff
@@ -0,0 +1,75 @@
...
+namespace aeron
+{
+
+typedef std::function<long long()> epoch_clock_t;
+typedef std::function<long long()> nano_clock_t;
+
+class ClientConductor
+{
+public:
+    ClientConductor(aeron_t* aeron) : m_aeron(aeron)
```
mjpt777, line 31: Style

aeron-client/src/main/cpp_wrapper/Counter.h
```diff
@@ -0,0 +1,76 @@
...
+ */
+
+#ifndef AERON_COUNTER_H
+#define AERON_COUNTER_H
+
+#include <cstdint>
+#include <memory>
+#include <atomic>
+
+#include <util/Index.h>
```
mjpt777, line 24: Include changes have not all been carried forward.

aeron-client/src/main/cpp_wrapper/concurrent/CountersReader.h
```diff
@@ -0,0 +1,251 @@
...
+/**
+ * Reads the counters metadata and values buffers.
+ *
+ * This class is threadsafe.
+ *
+ * <b>Values Buffer</b>
+ * <pre>
+ *   0                   1                   2                   3
+ *   0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
+ *  +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
```
mjpt777, line 38: Changes for owner id are not carried across.

mjpt777:
On the whole it looks like a good approach and quite clean. My main concern is a number of the changes since the code split have not been carried across.

PR #994, opened 2020-07-25
Remove usage AERON_CPU_X64 as much as possible.

lygstate:
There is no need guard these code with AERON_CPU_X64

### DIFF
--- a/aeron-client/src/main/c/aeron_context.c
+++ b/aeron-client/src/main/c/aeron_context.c
@@ -20,7 +20,7 @@
 #include <stdlib.h>
 
 #include "util/aeron_platform.h"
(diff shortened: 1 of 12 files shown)


aeron-client/src/main/c/util/aeron_platform.h
```diff
@@ -26,9 +26,8 @@
 
     #if defined(_M_X64)
         #define AERON_CPU_X64 1
-
     #else
-        #error Unsupported CPU!
+        #define AERON_CPU_X86 1
```
tmontgomery, line 30: This is not a supported build option for us and should be removed.

lygstate: resoved.

PR #995, opened 2020-07-25
[C] Convert AERON_GET_AND_ADD_INT64 and AERON_GET_AND_ADD_INT32 from macro to inline function

lygstate:
With this patch, the usage also are simplified.


tmontgomery:
We want to keep the current arrangement of macros. Thanks, but no.

PR #998, opened 2020-07-25
Enable archive win32

lygstate:
Depends on #997

### DIFF
--- a/aeron-archive/src/main/cpp/CMakeLists.txt
+++ b/aeron-archive/src/main/cpp/CMakeLists.txt
@@ -111,6 +111,10 @@ target_link_libraries(aeron_archive_client
     PRIVATE aeron_client
     INTERFACE ${CMAKE_THREAD_LIBS_INIT})
 
+if (MSVC)
+    target_compile_options(aeron_archive_client PRIVATE -std:c++17 -Zc:__cplusplus)
+endif()
+
 if (AERON_INSTALL_TARGETS)
     install(TARGETS aeron_archive_client ARCHIVE DESTINATION lib)
     install(DIRECTORY . DESTINATION include FILES_MATCHING PATTERN "*.h")
--- a/aeron-client/src/main/cpp/Aeron.h
+++ b/aeron-client/src/main/cpp/Aeron.h
@@ -413,6 +413,8 @@ public:
      */
     static std::string version();
 
+    static MemoryMappedFile::ptr_t mapCncFile(const std::string cncFileName, long mediaDriverTimeoutMs);
+
 private:
     std::random_device m_randomDevice;
     std::default_random_engine m_randomEngine;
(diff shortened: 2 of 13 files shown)


aeron-archive/src/main/cpp/CMakeLists.txt
```diff
@@ -111,6 +111,10 @@ target_link_libraries(aeron_archive_client
     PRIVATE aeron_client
     INTERFACE ${CMAKE_THREAD_LIBS_INIT})
 
+if (MSVC)
+    target_compile_options(aeron_archive_client PRIVATE -std:c++17 -Zc:__cplusplus)
```
tmontgomery, line 115: specifying C++17 is too far. Should not be there. And definitely should not be different than top level CMakeLists.

aeron-client/src/main/cpp/Aeron.h
```diff
@@ -413,6 +413,8 @@ class CLIENT_EXPORT Aeron
      */
     static std::string version();
 
+    static MemoryMappedFile::ptr_t mapCncFile(const std::string cncFileName, long mediaDriverTimeoutMs);
```
tmontgomery, line 416: This changes the API. Do not change the API without discussion.

tmontgomery:
This combines some things that are good with several that are not. If you can separate the changes to the test from changes to the API or build, we will review again.

aeron-archive/src/main/cpp/CMakeLists.txt
lygstate: SBE depends on this, otherwise compiling error arise

aeron-client/src/main/cpp/Aeron.h
lygstate: mapCncFile need to be accesed in Context::requestDriverTermination, maybe move it into Context is a good idea..

tmontgomery: The current version works just fine. There does not need to be a new API method for it.

aeron-archive/src/main/cpp/CMakeLists.txt
tmontgomery: C++17 is not required for SBE. It can use it. But it is not required.

lygstate: This is the compiling error, any suggetion
```
[build] ..\..\aeron-archive\src\main\cpp\client\ControlResponseAdapter.cpp(68): error C2051: case 表达式不是常量
[build] ..\..\aeron-archive\src\main\cpp\client\ControlResponseAdapter.cpp(85): error C2051: case 表达式不是常量
[build] ..\..\aeron-archive\src\main\cpp\client\ControlResponseAdapter.cpp(115): warning C4065: switch 语句包含“default”但是未包含“case”标签
```

PR #1002, opened 2020-07-25
[CPP] Fixes compiling error for aeron-archive on Windows

lygstate:


mjpt777:
This PR is not about compilation errors and more about how to launch the driver for tests. Please separate and provide a better name.

The use of posix is not a good approach. An alternative like `CreateProcess` would be better on Windows.

PR #1013, opened 2020-07-31
[C++] Fix add/remove counter availability and close handlers to work …

mikeb01:
…with lvalue lambdas and not segfault.


mjpt777:
For the "token" I would use a signed positive 64-bit correlation id. This way it is consistent with other resources.

mikeb01:
Changed this around a bit.  The add function returns the int64_t registrationId and can be used as a token to remove it again.  I don't use the token internally, so removing by reference to the std::function will only work for static methods.  However I've caught the nullptr case so it won't segfault any more.  The registrationId is managed through an atomic increment on the conductor.

PR #1016, opened 2020-08-02
[C] Remove the usage of HAVE_STRUCT_MMSGHDR, using aeron_mmsghdr instead of mmsghdr for more platform independent socket.

lygstate:


tmontgomery:
We use the check for other things. It's not good to remove it.

PR #1019, opened 2020-08-06
[C] Add initial term id to message header in C API.

mikeb01:
Needed in order to access the initial_term_id in the C++ Wrapper.  Useful for feature parity with the Java driver also.

### DIFF
--- a/aeron-client/src/main/c/aeronc.h
+++ b/aeron-client/src/main/c/aeronc.h
@@ -36,21 +36,23 @@ typedef struct aeron_exclusive_publication_stct aeron_exclusive_publication_t;
 typedef struct aeron_header_stct aeron_header_t;
 #pragma pack(push)
 #pragma pack(4)
+typedef struct aeron_header_values_frame_stct
+{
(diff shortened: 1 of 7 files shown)


aeron-client/src/main/c/aeronc.h
```diff
@@ -38,15 +38,20 @@ typedef struct aeron_header_stct aeron_header_t;
...
-    int32_t frame_length;
-    int8_t version;
-    uint8_t flags;
-    int16_t type;
-    int32_t term_offset;
-    int32_t session_id;
-    int32_t stream_id;
-    int32_t term_id;
-    int64_t reserved_value;
+    struct aeron_header_values_frame_stct
```
lygstate, line 41: consider typedef this out?

mikeb01: We typically don't add the typedefs if it is nested.  @tmontgomery Thoughts?

tmontgomery: I think this is fine.

PR #1020, opened 2020-08-06
[C] Surface aeron_default_path function for C API.

mikeb01:


lygstate:
Both driver and client use this API, how about move to a shared c source file?

tmontgomery:
Looks good.

PR #1021, opened 2020-08-06
[C] Remove the AERON_IOVEC that not used at all.

lygstate:


tmontgomery:
This is used, just not in the project.

PR #1030, opened 2020-08-07
[C] Change log_file_name to fixed size string

lygstate:


tmontgomery:
It would be better to adjust the size of the allocate to add the string at the end of the struct and simply point the pointer at it. Thus 1 alloc and no wasted space.

lygstate:
I didn't got your idea, you can pick and modify my patch to show you idea

mjpt777:
Closing due to lack of progress.

PR #1036, opened 2020-08-14
[C] Collapse destination add/remove poll to a single function per resource to match the shape of the C++ API.

mikeb01:
* Add/correct functions declared in aeronc.h
* Uses a single poll function per resource for checking for the completion of destination commands (add/remove).  This matches the shape of the C++ API which has a single findDestinationResponse function shared for both adds and removes.
* Removes the older add/remove destination functions that did not use a async tracking for completion.


tmontgomery:
Looks good to me.

PR #1038, opened 2020-08-15
[C] Remove the usage of duplicated setting of errno by errno = ERROR_…

lygstate:
…NO_MACRO

### DIFF
--- a/aeron-client/src/main/c/aeron_client.c
+++ b/aeron-client/src/main/c/aeron_client.c
@@ -51,7 +51,6 @@ int aeron_init(aeron_t **client, aeron_context_t *context)
 
     if (NULL == client || NULL == context)
     {
-        errno = EINVAL;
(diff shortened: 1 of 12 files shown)


tmontgomery:
We want to set the errno in these cases as we do not know what the app is doing.

lygstate:
@tmontgomery You didn't got the idea, errno  are already setted in aeron_set_err

tmontgomery:
I got it. But this is future proofing. If we later decide not to set errno in `aeron_set_err`, then it will break. So, no we want to set it explicitly at the API so that it won't break later.

lygstate:
@tmontgomery So why not make a consistence decision.

tmontgomery:
`aeron_set_err` setting errno is not valid in some cases, BTW, as the err code is not the same range as errno. So, it should be conditionally set or not set at all.

PR #1039, opened 2020-08-15
[C] aeron_set_err use EINVAL instead -1.

lygstate:


tmontgomery:
@mikeb01 is this correct with what you had in mind?

mikeb01:
I think EINVAL is relevant for some of these cases, but not all.

mikeb01:
Closing this in favour of #1053 and #1054.

PR #1040, opened 2020-08-15
[C] bit_set check NULL pointer

lygstate:


mikeb01:
I think the fix is good, but would like to see a couple of unit test cases to accompany the change.

lygstate:
> I think the fix is good, but would like to see a couple of unit test cases to accompany the change.

done

PR #1042, opened 2020-08-15
[C] include intrin.h in bitutil instead of windows.h  but still prese…

lygstate:
…rve standalone property.

### DIFF
--- a/aeron-client/src/main/c/util/aeron_bitutil.h
+++ b/aeron-client/src/main/c/util/aeron_bitutil.h
@@ -23,8 +23,7 @@
 #include "util/aeron_platform.h"
 
 #if defined(AERON_COMPILER_MSVC)
-#include <winsock2.h>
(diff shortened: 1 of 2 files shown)


aeron-client/src/main/c/util/aeron_bitutil.h
```diff
@@ -89,7 +88,7 @@ inline int aeron_number_of_trailing_zeroes_u64(uint64_t value)
     }
 
     return __builtin_ctzll(value);
-#elif defined(_MSC_VER)
+#elif defined(_MSC_VER) && defined(AERON_CPU_X64)
```
mjpt777, line 91: Why add the x86 check back?

lygstate: _BitScanForward64 is only available on X64 target

PR #1044, opened 2020-08-17
[C] Convert AERON_GET_AND_ADD_INT64 and AERON_GET_AND_ADD_INT32 from …

lygstate:
…macro to inline function


mjpt777:
At present we are not going to take such an approach.

PR #1048, opened 2020-08-27
[C] Move some function prototypes to aeronc.h for counters reader.  A…

mikeb01:
…dd define for max label length (with static assert for correctness).


aeron-client/src/main/c/aeronc.h
```diff
@@ -596,6 +674,62 @@ void aeron_counters_reader_foreach_counter(
...
+
+/**
+ * Get the free for reuse deadline (ms) for a counter
+ *
+ * @param counters_reader that contains the counter
+ * @param counter_id to find
+ * @param deadline_ms output value to store the deadline
+ * @return -1 on failure, 0 on success.
+ */
+int aeron_counters_reader_free_for_reuse_deadline_ms(
```
tmontgomery, line 730: Is this necessary to expose?

PR #1051, opened 2020-09-04
fix cluster example start script

QIvan:
without this change, the example doesn't start with an error like 
```
Error: Could not find or load main class io.aeron.samples.tutorial.cluster.BasicAuctionClusteredServiceNode
Caused by: java.lang.ClassNotFoundException: io.aeron.samples.tutorial.cluster.BasicAuctionClusteredServiceNode
Error: Could not find or load main class io.aeron.samples.tutorial.cluster.BasicAuctionClusteredServiceNode
Caused by: java.lang.ClassNotFoundException: io.aeron.samples.tutorial.cluster.BasicAuctionClusteredServiceNode
Error: Could not find or load main class io.aeron.samples.tutorial.cluster.BasicAuctionClusteredServiceNode
Caused by: java.lang.ClassNotFoundException: io.aeron.samples.tutorial.cluster.BasicAuctionClusteredServiceNode
```

should the parameters for the example contain the word "tutorial", like in `aeron.cluster.tutorial.bidIntervalMs`?

thanks!

### DIFF
--- a/aeron-samples/scripts/cluster/basic-auction-cluster
+++ b/aeron-samples/scripts/cluster/basic-auction-cluster
@@ -34,7 +34,7 @@ function startNode() {
         -Daeron.event.cluster.log=all \
         -Daeron.cluster.tutorial.nodeId=$1 \
         ${JVM_OPTS} ${ADD_OPENS} \
-        io.aeron.samples.tutorial.cluster.BasicAuctionClusteredServiceNode > logs/cluster-$1.log &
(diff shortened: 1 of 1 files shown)


mjpt777:
I agree that including the scope of `tutorial` does help.

QIvan:
> I agree that including the scope of tutorial does help.
ok then nothing more should be changed. 
Changes for this PR is necessary because you moved package `cluster` (with classes like `BasicAuctionClusteredServiceNode`) from `tutorial` to `samples`

mikeb01:
I think moving them into a tutorial subpackage is better.  Added PR #1055.

PR #1053, opened 2020-09-05
[C] Add aeron specific errors codes.  Change error codes in conductor…

mikeb01:
… so that they can be distinguished in the wrapper.  Use EXFULL for situations where the client queue or driver ring buffer is full.

The use of `EXFULL` as an error code is a debatable one.  It was the closest standard error code that represented the case of a full buffer or queue (wanted avoid overloading `ETIMEDOUT`).  We need something fairly distinct in order to map it to an IllegalStateException in the wrapper.  If that naming is not particularly useful, then we should add a new Aeron specific code.


mikeb01:
Okay, well that question was answered.  `EXFULL` not defined on Mac and Windows, custom error code it is then.

PR #1060, opened 2020-09-11
Misc fixes for cpp_wrapper merge

lygstate:
Still have a api differencies

```
TEST_F(SystemTest, shouldNotSegfaultWhenRemovedByReference)
{
    on_available_counter_t dynamicAvailableHandler =
        [&](CountersReader &countersReader, std::int64_t registrationId, std::int32_t counterId)
        {
        };

    on_available_counter_t dynamicUnavailableHandler =
        [&](CountersReader &countersReader, std::int64_t registrationId, std::int32_t counterId)
        {
        };

    Context ctx;
    ctx.useConductorAgentInvoker(true);
    std::shared_ptr<Aeron> aeron = Aeron::connect(ctx);
    AgentInvoker<ClientConductor> &invoker = aeron->conductorAgentInvoker();
    invoker.start();

    aeron->addAvailableCounterHandler(dynamicAvailableHandler);
    aeron->addUnavailableCounterHandler(dynamicUnavailableHandler);
    invoker.invoke();

#ifndef AERON_C_CLIENT_WRAPPER_TEST
    aeron->removeAvailableCounterHandler(dynamicAvailableHandler);
    aeron->removeUnavailableCounterHandler(dynamicUnavailableHandler);
#endif
}

```


mikeb01:
> Still have a api differencies

Those have been deliberately left off the wrapper.  We've deprecated this approach in the C++ API and the registration id should be used to remove the handler.

lygstate:
> > Still have a api differencies
> 
> Those have been deliberately left off the wrapper. We've deprecated this approach in the C++ API and the registration id should be used to remove the handler.

gotcha so we can convert the test use  registration id, and how about the TODO in system test?

currently it's not failling

mikeb01:
> gotcha so we can convert the test use registration id, 

No, there was a bug in the existing C++ API implementation that caused a seqfault when that function was used.  This test was to ensure that bug was fixed.

> and how about the TODO in system test?  currently it's not failling

It fails when running the sanitizer.

lygstate:
> > gotcha so we can convert the test use registration id,
> 
> No, there was a bug in the existing C++ API implementation that caused a seqfault when that function was used. This test was to ensure that bug was fixed.
> 
> > and how about the TODO in system test?  currently it's not failling
> 
> It fails when running the sanitizer.

OK, update with fixes, guard with macro

mikeb01:
We've merged a few of these changes manually.  I'm not sure about the collapsing of the C++ and C++ wrapper tests.  I'd like to think about this a bit more.  We may eventually move over to the wrapper and remove the C++ API, so I'd like to be able to remove it without impacting the wrapper.

PR #1062, opened 2020-09-11
[C/Java] Add Archive/Cluster counter typeIds to the constants list.

vyazelenko:


aeron-cluster/src/main/java/io/aeron/cluster/ClusterBackup.java
```diff
@@ -52,22 +52,22 @@
...
      * The type id of the {@link Counter} used for the backup state.
      */
-    public static final int BACKUP_STATE_TYPE_ID = 208;
+    public static final int BACKUP_STATE_TYPE_ID = AeronCounters.CLUSTER_BACKUP_STATE_TYPE_ID;
 
     /**
      * The type id of the {@link Counter} used for the live log position counter.
      */
-    public static final int LIVE_LOG_POSITION_TYPE_ID = 209;
+    public static final int LIVE_LOG_POSITION_TYPE_ID = AeronCounters.CLUSTER_LIVE_LOG_POSITION_TYPE_ID;
```
mjpt777, line 60: It is worth adding that this and the next one are cluster backup and not just cluster.

PR #1069, opened 2020-10-01
[Java/C++] Archive variable length recordings.

vyazelenko:
This PR changes Archive recordings from being fixed length (1KB) to variable-length.

Summary:
- Recording descriptors are variable-length.
- No limitation on the length of the `strippedChannel`, `originalChannel` and `sourceIdentity` Strings.
- Ability to purge recordings (`io.aeron.archive.client.AeronArchive#purgeRecording`), i.e. mark recording as `INVALID` and delete the corresponding segment files.
- `max-entries` commands in ArchiveTool have been deprecated in favor of `capacity`, i.e. Catalog no longer has a max entries limit but has a capacity expressed in bytes. Catalog is auto-growing it's capacity when needed but an explicit command can be used as well.
- New command `ArchiveTool#compact` to compact the Catalog file, i.e. remove all recordings in state `INVALID` and delete the corresponding segment files.
- Add checksum to the `RecordingDescriptorHeader`. Now if the `io.aeron.archive.Archive.Context#recordChecksum()` is set the checksum of the entire `RecordingDescriptor` will be taken and stored in the header.
- `ArchiveTool#verify` - validate `RecordingDescriptor`  checksums.
- `ArchiveTool#checksum` - compute and persist `RecordingDescriptor`  checksums.
- Add support for jumbo `RecordingDescriptors`, i.e. allow `RecordingDescriptors` bigger than the `tryClaim` size by using `Publication#offer` API.
- Migration task to migrate from version `2` to version `3` of the Archive.

### DIFF
--- a/aeron-archive/src/test/java/io/aeron/archive/ArchiveMigrationUtils.java
+++ b/aeron-archive/src/test/java/io/aeron/archive/ArchiveMigrationUtils.java
@@ -16,6 +16,7 @@
 package io.aeron.archive;
 
 import io.aeron.archive.client.ArchiveException;
+import org.agrona.IoUtil;
 import org.agrona.LangUtil;
 import org.agrona.SemanticVersion;
 import org.agrona.concurrent.EpochClock;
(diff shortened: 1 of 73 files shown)


aeron-archive/src/test/java/io/aeron/archive/ArchiveMigrationUtils.java
```diff
@@ -0,0 +1,319 @@
...
+                buffer.putInt(index, VERSION_2_1_0); // version
+                index += SIZE_OF_INT;
+
+                buffer.putLong(index, currentTime, LITTLE_ENDIAN); // activityTimestamp
+                index += SIZE_OF_LONG;
+
+                buffer.putLong(index, currentTime, LITTLE_ENDIAN); // startTimestamp
+                index += SIZE_OF_LONG;
+
+                buffer.putLong(index, 1, LITTLE_ENDIAN); // pid
```
mjpt777, line 83: The use of magic numbers here is not good.

vyazelenko: Agree. Those can be pulled out either as constants or better yet as parameters to this method. This class is used only in tests.

mjpt777: The comments are a smell.

vyazelenko: I'll refactor it.

PR #1078, opened 2020-10-22
[C] Add common string equals functions.

vyazelenko:


vyazelenko:
@mjpt777 Done.

aeron-client/src/main/c/util/aeron_strutil.h
```diff
@@ -58,6 +61,64 @@ inline uint64_t aeron_fnv_64a_buf(uint8_t *buf, size_t len)
...
+ *
+ * @param str1 first string.
+ * @param str2 second string.
+ * @return true if strings are equal (or if first 1024 characters are equal in case of very long strings).
+ */
+inline bool aeron_str_equals(const char *str1, const char *str2)
+{
+    static const size_t max_length = 1024;
+    assert (strlen(str1) <= max_length);
+    assert (strlen(str2) <= max_length);
```
jvirtanen, line 117: Is the "Note:" above in synchronization with these assertions?

aeron-client/src/main/c/util/aeron_strutil.h
```diff
@@ -58,6 +61,64 @@ inline uint64_t aeron_fnv_64a_buf(uint8_t *buf, size_t len)
...
+    for (size_t i = 0; i < length; i++)
+    {
+        if (str1[i] != str2[i])
+        {
+            return false;
+        }
+        else if (str1[i] == '\0')
+        {
+           return true;
+        }
```
jvirtanen, line 93: Is the `@return` documentation correct for `aeron_strn_equals("foo", "foobar", 6)`?

aeron-client/src/main/c/util/aeron_strutil.h
vyazelenko: yes

aeron-client/src/main/c/util/aeron_strutil.h
vyazelenko: Updated the doc.

aeron-client/src/main/c/util/aeron_strutil.h
jvirtanen: Wouldn't the `assert()` call `abort()`, however, when either string is longer than 1024, unless `NDEBUG` was defined during the build? As I read it, the "Note:" says that the behavior would be equal to calling `aeron_strn_equals` with `length` set to 1024 in that case, however, `aeron_strn_equals` would not call `abort()` regardless of the `NDEBUG` macro.

vyazelenko: It is now a dedicated assert that is checked always.

PR #1085, opened 2020-11-04
[Java] unstable test for Archive connection

QIvan:
this test is very unstable even on my local Linux box which shouldn't be starved or having any resource issue... 

I didn't disable it because I wanted to see how it works on your build servers.

thanks.

### DIFF
--- a/aeron-system-tests/src/test/java/io/aeron/archive/ArchiveTest.java
+++ b/aeron-system-tests/src/test/java/io/aeron/archive/ArchiveTest.java
@@ -333,6 +333,28 @@ public class ArchiveTest
             messageCount);
     }
 
+    @ParameterizedTest
+    @MethodSource("threadingModes")
+    @Timeout(30)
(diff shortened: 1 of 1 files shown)


QIvan:
usually, it fails with 
```
io.aeron.exceptions.DriverTimeoutException: FATAL - no response from MediaDriver within (ns): 10000000000
```
or 

```
ERROR - Archive connect timeout: correlationId=13 step=4
io.aeron.exceptions.TimeoutException: ERROR - Archive connect timeout: correlationId=13 step=4
```

QIvan:
[According to Martin](https://gitter.im/real-logic/Aeron?at=5fa154f77cac87158f83b21d) there is nothing wrong with this code from Aeron point of view.

mikeb01:
I'm going to close this now.  Notes are in gitter chat.  This is caused by re-using an Aeron client that is 'owned' by another archive.

PR #1089, opened 2020-11-15
[C] Remove #ifdefs for epoll/poll/wsapoll from poller header files an…

mikeb01:
…d use binding_clientd to track poller specific information.

This encapsulates the poll/epoll/wsapoll condition compilation within the .c file.  It also means that the `aeron_udp_transport_poller` struct has a fixed size, regardless of the compilation options, which tripped me up in tests at one point.


mjpt777 (commented):
LGTM

PR #1090, opened 2020-11-20
Leader heartbeat timeout for custom NameResolver

segabriel:
Hi all
I tried to update the Aeron version from `1.29.0` to the latest and faced with an issue kinda `io.aeron.cluster.client.ClusterException: leader heartbeat timeout` and it reproduced on Linux machines only (Windows and MacOS work properly).
I work with a custom implementation of the `NameResolver ` and during starting cluster nodes they can not always know the right IPs other members and if my `NameResolver` was not updated yet it returns `localhost:7` to avoid immediate failure by `java.net.UnknownHostException` (have no ability return `null`). And such workaround worked till `1.30.0`.
Then I found the commit that broke the previous behavior -> https://github.com/real-logic/aeron/commit/58e7dbb2f5b07ebafabf24252734eff44075a3a6 and after reverting of `timeOfLastResolutionNs` updating in  `SendChannelEndpoint#onStatusMessage` it works again. 
Also, I added my test scenario with a shared `NameResolver` implementation and it's populated each time when a cluster node is starting `TestClusterWithNameResolver:174` to simulate dynamic populating addresses by names.

I am still not sure that is a bug and why it is reproduced on Linux machines only.

### DIFF
--- a/aeron-cluster/src/test/java/io/aeron/cluster/CustomNameResolverTest.java
+++ b/aeron-cluster/src/test/java/io/aeron/cluster/CustomNameResolverTest.java
@@ -1,0 +1,47 @@
+package io.aeron.cluster;
+
+import org.agrona.CloseHelper;
+import org.agrona.LangUtil;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestInfo;
+
+public class CustomNameResolverTest
+{
+
+    private TestClusterWithNameResolver cluster;
+
+    @AfterEach
+    void tearDown()
+    {
+        CloseHelper.close(cluster);
+    }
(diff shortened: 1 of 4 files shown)


mjpt777:
Does this make a difference? https://github.com/real-logic/aeron/commit/efaee163944ec925260ef280c2a66d9c7f26ae55

segabriel:
> Does this make a difference? [efaee16](https://github.com/real-logic/aeron/commit/efaee163944ec925260ef280c2a66d9c7f26ae55)

I just checked yes it works. Thanks!

One more question, did you think to allow `io.aeron.driver.NameResolver#lookup` return `null`?

segabriel:
Yes, that's it, look forward to the new release, thank you!

PR #1099, opened 2020-12-04
Error Stack Experiment

mikeb01:
More for discussion that immediate merge.  This introduces a different approach for handling errors and capturing context as we traverse the stack.  The idea being that if an error occurs, we set the error similar to normal, with an error code and a formatted message.  As we return back up through the stack, when it looks like a call further down has caused the error, then we can append additional context to the current error stack.  Adds 2 macros:
- `AERON_SET_ERROR(error_code, format, ...)`
- `AERON_APPEND_ERROR(format, ...)`

`AERON_SET_ERROR` should be used at the edge of the Aeron driver (or client), when we have tried to do something and it has failed.  E.g. system call or some internal validation.  At this point we set the error code and and capture any immediate context relevant data.  This call will also reset the error stack discarding any previously captured messages.

`AERON_APPEND_ERROR` is used when another Aeron function has returned `-1` in which case we just append some local context to the error stack and return `-1` again.  This allow for additional context information to be included in the error message without having to pass that information up or down the stack and break encapsulation for the purposes of error reporting.

I've demonstrated how this would work for initialising the UDP channel transport.  On failure the error message reporting back to the client ends up being:
```
io.aeron.exceptions.RegistrationException: ERROR - (98) Address already in use
[aeron_udp_channel_transport_init, aeron_udp_channel_transport.c:85] unicast bind(127.0.0.1:24325)
[aeron_receive_destination_create, aeron_receive_destination.c:55] uri = aeron:udp?endpoint=localhost:24325
[aeron_driver_conductor_get_or_add_receive_channel_endpoint, aeron_driver_conductor.c:1644] correlation_id = 3
```
Similarly the distinct error log ends up containing:
```
1 observations from 2020-12-04 15:31:41.855+1300 to 2020-12-04 15:31:41.855+1300 for:
 0: (98) Address already in use
[aeron_udp_channel_transport_init, aeron_udp_channel_transport.c:85] unicast bind(127.0.0.1:24325)
[aeron_receive_destination_create, aeron_receive_destination.c:55] uri = aeron:udp?endpoint=localhost:24325
[aeron_driver_conductor_get_or_add_receive_channel_endpoint, aeron_driver_conductor.c:1644] correlation_id = 2
```
I've limited the depth of the stack to 16.  If it overflows, then we track the lowest 15 elements of the error stack and the highest one.  When dumping out the stack we would potentially see something similar to:
```
[TestBody, aeron_error_test.cpp:70] this is the root error
[TestBody, aeron_error_test.cpp:73] this is a nested error: 0
[TestBody, aeron_error_test.cpp:73] this is a nested error: 1
[TestBody, aeron_error_test.cpp:73] this is a nested error: 2
[TestBody, aeron_error_test.cpp:73] this is a nested error: 3
[TestBody, aeron_error_test.cpp:73] this is a nested error: 4
[TestBody, aeron_error_test.cpp:73] this is a nested error: 5
[TestBody, aeron_error_test.cpp:73] this is a nested error: 6
[TestBody, aeron_error_test.cpp:73] this is a nested error: 7
[TestBody, aeron_error_test.cpp:73] this is a nested error: 8
[TestBody, aeron_error_test.cpp:73] this is a nested error: 9
[TestBody, aeron_error_test.cpp:73] this is a nested error: 10
[TestBody, aeron_error_test.cpp:73] this is a nested error: 11
[TestBody, aeron_error_test.cpp:73] this is a nested error: 12
[TestBody, aeron_error_test.cpp:73] this is a nested error: 13
(11 lines omitted...)
[TestBody, aeron_error_test.cpp:73] this is a nested error: 25
```
So the error message would inform any user that some of the stack elements had to be omitted.  I think 16 should be plenty.  Our call stacks a pretty shallow.


mjpt777:
I really like this. So useful to build up context.

mikeb01:
I think this is good to go now.  I've ended up doing a fair overhaul of the error reporting in the C driver and client.  I've simplified the way that the distinct error log reporter works in the C driver and it is more like the Java implementation.  I.e. it matches against the shape of the stack trace (with the C driver it is just a string comparison).  It also decouples cleanly the formatting of error messages from the recording of them.  As a result I've updated a lot of the top of stack reporting in the conductor, sender and receiver.

I've also simplified the implementation of the error reporting itself.  It no longer maintains a stack, just a single buffer.  This reduces some of the headaches when the output of a error stack from the server gets embedded in an error line for the client.  Now there is less chance of truncation.

aeron-client/src/main/c/aeron_alloc.c
```diff
@@ -52,10 +53,7 @@ int aeron_alloc(void **ptr, size_t size)
 
     if (NULL == *ptr)
     {
-        errno = ENOMEM;
-#if defined(AERON_COMPILER_MSVC)
-        SetLastError(ERROR_OUTOFMEMORY);
-#endif
+        AERON_SET_ERR(ENOMEM, "Failed to allocate %llu bytes", (unsigned long long)size);
```
mjpt777, line 56: Would it be better to cast to `uint64_t` and use `PRI` macro?

aeron-client/src/main/c/aeron_exclusive_publication.c
```diff
@@ -587,8 +592,12 @@ int aeron_exclusive_publication_local_sockaddrs(
...
     if (NULL == publication || NULL == address_vec || address_vec_len < 1)
     {
-        errno = EINVAL;
-        aeron_set_err(EINVAL, "%s", strerror(EINVAL));
+        AERON_SET_ERR(
+            EINVAL,
+            "Parameters must correct, publication: %s, address_vec: %s, address_vec_len: %llu < 1",
+            AERON_NULL_STR(publication),
+            AERON_NULL_STR(address_vec),
+            (unsigned long long)address_vec_len);
```
mjpt777, line 600: Better to use fix length types.

aeron-client/src/main/c/aeron_publication.c
```diff
@@ -498,10 +512,19 @@ int32_t aeron_publication_session_id(aeron_publication_t *publication)
...
+            AERON_NULL_STR(publication),
+            AERON_NULL_STR(address_vec));
+        return -1;
+    }
+    if (address_vec_len < 1)
     {
-        errno = EINVAL;
-        aeron_set_err(EINVAL, "%s", strerror(EINVAL));
+        AERON_SET_ERR(
+            EINVAL, "Parameters must be valid, address_vec_len (%llu) < 1", (unsigned long long) address_vec_len);
```
mjpt777, line 527: Better to use fixed length types.

aeron-client/src/main/c/aeron_subscription.c
```diff
@@ -128,9 +126,7 @@ int aeron_subscription_alloc_image_list(volatile aeron_image_list_t **image_list
     *image_list = NULL;
     if (aeron_alloc((void **)&_image_list, AERON_IMAGE_LIST_ALLOC_SIZE(length)) < 0)
     {
-        int errcode = errno;
-
-        aeron_set_err(errcode, "aeron_subscription_create (%d): %s", errcode, strerror(errcode));
+        AERON_APPEND_ERR("Unable to allocate image list, length: %lld", (unsigned long long)length);
```
mjpt777, line 129: Better to use fixed length types.

aeron-client/src/main/c/aeron_subscription.c
```diff
@@ -532,12 +540,20 @@ int64_t aeron_header_position(aeron_header_t *header)
...
+            EINVAL,
+            "Parameters must not be null, subscription: %s, address_vec: %s",
+            AERON_NULL_STR(subscription),
+            AERON_NULL_STR(address_vec));
         return -1;
     }
+    if (address_vec_len < 1)
+    {
+        AERON_SET_ERR(
+            EINVAL, "Parameters must be valid, address_vec_len (%llu) < 1", (unsigned long long) address_vec_len);
```
mjpt777, line 555: Better to use fixed length types.

aeron-client/src/main/c/aeron_subscription.c
```diff
@@ -549,12 +565,20 @@ int aeron_subscription_local_sockaddrs(
...
+            EINVAL,
+            "Parameters must not be null, subscription: %s, address: %s",
+            AERON_NULL_STR(subscription),
+            AERON_NULL_STR(address));
         return -1;
     }
+    if (address_len < 1)
+    {
+        AERON_SET_ERR(
+            EINVAL, "Parameters must be valid, address_len (%llu) < 1", (unsigned long long) address_len);
```
mjpt777, line 580: Better to use fixed length types.

aeron-client/src/main/c/aeron_subscription.c
```diff
@@ -603,12 +627,20 @@ static int aeron_subscription_update_uri_with_resolved_endpoint(
...
+            EINVAL,
+            "Parameters must not be null, subscription: %s, uri: %s",
+            AERON_NULL_STR(subscription),
+            AERON_NULL_STR(uri));
         return -1;
     }
+    if (uri_len < 1)
+    {
+        AERON_SET_ERR(
+            EINVAL, "Parameters must be valid, uri_len (%llu) < 1", (unsigned long long) uri_len);
```
mjpt777, line 642: Better to use fixed length types.

aeron-client/src/main/c/concurrent/aeron_broadcast_receiver.c
```diff
@@ -43,7 +43,7 @@ int aeron_broadcast_receiver_init(aeron_broadcast_receiver_t *receiver, void *bu
     }
     else
     {
-        aeron_set_err(EINVAL, "%s:%d: %s", __FILE__, __LINE__, strerror(EINVAL));
+        AERON_SET_ERR(EINVAL, "Capacity: %llu invalid, must be power of two", (unsigned long long) capacity);
```
mjpt777, line 46: Better to use fixed length types.

aeron-client/src/main/c/concurrent/aeron_broadcast_receiver.c
```diff
@@ -65,7 +65,11 @@ int aeron_broadcast_receiver_receive(
     {
         if (last_seen_lapped_count != receiver->lapped_count)
         {
-            aeron_set_err(EINVAL, "unable to keep up with broadcast");
+            AERON_SET_ERR(
+                EINVAL,
+                "unable to keep up with broadcast, last_seen_lapped_count: %ld, receiver->lapped_count: %ld",
```
mjpt777, line 70: Better to use fixed length types.

aeron-client/src/main/c/concurrent/aeron_broadcast_receiver.c
```diff
@@ -76,7 +80,11 @@ int aeron_broadcast_receiver_receive(
 
         if (length > sizeof(receiver->scratch_buffer))
         {
-            aeron_set_err(EINVAL, "scratch buffer too small");
+            AERON_SET_ERR(
+                EINVAL,
+                "scratch buffer too small, required: %llu, found: %llu",
+                (unsigned long long) length,
+                (unsigned long long) sizeof(receiver->scratch_buffer));
```
mjpt777, line 87: Better to use fixed length types.

aeron-client/src/main/c/concurrent/aeron_broadcast_transmitter.c
```diff
@@ -35,7 +36,7 @@ int aeron_broadcast_transmitter_init(aeron_broadcast_transmitter_t *transmitter,
     }
     else
     {
-        aeron_set_err(EINVAL, "%s:%d: %s", __FILE__, __LINE__, strerror(EINVAL));
+        AERON_SET_ERR(EINVAL, "Capacity: %llu invalid, must be power of two", (unsigned long long) capacity);
```
mjpt777, line 39: Better to use fixed length types.

aeron-client/src/main/c/concurrent/aeron_counters_manager.c
```diff
@@ -49,7 +50,11 @@ int aeron_counters_manager_init(
     }
     else
     {
-        aeron_set_err(EINVAL, "%s:%d: %s", __FILE__, __LINE__, strerror(EINVAL));
+        AERON_SET_ERR(
+            EINVAL,
+            "Counter buffer lengths invalid, metadata_length: %llu, values_length: %llu",
+            (unsigned long long)metadata_length,
```
mjpt777, line 56: Better to use fixed length types.

mjpt777 (changes requested):
LGTM

PR #1102, opened 2020-12-06
[C/C++] Have ErrorLogReader delegate to the C implementation.  Remove…

mikeb01:
… DistinctErrorLog.h and ErrorLogDecriptor.h from the wrapper (unused).  Add test.


mikeb01:
This code will end up being removed.

PR #1113, opened 2020-12-31
Topology tests & allow multicast for log channel

mikeb01:
There is still one outstanding test failure as result of this change.  In `shouldRecoverQuicklyAfterKillingFollowersThenRestartingOne` it appears that when extending the recording the image does not become available to the archive's subscription therefore it never constructs the recording counter and ConsensusModule's conductor times out.

### DIFF
--- a/aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModule.java
+++ b/aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModule.java
@@ -39,6 +39,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
 import java.util.function.Supplier;
 
+import static io.aeron.CommonContext.ENDPOINT_PARAM_NAME;
 import static io.aeron.cluster.ConsensusModule.Configuration.*;
 import static io.aeron.cluster.service.ClusteredServiceContainer.Configuration.SNAPSHOT_CHANNEL_PROP_NAME;
 import static io.aeron.cluster.service.ClusteredServiceContainer.Configuration.SNAPSHOT_STREAM_ID_PROP_NAME;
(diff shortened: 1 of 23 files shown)


aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModule.java
```diff
@@ -1306,6 +1308,9 @@ public void conclude()
                 egressPublisher = new EgressPublisher();
             }
 
+            final ChannelUri channelUri = ChannelUri.parse(logChannel());
+            logChannelIsMultiDestination = channelUri.isUdp() && null == channelUri.get(ENDPOINT_PARAM_NAME);
```
mjpt777, line 1312: Boolean naming convention starts with `is` or `has`, therefore `isLogChannelMultiDestination`.

aeron-test-support/src/main/java/io/aeron/test/cluster/TestCluster.java
```diff
@@ -60,7 +60,9 @@
 {
     private static final int SEGMENT_FILE_LENGTH = 16 * 1024 * 1024;
     private static final long CATALOG_CAPACITY = 128 * 1024;
-    private static final String LOG_CHANNEL = "aeron:udp?term-length=512k";
+//    private static final String LOG_CHANNEL = "aeron:udp?term-length=512k";
+    private static final String LOG_CHANNEL =
+        "aeron:udp?term-length=512k|endpoint=224.20.30.39:24326|interface=localhost";
```
mjpt777, line 65: Do we always want to test with multicast? Most use will be MDC.

mjpt777:
The general approach looks good.

aeron-test-support/src/main/java/io/aeron/test/cluster/TestCluster.java
mikeb01: I wanted to see some runs in CI with multicast enabled.  I've moved it back but left the configuration commented.  I'll pick up a later task to make this a parameter to our cluster tests.

PR #1117, opened 2021-01-12
[Java] Implementation to address #1115.  Allow heartbeats when periodic SETUP frames are requested.

mikeb01:
~~…ing status messages by controlAddress/stream/session.  Retain liveness for images when receiving setup messages, even if heartbeats aren't being received.~~

### DIFF
--- a/aeron-driver/src/main/java/io/aeron/driver/NetworkPublication.java
+++ b/aeron-driver/src/main/java/io/aeron/driver/NetworkPublication.java
@@ -78,7 +78,7 @@ class NetworkPublicationPadding2 extends NetworkPublicationConductorFields
 
 class NetworkPublicationSenderFields extends NetworkPublicationPadding2
 {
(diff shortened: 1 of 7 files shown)


mikeb01:
Changed the implementation to allow heartbeats more often to keep images alive.

aeron-driver/src/main/java/io/aeron/driver/NetworkPublication.java
```diff
@@ -78,7 +78,7 @@
 
 class NetworkPublicationSenderFields extends NetworkPublicationPadding2
 {
-    long timeOfLastSendOrHeartbeatNs;
+    long timeOfLastDataOrHeartbeatNs;
```
mjpt777, line 81: `timeOfLastDataFrameNs` is sufficient as a heartbeat is a data frame.

PR #1123, opened 2021-01-14
[C++] Archive client shared library addition

sshanks-kx:
Created additional shared lib version of aeron_archive_client.a (ie aeron_archive_client_shared.so).

This is similar to previous changes made to aeron_client.a were there was an addition to generate aeron_client_shared.so ( https://github.com/real-logic/aeron/commit/0e716c5cdee513e2f051606b7ee4ce73e89608fe#diff-751998d20f15036aea603a63c5dae00781bdc2258315bc2269a1698dd6fff51d )

### DIFF
--- a/aeron-archive/src/main/cpp/CMakeLists.txt
+++ b/aeron-archive/src/main/cpp/CMakeLists.txt
@@ -96,9 +96,11 @@ SET(HEADERS
 
 # static library
 add_library(aeron_archive_client STATIC ${SOURCE} ${HEADERS})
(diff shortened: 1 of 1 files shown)


mjpt777:
We have come to regret merging the previous support for creating a DLL. The implementation was partial and fills our build with so many warnings that important issues get missed. We are considering the removal of the DLL creation.

sshanks-kx:
We are currently using it to try archiving, as existing binary requires shared lib dlopen for integration with any 3rd party.

PR #1127, opened 2021-01-24
[C++] Use MMAP_POPULATE for faster prefaulting on Linux

StephanDollberg:
Prefaulting can be slow. MMAP_POPULATE speeds that up a bit as the
prefaulting happens directly in the kernel on creation and doesn't
require on demand interrupts and context switches.

I have left the windows version as is.

### DIFF
--- a/aeron-client/src/main/c/util/aeron_fileutil.c
+++ b/aeron-client/src/main/c/util/aeron_fileutil.c
@@ -19,6 +19,10 @@
 #define _GNU_SOURCE
 #endif
 
+#if defined(__linux__) || defined(AERON_COMPILER_MSVC)
+#define AERON_NATIVE_PRETOUCH
+#endif
+
 #include <sys/types.h>
 #include <sys/stat.h>
 #include <fcntl.h>
@@ -59,7 +63,7 @@
 #define S_IROTH 0
 #define S_IWOTH 0
 
-static int aeron_mmap(aeron_mapped_file_t *mapping, int fd, off_t offset, bool /* pre_touch */)
+static int aeron_mmap(aeron_mapped_file_t *mapping, int fd, off_t offset, bool pre_touch)
 {
     HANDLE hmap = CreateFileMapping((HANDLE)_get_osfhandle(fd), 0, PAGE_READWRITE, 0, 0, 0);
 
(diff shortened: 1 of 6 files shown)


StephanDollberg:
Not sure what's wrong with that macosx test, seems to complain about:

```
                Caused by:
                org.mockito.exceptions.base.MockitoInitializationException: 
                Could not initialize inline Byte Buddy mock maker.
```

mjpt777:
I'm warm to the concept. We need to be consistent across the C++ and C clients on this and consider the C media driver. I can have a look later in this week or next week. For BSD we could use `MAP_PREFAULT_READ` or `MADV_WILLNEED ` for more portability, and for Windows there is `PrefetchVirtualMemory`.  I'd prefer a more comprehensive solution so the benefits are not so limited.

StephanDollberg:
Ok cool sounds good. Happy to help with the C media driver part.

I don't have too much experience with either BSD or windows. Read about PrefetchVirtualMemory but somehow the MS docs 404ed for me.

StephanDollberg:
I checked `aeron_fileutil.c` and one could probably apply almost the same pattern there which from my understanding would cover both the C client and AMD? Will do if you think that makes sense.

I have checked using MADV_WILLNEED on Linux but that doesn't really work as nicely from what I can tell. From my understanding of https://www.spinics.net/lists/linux-mm/msg102603.html and grepping the Linux kernel log it seems that it's limited to a quite small number (128K or something). 

In regards to BSD: 

```
     MAP_PREFAULT_READ	Immediately update the calling process's lowest-level
			virtual	address	translation structures,	such as	its
			page table, so that every memory resident page within
			the region is mapped for read access.  Ordinarily
			these structures are updated lazily.  The effect of
			this option is to eliminate any	soft faults that would
			otherwise occur	on the initial read accesses to	the
			region.	 Although this option does not preclude	prot
			from including PROT_WRITE, it does not eliminate soft
			faults on the initial write accesses to	the region.
```

from https://www.freebsd.org/cgi/man.cgi?query=mmap&sektion=2&manpath=freebsd-release-ports. I understand it so that that doesn't work for writes? Not sure how the two are even separable but it sounds like it.

mjpt777:
Yes making the changes to `aeron_fileutil.c` would be the right place. If we can make it more efficient for Linux and Windows then the others can fail back to the touching the pages.,

StephanDollberg:
Have updated the C side. Trying to get my windows setup working now so that I can have a look at PrefetchVirtualMemory

StephanDollberg:
This should now be ready btw with support on Linux and Windows.

PR #1128, opened 2021-01-26
Multi bus topology test

mikeb01:
- Construct a virtual network that has 2 network buses.
- Add parameters to topology test that puts cluster traffic on a separate bus to ingress.
- Add CI build target for topology tests.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -118,6 +118,7 @@ jobs:
         run:  |
           ping -c 3 10.42.0.10
           ip route
+          sudo iptables -L
       - name: Run topology tests with Gradle
         run: ./gradlew topologyTest
         env:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -70,6 +70,71 @@ jobs:
           name: crash-logs-${{ matrix.os }}-java-${{ matrix.java }}
           path: ${{ steps.copy_test_logs.outputs.dir }}
 
+  java-topology-build:
+    name: Java Topology Tests (JDK ${{ matrix.java }}, ${{ matrix.os }})
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        java: [ '8' ]
+        os: [ 'ubuntu-20.04' ]
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Cache Gradle dependencies
+        uses: actions/cache@v2
+        with:
+          path: ~/.gradle/caches
+          key: ${{ runner.os }}-gradle-caches-${{ hashFiles('**/*.gradle') }}
+          restore-keys: |
+            ${{ runner.os }}-gradle-caches-
+      - name: Cache Gradle wrappers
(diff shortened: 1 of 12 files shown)


.github/workflows/ci.yml
```diff
@@ -70,6 +70,72 @@ jobs:
...
+        with:
+          path: ~/.gradle/wrapper
+          key: ${{ runner.os }}-gradle-wrapper-${{ hashFiles('**/gradle/wrapper/gradle-wrapper.properties') }}
+      - name: Setup java
+        uses: actions/setup-java@v1
+        with:
+          java-version: ${{ matrix.java }}
+      - name: Setup BUILD_JAVA_HOME (Linux)
+        if: runner.os == 'Linux'
+        run: echo "BUILD_JAVA_HOME=${JAVA_HOME}" >> $GITHUB_ENV
```
vyazelenko, line 102: Should be replaced with:
```
      - name: Setup BUILD_JAVA_HOME & BUILD_JAVA_VERSION
        run: |
         echo "BUILD_JAVA_HOME=${JAVA_HOME}" >> $GITHUB_ENV
         echo "BUILD_JAVA_VERSION=${{ matrix.java }}" >> $GITHUB_ENV
```

.github/workflows/ci.yml
```diff
@@ -70,6 +70,72 @@ jobs:
...
+      - name: Setup java 8 to run the Gradle script
+        uses: actions/setup-java@v1
+        with:
+          java-version: 8
```
vyazelenko, line 106: This step is only necessary if we intend to test on multiple JDKs. If we just use JDK 8 anyway we can remove this step.

.github/workflows/ci.yml
```diff
@@ -70,6 +70,72 @@ jobs:
...
+      - name: Setup virtual network interfaces
+        run: ./aeron-samples/scripts/cluster/setup-namespaces
+      - name: Build aeron with Gradle
+        run: ./gradlew -x javadoc -x test
```
vyazelenko, line 110: I would simply go for:
```
./gradlew clean assemble -x javadoc
```
This way it will be independent of the default commands in the `build.gradle` file.

.github/workflows/ci.yml
```diff
@@ -70,6 +70,72 @@ jobs:
...
+        run: ./agent-ns
+      - name: Get network information
+        run:  |
+          sudo sysctl -w net.ipv4.ip_forward=1
+          sudo iptables -F FORWARD
+          sudo iptables -P FORWARD ACCEPT
+      - name: Run topology tests with Gradle
+        run: ./gradlew topologyTest
+        env:
+          BUILD_JAVA_VERSION: ${{ matrix.java }}
```
vyazelenko, line 124: No need to set `BUILD_JAVA_VERSION` as it will be done in a previous step `Setup BUILD_JAVA_HOME & BUILD_JAVA_VERSION`. Otherwise one has to set each time we invoke the Gradle commands.

vyazelenko (changes requested):

vyazelenko (approved):

vyazelenko:
LGTM.

PR #1129, opened 2021-01-26
Event log disable

mikeb01:
Adds:
- `aeron.event.log.disabled`
- `aeron.event.archive.log.disabled`
- `aeron.event.cluster.log.disabled`

That allows disabling of specific events from the logging, e.g. `-Daeron.event.log=all` `-Daeron.event.log.disabled=FRAME_IN,FRAME_OUT` is everything except frames.

### DIFF
--- a/aeron-agent/src/main/java/io/aeron/agent/EventConfiguration.java
+++ b/aeron-agent/src/main/java/io/aeron/agent/EventConfiguration.java
@@ -52,10 +52,10 @@ final class EventConfiguration
 
     /**
      * Disabled Driver Event tags system property name. Follows the format specified for
-     * {@link EventConfiguration#ENABLED_EVENT_CODES_PROP_NAME}.  This property will disable any codes in the set
-     * specified there.  Defined on its own has no effect.
+     * {@link EventConfiguration#ENABLED_EVENT_CODES_PROP_NAME}. This property will disable any codes in the set
+     * specified there. Defined on its own has no effect.
      */
-    public static final String DISABLED_EVENT_CODES_PROP_NAME = "aeron.event.log.disabled";
+    public static final String DISABLED_EVENT_CODES_PROP_NAME = "aeron.event.log.disable";
 
     /**
      * Archive Event tags system property name. This is either:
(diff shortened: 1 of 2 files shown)


vyazelenko:
@mikeb01 The same needs to be done on the C driver side, i.e. to the `aeron_driver_agent.c`.

mikeb01:
@vyazelenko Was looking to get feedback on the approach first.  Adding it to the C driver will be no problem.

vyazelenko:
@mikeb01 LGTM. I would also add a new test to `EventConfigurationTest` which would invoke `EventConfiguration#init` and assert that the corresponding EnumSets are set according to the enabled/disabled properties.

mikeb01:
I initially avoided that test as it does manipulate system properties, which I'm a bit dubious about in unit tests, but it appears to work locally.

aeron-agent/src/main/java/io/aeron/agent/EventConfiguration.java
```diff
@@ -50,6 +50,13 @@
      */
     public static final String ENABLED_EVENT_CODES_PROP_NAME = "aeron.event.log";
 
+    /**
+     * Disabled Driver Event tags system property name. Follows the format specified for
+     * {@link EventConfiguration#ENABLED_EVENT_CODES_PROP_NAME}.  This property will disable any codes in the set
+     * specified there.  Defined on its own has no effect.
+     */
+    public static final String DISABLED_EVENT_CODES_PROP_NAME = "aeron.event.log.disabled";
```
mjpt777, line 58: I'd change the name to `aeron.event.log.disable`.

aeron-agent/src/main/java/io/aeron/agent/EventConfiguration.java
```diff
@@ -50,6 +50,13 @@
      */
     public static final String ENABLED_EVENT_CODES_PROP_NAME = "aeron.event.log";
 
+    /**
+     * Disabled Driver Event tags system property name. Follows the format specified for
+     * {@link EventConfiguration#ENABLED_EVENT_CODES_PROP_NAME}.  This property will disable any codes in the set
```
mjpt777, line 55: Double space between sentences breaks formatting in HTML.

PR #1132, opened 2021-02-10
Ensure session open event occurs before session messages

JPWatson:
We've observed messages arriving in the ClusteredService with a `null` session (when the service message mechanism isn't being used.).

If there is back-pressure at the point the consensus module attempts to append the session open event, it won't be retried as it is in an AUTHENTICATED state but session messages will be allowed through.

Note: I've not been able to track down why the ` else if (session.state() == CONNECTED)` branch existed so I may have missed something.


mjpt777:
This is a refinement on the approach. https://github.com/real-logic/aeron/commit/4b8822e71f90818834ab777317682afb49387841

PR #1143, opened 2021-03-08
Channel uri socket params

mikeb01:
- Allow for channels to configure SO_RCVBUF and SO_SNDBUF via the channel uri.
- Add support for the parameters to the builder in the Java client and C++ wrapper client.


aeron-driver/src/main/c/media/aeron_receive_destination.c
```diff
@@ -41,14 +41,17 @@ int aeron_receive_destination_create(
...
+
     if (context->udp_channel_transport_bindings->init_func(
         &_destination->transport,
         &channel->remote_data,
         &channel->local_data,
         channel->interface_index,
         0 != channel->multicast_ttl ? channel->multicast_ttl : context->multicast_ttl,
-        context->socket_rcvbuf,
-        context->socket_sndbuf,
+        socket_rcvbuf,
```
mjpt777, line 53: Maybe we should pass the channel to be more efficient and future proof.

mikeb01: Probably is a good clean up.  I think I will come back to it later as the current PR hasn't required a change there yet.  I am changing the congestion control supplier in a similar way that to pass through the `aeron_udp_channel_t` instead of the string value to avoid re-parsing.

PR #1144, opened 2021-03-08
[C] Lower CMake requirement to 3.5.1

jvirtanen:
Although CMake 3.6.1 is currently required, the build appears to work well with CMake 3.5.1. As Ubuntu 16.04 ships with that version, lower the CMake requirement to 3.5.1.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,8 +13,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-cmake_minimum_required(VERSION 3.6.1 FATAL_ERROR)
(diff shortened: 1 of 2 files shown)


jvirtanen:
> Thank you for the PR, but I am unsure whether we want to anchor ourselves to an older version of CMake to support Ubuntu 16.04, when that version will EOL in a couple of months.

The standard support ends in April but EOL takes place in 2024. 🙂 However, based on [the CMake home page](https://cmake.org/download/), it looks like installing a newer CMake on Ubuntu 16.04 would not be much of a hassle either, so closing this.

PR #1148, opened 2021-03-11
Channel uri receiver window

mikeb01:
- Adds the ability to set the receiver initial window in the URI.
- Adds validation to C and Java media drivers that compare the mtu to the socket sndbuf and receiver window to the socket rcvbuf.
- Adds validation to prevent creating subscriptions or publications that try to change the socket sndbuf and rcvbuf values.


aeron-client/src/main/java/io/aeron/ChannelUriStringBuilder.java
```diff
@@ -70,6 +70,7 @@
     private boolean isSessionIdTagged;
     private Integer socketSndbufLength;
     private Integer socketRcvbufLength;
+    private Integer receiverWindowLength;
```
mjpt777, line 73: The reset of this value is missing from the `clear()` method.

aeron-driver/src/main/c/media/aeron_receive_channel_endpoint.h
```diff
@@ -186,7 +188,10 @@ int aeron_receive_channel_endpoint_on_remove_publication_image(
     aeron_receive_channel_endpoint_t *endpoint, aeron_publication_image_t *image);
 
 int aeron_receiver_channel_endpoint_validate_sender_mtu_length(
-    aeron_receive_channel_endpoint_t *endpoint, size_t sender_mtu_length, size_t window_max_length);
+    aeron_receive_channel_endpoint_t *endpoint,
+    size_t sender_mtu_length,
+    size_t window_max_length,
+    aeron_driver_context_t *pStct);
```
mjpt777, line 194: Naming convention is not camel case in C.

aeron-driver/src/main/java/io/aeron/driver/Configuration.java
```diff
@@ -1693,7 +1689,8 @@ public static void validateInitialWindowLength(final int initialWindowLength, fi
     {
         if (mtuLength > initialWindowLength)
         {
-            throw new ConfigurationException("initial window length must be >= to MTU length: " + mtuLength);
+            throw new ConfigurationException(
+                "initial window length (" + initialWindowLength + ") must be >= MTU length (" + mtuLength + ")");
```
mjpt777, line 1693: `"mtuLength=" + mtuLength + " > initialWindowLength=" + initialWindowLength ` is easier to reason about given check.

aeron-driver/src/main/java/io/aeron/driver/Configuration.java
```diff
@@ -1767,55 +1764,43 @@ public static TerminationValidator terminationValidator()
...
-            if (maxSoRcvBuf < ctx.socketRcvbufLength())
-            {
-                System.err.format(
-                    "WARNING: Could not get desired SO_RCVBUF, adjust OS to allow %s: attempted=%d, actual=%d%n",
-                    SOCKET_RCVBUF_LENGTH_PROP_NAME,
-                    ctx.socketRcvbufLength(),
-                    maxSoRcvBuf);
-            }
+            System.err.format(
+                "WARNING: Could not get desired SO_SNDBUF, adjust OS to allow %s: attempted=%d, actual=%d%n",
```
mjpt777, line 1770: get -> set

aeron-driver/src/main/java/io/aeron/driver/Configuration.java
```diff
@@ -1767,55 +1764,43 @@ public static TerminationValidator terminationValidator()
...
 
-            if (ctx.initialWindowLength() > maxSoRcvBuf)
-            {
-                throw new ConfigurationException("window length greater than socket SO_RCVBUF, increase '" +
-                    Configuration.INITIAL_WINDOW_LENGTH_PROP_NAME +
-                    "' to match window: windowLength=" + ctx.initialWindowLength() + ", SO_RCVBUF=" + maxSoRcvBuf);
-            }
+        if (ctx.mtuLength() > soSndBuf)
+        {
+            throw new ConfigurationException(String.format(
```
mjpt777, line 1790: By default we should avoid use of string format.

mjpt777:
This is a nice step forward.

aeron-driver/src/main/java/io/aeron/driver/Configuration.java
mikeb01: I've changed to this, but I've modified the C code to have the same error format as it simplifies the error checking in the system tests and results in a more consistent experience for users.  However in the C driver the naming convention won't match.

PR #1151, opened 2021-03-17
Cluster log replication failure cases

mikeb01:
- More robust handling of log replication completion.
- Wait for leader commit position during followerLogReplication.
- Timeout leader if commit position not receieved.
- Track log replication progress and timeout if not progressing.
- Check for replicating too much data.


aeron-cluster/src/main/java/io/aeron/cluster/Election.java
```diff
@@ -688,21 +695,45 @@ private int followerLogReplication(final long nowNs)
...
+                    else if (nowNs >= leaderLogReplicationCommitPositionDeadlineNs)
+                    {
+                        throw new ClusterException("Timeout waiting for leader to commit replicated log position");
+                    }
+                    else
+                    {
+                        if (nowNs + ctx.leaderHeartbeatIntervalNs() >= timeOfLastUpdateNs)
+                        {
+                            timeOfLastUpdateNs = nowNs;
+                            consensusPublisher.appendPosition(
```
mjpt777, line 718: Only update time if the `appendPosition` is successful so it is retried on back pressure.

aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModuleAgent.java
```diff
@@ -2546,7 +2548,9 @@ LogReplication newLogReplication(
             logRecordingId,
             ctx.archiveContext().controlRequestStreamId(),
             leaderArchiveEndpoint,
-            stopPosition);
+            stopPosition,
+            LOG_REPLICATION_PROGRESS_CHECK_TIMEOUT_NS,
```
mjpt777, line 2552: Should these be members of the Election class?

aeron-cluster/src/main/java/io/aeron/cluster/Election.java
```diff
@@ -707,21 +714,49 @@ private int followerLogReplication(final long nowNs)
...
+                    logReplication.close();
+                    appendPosition = logReplication.position();
+                    consensusModuleAgent.logRecordingId(logReplication.recordingId());
+                    if (consensusPublisher.appendPosition(
+                        leaderMember.publication(), leadershipTermId, appendPosition, thisMember.id()))
+                    {
+                        timeOfLastUpdateNs = nowNs;
+                    }
+
+                    leaderLogReplicationCommitPositionDeadlineNs = nowNs + ctx.leaderHeartbeatTimeoutNs();
```
mjpt777, line 757: Should this be updated here?

aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModuleAgent.java
mikeb01: I was going to add configuration options for these, but instead I've reused the leader heartbeat interval/timeout.

aeron-cluster/src/main/java/io/aeron/cluster/Election.java
mikeb01: I've changed this around a bit now.  Previously I was relying on the transition to closed on the `LogReplication` to manage a lot of these updates.  Instead now the LogReplication is either done or not and the logic is similar in both cases.  It will also pulse out the append position as the replication progresses (if it takes longer than the leader heartbeat interval).  I've updating the `leaderLogReplicationCommitPositionDeadlineNs` whenever the append position moves forward (so eventually once log replication is done, will it have the one `leaderHeartbeatTimeoutNs` period in which to commit the final append position).

PR #1155, opened 2021-03-23
Fix a short write in sending the RecordingStarted event.

billsegall:
Add the length of the message header to what must be written.


mjpt777:
Thanks Bill.

PR #1156, opened 2021-03-24
Add termBaseLogPosition to NewLeadershipTerm

mikeb01:
As this allows log replication to run up until the beginning of the current term regardless of the number of terms behind the leader.  Update logging to include extra fields.  Remove duplicate test.


aeron-cluster/src/main/resources/cluster/aeron-cluster-codecs.xml
```diff
@@ -2,7 +2,7 @@
 <sbe:messageSchema xmlns:sbe="http://fixprotocol.io/2016/sbe"
                    package="io.aeron.cluster.codecs"
                    id="111"
-                   version="7"
+                   version="8"
```
mjpt777, line 5: We should only update this once between our public releases when changes occur.

PR #1172, opened 2021-05-03
[C] Prevent sending messages if the endpoint is not active.

mikeb01:


aeron-driver/src/main/c/media/aeron_receive_channel_endpoint.c
```diff
@@ -187,11 +187,17 @@ int aeron_receive_channel_endpoint_close(aeron_receive_channel_endpoint_t *endpo
 
 int aeron_receive_channel_endpoint_sendmsg(aeron_receive_channel_endpoint_t *endpoint, struct msghdr *msghdr)
 {
+    if (AERON_RECEIVE_CHANNEL_ENDPOINT_STATUS_ACTIVE != endpoint->conductor_fields.status)
```
mjpt777, line 190: Do we need to consider concurrent access?

mikeb01: You're right we do.  This field is set in the conductor and this change reads in the receiver.  Ordered write and volatile read should work.  However, I'm looking into the code in more detail and there may be a better way to handle this.

PR #1179, opened 2021-05-18
Fix NPE in cluster client after multiple redirects.

JPWatson:
Observed this exception during failover testing:

```
java.lang.NullPointerException
	at io.aeron.cluster.client.AeronCluster$AsyncConnect.awaitPublicationConnected(AeronCluster.java:1600)
	at io.aeron.cluster.client.AeronCluster$AsyncConnect.poll(AeronCluster.java:1529)
        ....
```

I believe thing can occur in the following scenario:
Step 1 => Step 2 => Step 3 => Redirect (member X) => Step 1 => Step 2 => Step 3 => Redirect (member Y) => Step 1 (Exception!)

This will occur due to the publication for member Y being closed at the point the client is redirected to member X.


mjpt777:
Thanks @JPWatson. Good catch.

PR #1180, opened 2021-05-19
Alternate re resolution tests

mikeb01:
Rewrite the CSV name resolver (Java and C):
- Works directly with host/port pairs
- Support toggle counters to control behaviour during tests.
- Tidied up how the tests configure the CSV name resolver for the C media driver.
- Use the name resolver in all Cluster Tests.
- Add (currently disabled) Cluster test that removes the name of the leader node at the same time as the node gets shutdown (Just like k8s & Docker).

### DIFF
--- a/aeron-test-support/src/main/java/io/aeron/test/driver/RedirectingNameResolver.java
+++ b/aeron-test-support/src/main/java/io/aeron/test/driver/RedirectingNameResolver.java
@@ -15,9 +15,13 @@
  */
 package io.aeron.test.driver;
 
+import io.aeron.Aeron;
 import io.aeron.driver.DefaultNameResolver;
+import io.aeron.driver.MediaDriver;
 import io.aeron.driver.NameResolver;
 import org.agrona.collections.Object2ObjectHashMap;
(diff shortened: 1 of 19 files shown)


aeron-test-support/src/main/java/io/aeron/test/driver/RedirectingNameResolver.java
```diff
@@ -0,0 +1,163 @@
...
+                counters.valuesBuffer(),
+                nameCounterId.get());
+
+            nameCounter.set(operationValue);
+        }
+
+        return counterFound;
+    }
+
+    private static final class NameEntry
```
mjpt777, line 112: Best not to use private classes as this will result in the generation of bridge classes < Java 11.

mikeb01: Should we use package-private inner classes or declare them as a separate unit?

mjpt777: Package private is a good way to go.

PR #1188, opened 2021-06-08
Add nanoTime to AeronArchive.Context

najones:
I have found that when using a dedicated media driver for Aeron Archive, I have needed to supply a nanoClock in order to debug, as by default the SystemNanoClock.INSTANCE is used and I could not see a way to override that.


mjpt777:
This is not used in the code so looks odd adding it.

najones:
> This is not used in the code so looks odd adding it.

It's used in line 931 if an aeron instance was not supplied.

mjpt777:
> It's used in line 931 if an aeron instance was not supplied.
OK, I missed that. Thanks.

PR #1189, opened 2021-06-10
Store errors in distinct log for cluster tests

mikeb01:
- Remove the try/catch block to track errors in cluster tests, instead uses a test watcher.
- Store errors from media driver in cnc.dat and cluster errors in their own distinct error log (so that it can be picked up by the cluster data capture used for building artifacts in CI).
- Check error logs on success and fail if errors are present.
- Specify errors to include in validation check (allows exclusion for expected errors).
- Modify ErrorStat (Java & C) to print out errors from a raw distinct error log

### DIFF
--- a/.github/workflows/ci-low-cadence.yml
+++ b/.github/workflows/ci-low-cadence.yml
@@ -20,11 +20,6 @@ jobs:
         java: [ '8', '11', '16', '17-ea' ]
         os: [ 'ubuntu-20.04', 'windows-latest' ]
     steps:
-      - name: Setup java
-        uses: actions/setup-java@v2
-        with:
-          distribution: 'zulu'
-          java-version: ${{ matrix.java }}
       - name: Checkout code
         uses: actions/checkout@v2
       - name: Cache Gradle dependencies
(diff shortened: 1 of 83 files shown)


aeron-system-tests/src/test/java/io/aeron/cluster/AppointedLeaderTest.java
```diff
@@ -16,50 +16,63 @@
...
     @Test
     @Timeout(20)
     public void shouldConnectAndSendKeepAlive()
     {
-        try (TestCluster cluster = aCluster().withStaticNodes(3).withAppointedLeader(LEADER_ID).start())
-        {
-            final TestNode leader = cluster.awaitLeader();
-            assertEquals(LEADER_ID, leader.index());
-            assertEquals(Cluster.Role.LEADER, leader.role());
+        final TestCluster cluster = aCluster().withStaticNodes(3).withAppointedLeader(LEADER_ID).start();
```
mjpt777, line 49: How does the `TestCluster` get closed with this approach?

.github/workflows/ci-low-cadence.yml
```diff
@@ -20,6 +20,11 @@ jobs:
         java: [ '8', '11', '16', '17-ea' ]
         os: [ 'ubuntu-20.04', 'windows-latest' ]
     steps:
+      - name: Setup java
```
vyazelenko, line 23: Caching jobs are not executing Gradle scripts. So it's ok to leave `setup-java` where it was. Also in this particular case we call `setup-java` twice:
- First time to install the target JVM (e.g. 17-ea).
- Second time to install the JDK 8 with which we are going to execute the Gradle script. 

The two are different, because Gradle is not able to run on JDK 17 yet hence we must run the script with a version that always works but then specify the target JVM we want for our build via the `BUILD_JAVA_HOME` and `BUILD_JAVA_VERSION` environment variables.

mikeb01: I'll revert that change out and go with the one that you've already made to avoid the auto-detect.

PR #1192, opened 2021-06-28
[Java] Add simple system tests for packet timestamps validate C drive…

mikeb01:
…r, verify error response for Java driver.  Add rejection logic for packet timestamps in Java driver.

### DIFF
--- a/aeron-system-tests/src/test/java/io/aeron/TimestampingSystemTest.java
+++ b/aeron-system-tests/src/test/java/io/aeron/TimestampingSystemTest.java
@@ -1,0 +1,71 @@
+package io.aeron;
+
+import io.aeron.driver.MediaDriver;
+import io.aeron.exceptions.RegistrationException;
+import io.aeron.test.Tests;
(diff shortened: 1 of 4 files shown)


aeron-system-tests/src/test/java/io/aeron/TimestampingSystemTest.java
```diff
@@ -0,0 +1,86 @@
...
+            Aeron aeron = Aeron.connect(new Aeron.Context().aeronDirectoryName(driver.aeronDirectoryName())))
+        {
+            assertThrows(
+                RegistrationException.class,
+                () -> aeron.addSubscription(CHANNEL_WITH_PACKET_TIMESTAMP, 1000));
+        }
+    }
+
+    @Test
+    void shouldSupportPacketTimestampsInJavaDriver()
```
tmontgomery, line 55: Shouldn't this be `inCDriver`?

mikeb01: Yes it should be.

PR #1193, opened 2021-06-29
[C] Restructure the setup of destination construction.

mikeb01:
So that the udp_channel of both the enclosing multi-destination channel and destination are provided.  For simple channels the same value is provided for both parameters.


tmontgomery:
LGTM

PR #1195, opened 2021-06-30
C driver send and receive timestamps

mikeb01:


tmontgomery (approved):

mjpt777:
I'm not sold on the `rx` and `receive` timestamp naming.  I wonder if there is better potentially name for `rx`.

mikeb01:
I picked `rx` as it matches with the kernel naming for the same timestamp.  One option would be to stick with receive but use a qualifying term.  E.g. packet receive timestamp and endpoint receive timestamp.  `pkt-rcv-ts-offset` and `endpoint-rcv-ts-offset` as uri parameter names?

mjpt777:
@mikeb01 I think that naming is getting better. How about  `media-rcv-ts-off', `aeron-rcv-ts-off`? Or some shortening of that?

mikeb01:
` media-rcv-ts-off` is good, but I think `aeron-rcv-ts-off` too vague as the whole system is Aeron so that timestamp could be taken anywhere.  I'll change `rx-ts-offset` to media tomorrow and a further think on the other one.

mikeb01:
I think channel works.

PR #1196, opened 2021-07-08
[C++] Archive Client AsyncConnect::poll: Avoid proceeding to step 1 until m_subscription is non-null

vaci:
findSubscription() may return a null pointer that is used unchecked in step 1.


vaci:
We also note that the timeout exception might access m_subscription and m_publication before those pointers are assigned.

tmontgomery:
> We also note that the timeout exception might access m_subscription and m_publication before those pointers are assigned.

If you want, add changes for these as well to this PR.

vaci:
Added a patch for the timeout message.

PR #1203, opened 2021-07-28
[Java] Make the default ingress channel null. Add constants to be used in cluster tests.

mikeb01:


aeron-system-tests/src/test/java/io/aeron/cluster/MultiModuleSharedDriverTest.java
```diff
@@ -126,6 +127,8 @@ public void shouldSupportTwoSingleNodeClusters()
                 final EgressListener egressListener = (clusterSessionId, timestamp, buffer, offset, length, header) ->
                     egress.set(buffer.getStringWithoutLengthAscii(offset, length));
 
+                Tests.sleep(1_000);
```
mjpt777, line 130: Why is this needed?

mikeb01: It's not, that was me trying to debug a failure.  Removed.

PR #1209, opened 2021-08-13
Name resolver address change event log

mikeb01:
- Adds an additional DistinctErrorLog for events that don't qualify as errors (works in a similar way to the error log).
- Logs an event when the address for a given name entry changes.
- Changes logic for DriverNameCache so that will update when the name is different.

### DIFF
--- a/aeron-driver/src/main/java/io/aeron/driver/DriverNameResolverCache.java
+++ b/aeron-driver/src/main/java/io/aeron/driver/DriverNameResolverCache.java
@@ -17,8 +17,10 @@ package io.aeron.driver;
 
 import io.aeron.protocol.ResolutionEntryFlyweight;
 import org.agrona.collections.ArrayListUtil;
+import org.agrona.concurrent.errors.DistinctErrorLog;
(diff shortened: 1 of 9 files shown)


aeron-driver/src/main/java/io/aeron/driver/DriverNameResolverCache.java
```diff
@@ -79,10 +83,13 @@ void addOrUpdateEntry(
             entry.timeOfLastActivityMs = nowMs;
             entry.deadlineMs = nowMs + timeoutMs;
 
-            if (port != entry.port || byteSubsetEquals(address, entry.address, addressLength))
+            if (port != entry.port || !byteSubsetEquals(address, entry.address, addressLength))
             {
                 entry.address = Arrays.copyOf(address, addressLength);
                 entry.port = port;
+
+                eventLog.record(new RuntimeException(
```
mjpt777, line 91: Should we record and `AeronEvent` here to save on the stack trace and allocation?

mikeb01: Yes

PR #1219, opened 2021-09-02
Remove custom error files and errorLog handlers

mikeb01:
This is some more work to make the error logging in system tests cleaner by removing use of the custom error log handlers and using the defaults.  This removes the additional error log files that were being created and uses the cnc file or mark files as appropriate.

It also breaks the dependency on TestCluster so it only needs an instance of the DataCollector in order to work correctly and will collect up the locations on failed builds as well.


aeron-system-tests/src/test/java/io/aeron/cluster/MultiModuleSharedDriverTest.java
```diff
@@ -47,30 +48,44 @@
...
+    @AfterEach
+    void tearDown()
+    {
+        assertEquals(0, clusterTestWatcher.errorCount(), "Errors observed in " + this.getClass().getSimpleName());
+    }
+
     @SuppressWarnings("methodlength")
     @Test
-    @InterruptAfter(20)
+    @InterruptAfter(20_000)
```
vyazelenko, line 70: Was the timeout increased for debug?

aeron-test-support/src/main/java/io/aeron/test/DataCollector.java
```diff
@@ -146,14 +169,67 @@ public Path dumpData(final String destinationDir)
...
+    public Collection<Path> allLocations()
+    {
+        return locations;
+    }
+
+    private static Stream<Path> list(final Path p)
+    {
+        try
+        {
+            return Files.list(p);
```
vyazelenko, line 225: `DataCollector` contains not just directories but also the individual files and thus `Files.list(p)` will fail when one is encountered. This can be seen from the failing C tests which add stdout/stderr files to the `DataCollector`.

You could use `java.nio.file.Files#find` method instead which works fine if `start` is a file.
To list all of the files with a particular name you can use `java.nio.file.Files#find` method starting at the

vyazelenko (changes requested):

aeron-test-support/src/main/java/io/aeron/test/DataCollector.java
mikeb01: Done.

aeron-test-support/src/main/java/io/aeron/test/DataCollector.java
```diff
@@ -138,22 +157,98 @@ public Path dumpData(final String destinationDir)
...
+     * Gets a collection of all registered locations.
+     *
+     * @return list of all locations.
+     */
+    public Collection<Path> allLocations()
+    {
+        return locations;
+    }
+
+    private static Stream<Path> list(final Path p)
```
vyazelenko, line 218: Seems to be unused.

vyazelenko (approved):
LGTM.

aeron-test-support/src/main/java/io/aeron/test/DataCollector.java
mikeb01: Yes, will come through in a subsequent PR.

aeron-cluster/src/main/java/io/aeron/cluster/service/ClusterMarkFile.java
```diff
@@ -176,6 +178,31 @@ public ClusterMarkFile(
...
+    /**
+     * Determines if this path name matches the service mark file name pattern
+     *
+     * @param p to examine
+     * @param attributes ignored, only needed for BiPredicate signature matching
+     * @return true if the name matches
+     */
+    public static boolean isServiceMarkFile(final Path p, final BasicFileAttributes attributes)
+    {
+        return p.getFileName().toString().startsWith(SERVICE_FILENAME_PREFIX) &&
```
mjpt777, line 190: Best to avoid the double allocation.

PR #1222, opened 2021-09-09
Always use error log

mikeb01:
- Setup errorHandlers so that errors are always sent to the configured distinct error log as well as the user specified error handler for Driver, Archive, ConsensusModule, ClusteredServiceContainer, and ClusterBackup.  
- Add stderr output suppression for specific tests.


vyazelenko (approved):
LGTM

tmontgomery (approved):
LGTM

PR #1233, opened 2021-09-23
[Java] Add support for changing term buffer length and MTU in the Cluster log.

vyazelenko:
This PR adds a feature that allows a Cluster to be restarted with the term buffer length and/or MTU changed. This requires truncating an existing recording log to the latest valid snapshot, i.e. Cluster is starting from a snapshot only. The new API to do such truncation is provided via the `ClusterTool` (i.e. `io.aeron.cluster.ClusterTool#truncateRecordingLog` and `truncate-recording-log` command). 

Besides this feature this PR also contains a fix for keeping entries properly sorted in the `RecodingLog`.

### DIFF
--- a/aeron-cluster/src/main/java/io/aeron/cluster/ClusterTool.java
+++ b/aeron-cluster/src/main/java/io/aeron/cluster/ClusterTool.java
@@ -56,6 +56,7 @@ import java.util.function.Consumer;
 
 import static io.aeron.Aeron.NULL_VALUE;
 import static java.nio.ByteOrder.LITTLE_ENDIAN;
+import static java.nio.file.StandardCopyOption.*;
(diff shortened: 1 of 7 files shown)


mjpt777:
Using the term "truncate" on a recording log is confusing. It implies the log on the archive is truncated.  We need to make it clear that this function creates a new recording log, old one is backed up, that allows the cluster to restart from the latest snapshot with a new cluster log. The old log remains untouched and can be archived or deleted as a separate operation. Something like "rebase log from snapshot.

vyazelenko:
@mjpt777 I pushed a change that renamed `truncateRecordingLog` to `seedRecordingLogFromSnapshot` (`seed-recording-log-from-snapshot`) and adds a backup of original recording log as `recording.log.bak` file. JavaDoc updated to reflect these changes.

PR #1235, opened 2021-09-29
[Java] Timely free direct ByteBuffers used by BufferBuilder/FragmentAssembler.

vyazelenko:
- `BufferBuilder.compact` - should resize only if capacity is not the same as the target capacity.
- Resizing internal buffer should free old direct ByteBuffer.
- `BufferBuilder.free` - adds an ability to free internal direct ByteBuffer.
- `FragmentAssembler/ControlledFragmentAssembler` - invoke `BufferBuilder.free` when direct buffers are used.


mjpt777:
For now we should not do this.

PR #1242, opened 2021-10-07
Add the ability to specify a RecordingSignalConsumer via AeronArchive.Context

mikeb01:
And poll for signals.


tmontgomery:
I get that. I guess my point is more why would there be a lock on the add/remove, but not on the onSignal? Aren't both controlled by the application? Maybe this can be clarified by Javadoc, though.

mikeb01:
I suspect this is potentially a Javadoc issue.  If an application is sharing AeronArchive across threads then, for signals, they should use add/remove/pollForEvents all of which are locked.  If they adhere to that all calls through to `RecordingSignalConsumer::onSignal`  (which is I assume what you mean by 'the onSignal') should be within the scope of the lock in AeronArchive.  I don't see any requirement for the application to call `RecordingSignalConsumer::onSignal` directly.

Do you have an example of a case where these assumptions would break?

PR #1252, opened 2021-11-03
Avoid missing SYNC signal due to race between the position + isClosed/isEndOfStream being read.

JPWatson:


mjpt777:
Good catch.

PR #1261, opened 2021-11-26
Support Static Linking

mikeb01:
Add aeron_symbol_table and use it to resovle internal symbols.  
Encapsulate dlsym lookups.  
Separate object and function pointers and correctly deal generic versions of each.
Remove multiple instances of warning ignores for -Wpendantic.


aeron-client/src/main/c/util/aeron_symbol_table.c
```diff
@@ -0,0 +1,133 @@
...
+        const char *name = table[i].name;
+
+        if (NULL == alias || NULL == name)
+        {
+            break;
+        }
+
+        if (0 == strcmp(alias, symbol) || 0 == strcmp(name, symbol))
+        {
+            result = table[i].object;
```
mjpt777, line 50: Why not break here?

mikeb01: Fixed.

mikeb01:
I've added a unit test and switched over to specifying a table length instead of relying on a NULL terminator.  It will still early exit on NULL fields in the table.

aeron-driver/src/main/c/agent/aeron_driver_agent.c
```diff
@@ -1011,7 +1010,7 @@ int aeron_driver_agent_init_logging_events_interceptors(aeron_driver_context_t *
 
         incoming_bindings->meta_info.name = "logging";
         incoming_bindings->meta_info.type = "interceptor";
-        incoming_bindings->meta_info.source_symbol = "aeron_driver_agent_context_init";
+        incoming_bindings->meta_info.source_symbol = (void (*)(void))aeron_driver_agent_context_init;
```
tmontgomery, line 1013: Should `void (*)(void)` be a typedef instead?

aeron-driver/src/main/c/media/aeron_udp_channel_transport_bindings.h
```diff
@@ -227,7 +227,7 @@ struct aeron_udp_channel_interceptor_bindings_stct
         const char *name;
         const char *type;
         const aeron_udp_channel_interceptor_bindings_t *next_interceptor_bindings;
-        const void *source_symbol;
+        void (*source_symbol)(void);
```
tmontgomery, line 230: this should be a typedef for `void (*)(void)`

aeron-client/src/main/c/util/aeron_symbol_table.c
```diff
@@ -0,0 +1,147 @@
...
+    {
+        const char *alias = table[(int)i].alias;
+        const char *name = table[(int)i].name;
+
+        if (NULL == alias || NULL == name)
+        {
+            break;
+        }
+
+        if (0 == strcmp(alias, symbol) || 0 == strcmp(name, symbol))
```
tmontgomery, line 47: should consider looking at length instead of a naked `strcmp` that may not terminate.

aeron-client/src/main/c/aeron_agent.c
```diff
@@ -282,6 +283,19 @@ aeron_idle_strategy_t aeron_idle_strategy_backoff =
         aeron_idle_strategy_backoff_state_init_args
     };
 
+static const aeron_symbol_table_obj_t aeron_idle_strategy_table[] =
+    {
+        { "sleeping", "aeron_idle_strategy_sleeping", &aeron_idle_strategy_sleeping },
```
mjpt777, line 288: Name should be `sleep` rather than `sleeping` to have consistent naming.

mikeb01: I agree, but I'd like to make that a separate change to this PR.  The current PR preserves behaviour: https://github.com/real-logic/aeron/blob/master/aeron-client/src/main/c/aeron_agent.c#L301.  I think it also requires a "sleep-ms" to be consistent with the Java driver as well.

PR #1265, opened 2021-12-07
Ensure that the log publication always has a consistent initial-term-id.

mikeb01:
And:
- Reset election state correctly after log truncation.

### DIFF
--- a/aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModuleAgent.java
+++ b/aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModuleAgent.java
@@ -27,7 +27,6 @@ import io.aeron.cluster.client.ClusterException;
 import io.aeron.cluster.codecs.MessageHeaderDecoder;
 import io.aeron.cluster.codecs.*;
 import io.aeron.cluster.service.*;
-import io.aeron.driver.Configuration;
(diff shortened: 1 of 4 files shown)


aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModuleAgent.java
```diff
@@ -1373,6 +1373,24 @@ int addLogPublication()
...
+            {
+                channelUri.put(TERM_ID_PARAM_NAME, "0");
+            }
+
+            if (!channelUri.containsKey(TERM_OFFSET_PARAM_NAME))
+            {
+                channelUri.put(TERM_OFFSET_PARAM_NAME, "0");
+            }
+        }
+
```
vyazelenko, line 1393: Shouldn't these be set together for a correct result? For example if the `init-term-id` is set to `87` but the `term-id` is not set this code will set `term-id=0` which will break.

Could we check if any of the values is missing and set all three to zero at that point?

mjpt777: These values should not be set in the template. We should error if they are.

mjpt777:
It is worth reflecting on how this will play with rebasing a system from just a snapshot.

PR #1266, opened 2021-12-13
[Java] Add an admin command to trigger Cluster snapshot.

vyazelenko:
This PR adds new API to `AeronCluster` which allows sending command to the Cluster to trigger a snapshot of the Cluster state. It includes the following:
- New `AuthorisationService` and `AuthorisationServiceSupplier` interfaces.
- `Archive.Context.authorisationServiceSupplier` and `aeron.archive.authorisation.service.supplier` property so that an `AuthorisationServiceSupplier` can be set for the Archive. Default supplier allows any command to be executed.
- `ConsensusModule.Context.authorisationServiceSupplier` and `aeron.cluster.authorisation.service.supplier` property so that an `AuthorisationServiceSupplier` can be set for the ConsensusModule. Default supplier forbids all commands.
- `AdminRequest` and `AdminResponse` messages added to the Cluster wire protocol.
- Support for `AdminResponse` was added to the client API, i.e. the `EgressListener#onAdminResponse` and `ControlledEgressListener#onAdminResponse` API.
- Support for sending `AdminRequest` of type `AdminRequestType.SNASPHOT`, i.e. `AeronCluster#sendAdminRequestToTakeASnapshot`.
- `ConsensusModuleAgent#onAdminRequest` implementation which triggers a snapshot by switching the `ClusterControl` into `ClusterControl.ToggleState.SNAPSHOT` state.

### DIFF
--- a/aeron-client/src/main/java/io/aeron/security/AuthorisationService.java
+++ b/aeron-client/src/main/java/io/aeron/security/AuthorisationService.java
@@ -26,20 +26,23 @@ public interface AuthorisationService
     /**
      * An {@link AuthorisationService} instance that allows every command.
      */
-    AuthorisationService ALLOW_ALL = (commandTemplateId, encodedCredentials) -> true;
+    AuthorisationService ALLOW_ALL = (templateId, type, encodedCredentials) -> true;
 
     /**
      * An {@link AuthorisationService} instance that forbids all commands.
      */
-    AuthorisationService DENY_ALL = (commandTemplateId, encodedCredentials) -> false;
+    AuthorisationService DENY_ALL = (templateId, type, encodedCredentials) -> false;
 
     /**
      * Checks if the client with the specified credentials is allowed to perform an operation indicated by the
      * given {@code templateId}.
      *
-     * @param templateId  of the command being checked, i.e. an SBE message id.
+     * @param templateId         of the command being checked, i.e. an SBE message id.
+     * @param type               optional type for the command being checked, may be {@code null}. For example for
+     *                           an admin request in the cluster it will contain {@code AdminRequestType} value which
+     *                           denotes the exact kind of the request.
      * @param encodedCredentials from the Challenge Response. Will not be {@code null}, but may be 0 length.
      * @return {@code true} if the client is authorised to execute the command or {@code false} otherwise.
      */
-    boolean isAuthorised(int templateId, byte[] encodedCredentials);
+    boolean isAuthorised(int templateId, Object type, byte[] encodedCredentials);
 }
--- a/aeron-client/src/main/java/io/aeron/security/AuthorisationService.java
+++ b/aeron-client/src/main/java/io/aeron/security/AuthorisationService.java
@@ -35,11 +35,11 @@ public interface AuthorisationService
 
     /**
      * Checks if the client with the specified credentials is allowed to perform an operation indicated by the
-     * given {@code commandTemplateId}.
+     * given {@code templateId}.
      *
-     * @param commandTemplateId  of the command being checked, i.e. an SBE message id.
+     * @param templateId  of the command being checked, i.e. an SBE message id.
      * @param encodedCredentials from the Challenge Response. Will not be {@code null}, but may be 0 length.
      * @return {@code true} if the client is authorised to execute the command or {@code false} otherwise.
      */
-    boolean isAuthorised(int commandTemplateId, byte[] encodedCredentials);
+    boolean isAuthorised(int templateId, byte[] encodedCredentials);
 }
--- a/aeron-client/src/main/java/io/aeron/security/AuthorisationService.java
+++ b/aeron-client/src/main/java/io/aeron/security/AuthorisationService.java
@@ -15,8 +15,6 @@
  */
 package io.aeron.security;
 
-import org.agrona.ErrorHandler;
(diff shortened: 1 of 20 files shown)


aeron-client/src/main/java/io/aeron/security/AuthorisationService.java
```diff
@@ -0,0 +1,48 @@
...
+     * given {@code templateId}.
+     *
+     * @param templateId         of the command being checked, i.e. an SBE message id.
+     * @param type               optional type for the command being checked, may be {@code null}. For example for
+     *                           an admin request in the cluster it will contain {@code AdminRequestType} value which
+     *                           denotes the exact kind of the request.
+     * @param encodedCredentials from the Challenge Response. Will not be {@code null}, but may be 0 length.
+     * @return {@code true} if the client is authorised to execute the command or {@code false} otherwise.
+     */
+    boolean isAuthorised(int templateId, Object type, byte[] encodedCredentials);
```
mikeb01, line 47: Would it make more sense to be using `encodedPrincipal` here?  I would have thought that Authenciation would validate the credentials for a principal and the actions they are authorised to carry out would be keyed by principal.

mikeb01: In the ConsensusModuleAgent, we are passing `ClusterSession::encodedPrincipal`, so I think this is just a naming thing?

aeron-system-tests/src/test/java/io/aeron/cluster/ClusterTest.java
```diff
@@ -1632,6 +1641,384 @@ public void shouldRecoverWhenFollowersIsMultipleTermsBehindFromEmptyLogAndPartia
...
+    {
+        cluster = aCluster().withStaticNodes(3).start();
+        systemTestWatcher.cluster(cluster);
+
+        final TestNode leader = cluster.awaitLeader();
+        final List<TestNode> followers = cluster.followers();
+
+        final long requestCorrelationId = System.nanoTime();
+        final MutableBoolean responseReceived = new MutableBoolean();
+        cluster.egressListener(new EgressListener()
```
mikeb01, line 1656: Could the EgressListener logic be moved over to the TestCluster or extracted elsewhere to avoid inlining the definition in the tests?

aeron-system-tests/src/test/java/io/aeron/cluster/ClusterTest.java
```diff
@@ -1632,6 +1641,384 @@ public void shouldRecoverWhenFollowersIsMultipleTermsBehindFromEmptyLogAndPartia
...
+            Tests.yield();
+        }
+
+        while (!responseReceived.get())
+        {
+            client.pollEgress();
+            Tests.yield();
+        }
+
+        assertEquals(0, cluster.getSnapshotCount(leader));
```
mikeb01, line 1707: This sort of validation can be tricky as your are checking to see if something doesn't happen.  The test should pause or retry for a period of time during which it would of been able to take the snapshot if the code was behaving incorrectly - perhaps a second or two.  E.g. I deliberately changed the code to send back an error response, but still carry out the snapshot and this test passed.

aeron-client/src/main/java/io/aeron/security/AuthorisationService.java
vyazelenko: Renamed and updated the JavaDoc.

aeron-system-tests/src/test/java/io/aeron/cluster/ClusterTest.java
vyazelenko: Good point. I'll update the test.

aeron-system-tests/src/test/java/io/aeron/cluster/ClusterTest.java
vyazelenko: Done.

aeron-cluster/src/main/java/io/aeron/cluster/EgressPublisher.java
```diff
@@ -128,4 +130,41 @@ boolean newLeader(
...
+            final long result = session.tryClaim(length, bufferClaim);
+            if (result > 0)
+            {
+                adminResponseEncoder
+                    .wrapAndApplyHeader(bufferClaim.buffer(), bufferClaim.offset(), messageHeaderEncoder)
+                    .clusterSessionId(session.id())
+                    .correlationId(correlationId)
+                    .requestType(adminRequestType)
+                    .responseCode(responseCode)
+                    .message(message)
```
mjpt777, line 159: Try claim should not be used when the message count be potentially fragmented due to variable length data.

aeron-cluster/src/main/java/io/aeron/cluster/client/AeronCluster.java
```diff
@@ -642,8 +702,41 @@ else if (SessionEventDecoder.TEMPLATE_ID == templateId)
                     sessionEventDecoder.detail());
             }
         }
+        else if (AdminResponseDecoder.TEMPLATE_ID == templateId)
+        {
+            adminResponseDecoder.wrap(
```
mjpt777, line 707: What if the response fails to be sent due to back-pressure?

PR #1267, opened 2021-12-14
[Java] Include Javadoc to describe retry actions on internal messaging.

mikeb01:


aeron-cluster/src/main/java/io/aeron/cluster/service/Cluster.java
```diff
@@ -220,6 +220,22 @@ public static Role get(final AtomicCounter counter)
      * {@link ClusteredService#onSessionOpen(ClientSession, long)}, or
      * {@link ClusteredService#onSessionClose(ClientSession, long, CloseReason)}.
      * If applied to other events then they are not guaranteed to be reliable.
+     * <p>
+     * Callers of this method should loop until the method succeeds.
+     *
+     * <pre>
```
mjpt777, line 226: The markup for code needs to be enhanced, e.g. https://github.com/real-logic/aeron/blob/master/aeron-cluster/src/main/java/io/aeron/cluster/service/Cluster.java#L279

PR #1269, opened 2021-12-20
[C] Fix building error under arm64/gcc

dpmabo:
Fix building error under arm64/gcc due to stdatomic.h included in cpp file. 

The issue is as below:
```text
arch: aarch64
system: aws t4g.large
os: Ubuntu 18.04.6 LTS
gcc: 7.5.0
command: ./cppbuild/cppbuild

[ 53%] Building CXX object aeron-client/src/test/c/CMakeFiles/spsc_rb_test.dir/concurrent/aeron_spsc_rb_test.cpp.o
In file included from aeron/aeron-client/src/main/c/concurrent/aeron_atomic64_c11.h:22,
                 from aeron/aeron-client/src/main/c/concurrent/aeron_atomic.h:25,
                 from aeron/aeron-client/src/main/c/concurrent/aeron_rb.h:22,
                 from aeron/aeron-client/src/main/c/concurrent/aeron_spsc_rb.h:20,
                 from aeron/aeron-client/src/test/c/concurrent/aeron_spsc_rb_test.cpp:26:
/usr/lib/gcc/aarch64-linux-gnu/9/include/stdatomic.h:40:9: error: ‘_Atomic’ does not name a type
   40 | typedef _Atomic _Bool atomic_bool;
      |         ^~~~~~~
/usr/lib/gcc/aarch64-linux-gnu/9/include/stdatomic.h:41:9: error: ‘_Atomic’ does not name a type
   41 | typedef _Atomic char atomic_char;
      |         ^~~~~~~
/usr/lib/gcc/aarch64-linux-gnu/9/include/stdatomic.h:42:9: error: ‘_Atomic’ does not name a type
   42 | typedef _Atomic signed char atomic_schar;
      |         ^~~~~~~
```

### DIFF
--- a/aeron-client/src/main/c/concurrent/aeron_atomic64_c11.h
+++ b/aeron-client/src/main/c/concurrent/aeron_atomic64_c11.h
@@ -19,7 +19,13 @@
 
 #include <stdbool.h>
 #include <stdint.h>
+#ifndef __cplusplus
 #include <stdatomic.h>
+#else
+#include <atomic>
+#define _Atomic(X) std::atomic< X >
+using namespace std;
+#endif
 
 #if defined(__clang__)
 #pragma clang diagnostic push
@@ -54,14 +60,14 @@ while (false) \
 #define AERON_GET_AND_ADD_INT64(original, dst, value) \
 do \
 { \
-    original = atomic_fetch_add((_Atomic(int64_t) *)&dst, value); \
+    original = atomic_fetch_add((_Atomic(int64_t) *)&dst, (int64_t)value); \
 } \
 while (false) \
 
 #define AERON_GET_AND_ADD_INT32(original, dst, value) \
 do \
 { \
-    original = atomic_fetch_add((_Atomic(int32_t) *)&dst, value); \
+    original = atomic_fetch_add((_Atomic(int32_t) *)&dst, (int32_t)value); \
 } \
 while (false) \
 
--- a/aeron-client/src/main/c/concurrent/aeron_atomic.h
+++ b/aeron-client/src/main/c/concurrent/aeron_atomic.h
@@ -22,7 +22,11 @@
 #if defined(AERON_COMPILER_GCC) && defined(AERON_CPU_X64)
     #include "concurrent/aeron_atomic64_gcc_x86_64.h"
 #elif defined(AERON_COMPILER_GCC) && defined(AERON_CPU_ARM)
+#if defined(__cplusplus) && !defined(_Atomic)
+    #include "concurrent/aeron_atomic64_cpp.h"
+#else
     #include "concurrent/aeron_atomic64_c11.h"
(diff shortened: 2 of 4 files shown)


aeron-client/src/main/c/concurrent/aeron_atomic64_c11.h
```diff
@@ -54,14 +60,14 @@ while (false) \
 #define AERON_GET_AND_ADD_INT64(original, dst, value) \
 do \
 { \
-    original = atomic_fetch_add((_Atomic(int64_t) *)&dst, value); \
+    original = atomic_fetch_add((_Atomic(int64_t) *)&dst, (int64_t)value); \
```
mjpt777, line 63: Why add a cast as this can hide a real type issue?

dpmabo: because in function aeron_term_appender_get_and_add_raw_tail, value type is size_t, this will not compile 
for std::atomic<int64_t> in cpp context.
This is a quick fix.
We could also change all occurrences of aeron_term_appender_get_and_add_raw_tail.

```cpp
 // in c/concurrent/aeron_term_appender.h
 inline int64_t aeron_term_appender_get_and_add_raw_tail(volatile int64_t *addr, size_t aligned_length)
 {
     int64_t result;
     AERON_GET_AND_ADD_INT64(result, *addr, aligned_length);
     return result;
 }
```

mjpt777: This is a more appropriate fix. https://github.com/real-logic/aeron/commit/2b923e7e0a3f066e93ab14f2cdc2761340086735

dpmabo:
add aeron_atomic64_cpp.h to adapt to g++  without touch original aeron_atomic64_c11.h.
This has minimum side effect to original logic.
Test passed on mac m1 and aws t4g.large.

aeron-client/src/main/c/concurrent/aeron_atomic64_c11.h
dpmabo: please review the new commits below.

tmontgomery:
This is the wrong direction.

The test is meant to be testing the C atomics and not shadowing them out to use C++. i.e. this change only affects the tests and removes them as useful as the tests don't test against the right compilation for C.

Also, Aeron C++ API has the `std::atomic` already and Apple M1 compiles fine. And this is a g++ issue that has been well known for a long time.

To handle the test issues, the tests probably need to instantiate the unit under test as a .c compilation unit that the C++ can call through to. In other words, a shim between the C++ and C side so that the actual tests can be against a C compiled version. Which is the intent.

dpmabo:
it's better to update the testcases rather than changing header files.
I'll close the pr and submit a issue to you.

PR #1272, opened 2021-12-20
Add background snapshots

mikeb01:
- Add control toggle option BACKGROUND_SNAPSHOTS
- Leader selects a follower to take the snapshot
- Non-snapshotting members notice the snapshot entry in the log and will try to replicate the snapshot locally.

### DIFF
--- a/aeron-agent/src/main/java/io/aeron/agent/ClusterEventCode.java
+++ b/aeron-agent/src/main/java/io/aeron/agent/ClusterEventCode.java
@@ -96,7 +96,12 @@ public enum ClusterEventCode implements EventCode
     /**
      * Event when a session is closed
      */
-    APPEND_SESSION_CLOSE(14, ClusterEventDissector::dissectAppendCloseSession);
+    APPEND_SESSION_CLOSE(14, ClusterEventDissector::dissectAppendCloseSession),
+
+    /**
+     * Event when the background replicator state changes.
+     */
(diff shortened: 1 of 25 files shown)


vachb:
thanks for this Mike, perfect timing! 
quick question : if we assume a service is taking big snapshots (lets say 10GB) when a snapshot is taken and it starts to replicate is it possible that it will saturate the network bandwidth? is there some sort of priority for snapshots so that they do not aggressively use bandwidth, for example if network is saturated one would prefer the snapshot to have much less priority than the messages that are getting replicated around...

mikeb01:
> if we assume a service is taking big snapshots (lets say 10GB) when a snapshot is taken and it starts to replicate is it possible that it will saturate the network bandwidth?

These are sorts of questions we have to work through and one of the reasons why this needs further understanding and review.  Options for this specific case could include using congestion control on the topic used for replicating the snapshots.

There are other concerns as well, like the effects on flow control while the follower is replicating.

dave5595:
@mikeb01 Hi Mike, out of curiosity, why was this PR not merged? Is this feature only available on Aeron Premium?

mjpt777:
@dave5595 Background snapshots are one of the features available in Aeron premium. There are numerous issues with doing background snapshots in the OSS version and a better implementation was possible using our standby nodes in Aeron Premium.

PR #1282, opened 2022-01-13
[Java] Modified ClusterClock and ConsensusModuleAgent to better support time travel tests.

kwbc:
ConsensusModuleAgent generally uses ClusterClock.time() for log timestamps and ClusterClock.timeNanos() for internal time-related tasks.
During time travel tests we want adjust/offset the log timestamps only.
Modified ConsensusModuleAgent to use ClusterClock.convertToNanos instead of timeUnit.convert to allow the implementations of ClusterClock to differentiate between timestamps to be used in the log versus internal time-keeping.
Also modified Election to take both the log timestamp and nano time in several methods to avoid converting back from nanoseconds to log time.

Final note: debated about what to call the log-timestamp parameter. Went with 'timestamp', but perhaps 'now' or 'nowLog' would be better?

### DIFF
--- a/aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModuleAgent.java
+++ b/aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModuleAgent.java
@@ -54,6 +54,7 @@ import static io.aeron.cluster.client.AeronCluster.SESSION_HEADER_LENGTH;
 import static io.aeron.cluster.service.ClusteredServiceContainer.Configuration.MARK_FILE_UPDATE_INTERVAL_NS;
 import static io.aeron.exceptions.AeronException.Category.WARN;
 import static java.lang.Math.min;
+import static java.util.concurrent.TimeUnit.NANOSECONDS;
 
 final class ConsensusModuleAgent implements Agent, TimerService.TimerHandler
 {
@@ -304,7 +305,8 @@ final class ConsensusModuleAgent implements Agent, TimerService.TimerHandler
                 ctx,
                 this);
 
-            election.doWork(clusterClock.timeNanos());
+            final long now = clusterClock.time();
+            election.doWork(now, clusterClock.convertToNanos(now));
             state(ConsensusModule.State.ACTIVE);
         }
 
@@ -317,7 +319,7 @@ final class ConsensusModuleAgent implements Agent, TimerService.TimerHandler
     public int doWork()
     {
         final long now = clusterClock.time();
-        final long nowNs = clusterTimeUnit.toNanos(now);
+        final long nowNs = clusterClock.convertToNanos(now);
         int workCount = 0;
 
         try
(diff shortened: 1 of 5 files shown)


aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModuleAgent.java
```diff
@@ -560,10 +562,10 @@ void onChallengeResponse(final long correlationId, final long clusterSessionId,
 
                 if (session.id() == clusterSessionId && session.state() == CHALLENGED)
                 {
-                    final long now = clusterClock.time();
-                    final long nowMs = clusterTimeUnit.toMillis(now);
-                    session.lastActivityNs(clusterTimeUnit.toNanos(now), correlationId);
-                    authenticator.onChallengeResponse(clusterSessionId, encodedCredentials, nowMs);
+                    final long nowNs = clusterClock.timeNanos();
+                    final long nowMillis = NANOSECONDS.toMillis(nowNs);
```
mikeb01, line 566: Should be `nowMs`.

aeron-cluster/src/test/java/io/aeron/cluster/ElectionTest.java
```diff
@@ -1294,4 +1294,9 @@ private Election newElection(
 
         return clusterMembers;
     }
+
+    private static void electionDoWork(final Election election, final long time)
```
mikeb01, line 1298: Should probably be `timeNs` for consistency.

mikeb01:
Overall I'm happy with the change (some small comments mentioned).

One question the addition of `ClusterClock::convertToNanos`.  I'm not sure I completely understand the necessity of it, would it be possible to show with a simple test/use case it's value over calling `clusterClock.timeUnit().toNanos()` directly?

> Final note: debated about what to call the log-timestamp parameter. Went with 'timestamp', but perhaps 'now' or 'nowLog' would be better?

I think `timestamp` is a bit generic.  Not sure what the best name should be (`nowClusterUnits` is a bit wordy, `logTimestamp` is another option).

aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModuleAgent.java
kwbc: I had initially named it `nowMs` but thought it looked too much like `nowNs`.  Will change it back to nowMs if that is preferred.

kwbc:
> One question the addition of `ClusterClock::convertToNanos`. I'm not sure I completely understand the necessity of it, would it be possible to show with a simple test/use case it's value over calling `clusterClock.timeUnit().toNanos()` directly?

In our time travel implementation of ClusterClock, `time()` returns `epochNanos() + timeTravelOffset`, and `convertToNanos(time)` returns `time - timeTravelOffset` thus converting the adjusted log timestamp back to real wall-clock time for internal time keeping purposes within the consensus module.  Does that make sense?

Would be simpler/clearer if consensus module used ClusterClock for log timestamps (and timers) only, and some other EpochClock for everything else.  But that would impose a performance penalty having to call ClusterClock.time and EpochClock.time in the same duty-cycle.

kwbc:
> I think `timestamp` is a bit generic. Not sure what the best name should be (`nowClusterUnits` is a bit wordy, `logTimestamp` is another option).

I like `logTimestamp`.  `logClusterUnit` is a bit wordy, but more consistent with `nowNs` and only slightly longer.  Anyone else have an opinion?

tmontgomery (approved):
`logTimestamp` is fine with me. I also am OK with the PR.

kwbc:
Thanks for the code review comments.  I've renamed timestamp parameters to logTimestamp in ConsensusModuleAgent and Election -- like how it is consistent with logPosition, etc.  And other requested parameter renames.

mjpt777:
I'll review this Friday or over the weekend.

PR #1285, opened 2022-01-27
Allow configuration number of buffers for sendmmsg and recvmmsg

mikeb01:


aeron-driver/src/main/c/aeron_driver_context.h
```diff
@@ -129,7 +129,9 @@ typedef struct aeron_driver_context_stct
     int32_t publication_reserved_session_id_low;            /* aeron.publication.reserved.session.id.low = -1 */
     int32_t publication_reserved_session_id_high;           /* aeron.publication.reserved.session.id.high = 1000 */
     uint8_t multicast_ttl;                                  /* aeron.socket.multicast.ttl = 0 */
-
+    uint32_t receiver_num_buffers;                          /* aeron.receiver.num.buffers = 2 */
```
mjpt777, line 132: A better name might be `aeron.receiver.io.vector.length`. Num buffers is too vague.

PR #1289, opened 2022-02-09
Use connect and send for handling single packets to improve latency

mikeb01:
- The sender_channel_endpoint will connect during transport initialisation
- When using multicast a second socket will be opened so that control traffic can still be received (like the Java implementation).
- Use sendmmsg if the socket isn't connected or there is more than one message to send


cppbuild/cppbuild.cmd
```diff
@@ -49,6 +49,13 @@ if not "%1"=="" (
         set "EXTRA_CMAKE_ARGS=!EXTRA_CMAKE_ARGS! -DGRADLE_WRAPPER=%2"
         echo "Setting -DGRADLE_WRAPPER=%2"
         shift
+    ) else if "%1"=="--parallel" (
```
vyazelenko, line 52: I've pushed a change on HEAD to run Windows build in parallel by default, i.e. it sets `CMAKE_BUILD_PARALLEL_LEVEL=%NUMBER_OF_PROCESSORS%` and supports a `--no-parallel` flag to run on a single CPU like the Linux build does.

aeron-test-support/src/main/java/io/aeron/test/MediaDriverTestUtil.java
```diff
@@ -99,8 +99,8 @@ private void deleteFiles()
         outputFilesByAeronDirectoryName.forEach(
             (aeronDirectoryName, files) ->
             {
-                IoUtil.delete(files.stdout, false);
-                IoUtil.delete(files.stderr, false);
+                IoUtil.delete(files.stdout, true);
+                IoUtil.delete(files.stderr, true);
```
vyazelenko, line 103: Why the change to throw on error? Is it because if one fails the other will fail as well, i.e. due to a running process?

vyazelenko (approved):

aeron-test-support/src/main/java/io/aeron/test/MediaDriverTestUtil.java
mikeb01: The change is the inverse, using `true` will ignore failures on delete.

cppbuild/cppbuild.cmd
mikeb01: It's also useful to be able to specify a number of CPUs.  I've found running parallel, but leaving a couple of CPUs free lets the IDE and desktop remain responsive during builds.

mikeb01: I've rebased with your changes and added `--parallel-cpus`.

aeron-test-support/src/main/java/io/aeron/test/MediaDriverTestUtil.java
vyazelenko: Got it.

aeron-driver/src/main/c/media/aeron_udp_channel_transport.c
```diff
@@ -275,10 +290,17 @@ int aeron_udp_channel_transport_init(
 
     if (aeron_set_socket_non_blocking(transport->fd) < 0)
     {
-        AERON_APPEND_ERR("%", "");
+        AERON_APPEND_ERR("%", "failed to set transport->fd to be non-blocking");
```
mjpt777, line 293: Should this be `transport->recv_fd`?

aeron-driver/src/main/c/media/aeron_udp_channel_transport.c
```diff
@@ -407,69 +459,176 @@ int aeron_udp_channel_transport_recvmmsg(
 #endif
 }
 
-int aeron_udp_channel_transport_sendmmsg(
-    aeron_udp_channel_data_paths_t *data_paths,
+static int aeron_udp_channel_transport_send_connected(
+        aeron_udp_channel_transport_t *transport,
```
mjpt777, line 463: double indent.

mjpt777:
Other than the minor issues this LGTM.

aeron-driver/src/main/c/media/aeron_udp_channel_transport.c
mikeb01: That looks okay to me.  It sets non-blocking and reports errors on recv_fd a few lines down from that code.

.github/workflows/ci.yml
```diff
@@ -539,7 +539,7 @@ jobs:
           7z x cmake-$Env:CMAKE_VERSION-win64-x64.zip
           echo "cmake-$Env:CMAKE_VERSION-win64-x64/bin" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8
       - name: Build
-        run: cmake --version && cppbuild/cppbuild
+        run: cmake --version && cppbuild/cppbuild --parallel
```
vyazelenko, line 542: There is no `--parallel` option anymore and thus the script will fail with an error.

cppbuild/cppbuild.cmd
```diff
@@ -57,6 +57,10 @@ if not "%1"=="" (
     ) else if "%1"=="--no-parallel" (
         set "CMAKE_BUILD_PARALLEL_LEVEL=1"
         echo "Disabling parallel build"
+    ) else if "%1"=="--parallel-cpus" (
```
vyazelenko, line 60: I think we shall align Linux and Windows scripts. Currently Linux script defines `--spare-cpu` option which simply leaves one core unused whereas this script adds `--parallel-cpus` option which allows specifying the exact number of cores.

mikeb01: I've changed the bash script to use `--parallel-cpus` rather than the other way round.

PR #1290, opened 2022-02-21
[Java] Cluster interface config.

vyazelenko:
This PR makes it possible to configure `interface` params for _all_ of the Cluster channels. To that end it:
- Replaces all of the ad hoc channels used for defining destinations with the call to `ChannelUri#createDestinationUri`.
- Adds two new configuration properties `aeron.cluster.follower.catchup.channel` and `aeron.cluster.leader.archive.control.channel`. The former is used to create the replay channel when doing a catch up on the follower node. And the latter is used for creating a control request channel to the leader archive.
- Adds an `egressChannel` configuration option to the `ConsensusModule` which is using the same configuration property as the `AeronCluster` (i.e. `aeron.cluster.egress.channel`). When the `egressChannel` is not defined then response channel from the leader to the client is based on the `responseChannel` sent by the client. However if it defined then it is used a basis for the response channel, i.e. if the client `responseChannel` contains an `endpoint` parameter then it will be used to set/substitute the value in the `egressChannel` of the leader node and otherwise leader's `egressChannel` is used as is.

### DIFF
--- a/aeron-client/src/main/java/io/aeron/ChannelUri.java
+++ b/aeron-client/src/main/java/io/aeron/ChannelUri.java
@@ -20,7 +20,10 @@ import org.agrona.AsciiEncoding;
 import org.agrona.collections.ArrayUtil;
 import org.agrona.collections.Object2ObjectHashMap;
 
-import java.util.*;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
 
 import static io.aeron.CommonContext.*;
 import static io.aeron.logbuffer.FrameDescriptor.FRAME_ALIGNMENT;
@@ -36,6 +39,7 @@ import static io.aeron.logbuffer.FrameDescriptor.FRAME_ALIGNMENT;
  * </pre>
  * <p>
  * Multiple params with the same key are allowed, the last value specified takes precedence.
+ *
  * @see ChannelUriStringBuilder
  */
 public final class ChannelUri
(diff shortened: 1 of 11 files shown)


aeron-client/src/main/java/io/aeron/ChannelUri.java
```diff
@@ -507,6 +511,28 @@ public static long getTag(final String paramValue)
...
+     * Create a channel URI for a destination, i.e. a channel that uses {@code media} and {@code interface} parameters
+     * of the original channel and adds specified {@code endpoint} to it. For example given the input channel is
+     * {@code aeron:udp?mtu=1440|ttl=0|endpoint=localhost:8090|term-length=128k|interface=eth0} and the endpoint is
+     * {@code 192.168.0.14} the output of this method will be {@code aeron:udp?endpoint=192.168.0.14|interface=eth0}.
+     *
+     * @param channel  for which the destination is being added.
+     * @param endpoint for the target destination.
+     * @return new channel URI for a destination.
+     */
+    public static String createDestinationUri(final String channel, final String endpoint)
```
mikeb01, line 524: Would this code be cleaner if we used the ChannelUriStringBuilder?

vyazelenko: Yes, it would but that would mean having a circular dependency between `ChannelUriStringBuilder` and `ChannelUri`. If `createDestinationUri` is moved elsewhere (maybe even to the `ChannelUriStringBuilder`) we could use the `ChannelUriStringBuilder`.

mikeb01: Fair point I missed that.

mikeb01:
Is it possible to write a system test showing the interface configuration?  Perhaps this could fall into the cluster topology tests?

PR #1291, opened 2022-02-21
fix build on ARM with clang

alextibbles:
Add -latomic as a link dep on aeron_client & aeron_driver. This is required when building on ARM with clang.
Eg. cmake -DCMAKE_C_COMPILER=clang-14 -DCMAKE_CXX_COMPILER=clang++-14 ../..


tmontgomery:
I am fine with this. @mjpt777 @vyazelenko and @mikeb01 what do you think?

tmontgomery (approved):

vyazelenko (approved):

PR #1294, opened 2022-02-24
Ensure catchup times out with lack of progress

mikeb01:
Separate timestamps of AppendPosition into send and update, use send to track resends and update to track/timeout progress.


aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModuleAgent.java
```diff
@@ -754,6 +755,10 @@ void onCatchupPosition(
                     logRecordingId, logPosition, Long.MAX_VALUE, channel, ctx.logStreamId()));
                 follower.catchupReplayCorrelationId(archive.lastCorrelationId());
             }
+            else if (null != follower)
+            {
+                System.out.println("Existing catchup already running: " + follower);
```
vyazelenko, line 760: Do we really want a `System.out` here? Could be this a `ClusterEvent` instead?

mikeb01: No we don't, I'll remove that.

PR #1296, opened 2022-03-01
[C] media driver - added support for configuring cpu affinity

kwbc:
On conductor, receiver and sender threads.


mikeb01:
I've been thinking about this one a bit and trying to decide whether thread affinity should be the responsibility of the driver library or the main function that starts it.  By decoupling it from the library/context it makes it easier for systems that manage affinity differently and require their own main to more easily plug in their own options.  DPDK is an example of this.  It might make sense for the context to still carry the configuration options.  I'll put together an alternate PR so we can contrast them.

tmontgomery:
@mikeb01 That is an excellent point. I'd like that PR and I agree with your points on it.

mikeb01:
We've decided to take a slightly different approach which we merged earlier today. However, it is functionally equivalent and uses that same config.

PR #1298, opened 2022-03-02
[C] Add thread affinity support for the C media driver.  

mikeb01:
- Put thread affinity configuration on the driver context and add configuration variables.
- Manage configuration of thread affinity from aeronmd.c
- Add script to show a process and threads CPU affinity (linux only).

```
$ ./aeron-samples/scripts/show_thread_affinity.sh 
PID: 15596 (aeronmd)
pid 15596's current affinity list: 2 (aeronmd)
pid 15597's current affinity list: 3 (sender)
pid 15598's current affinity list: 5 (receiver)
```


tmontgomery (approved):
I like this better. I'm good with this if it passes CI.

PR #1299, opened 2022-03-10
Set a flag on status messages when the group tag is present.

mikeb01:
Make it clearer when the group tag is present.

### DIFF
--- a/aeron-client/src/main/c/protocol/aeron_udp_protocol.c
+++ b/aeron-client/src/main/c/protocol/aeron_udp_protocol.c
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+#include <stdbool.h>
 #include <string.h>
 
 #include "util/aeron_bitutil.h"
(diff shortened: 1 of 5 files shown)


aeron-client/src/main/c/protocol/aeron_udp_protocol.c
```diff
@@ -25,18 +26,22 @@ int aeron_udp_protocol_group_tag(aeron_status_message_header_t *sm, int64_t *gro
         offsetof(aeron_status_message_optional_header_t, group_tag);
     const size_t group_tag_size = sizeof(*group_tag);
     const size_t frame_length_with_group_tag = group_tag_offset + group_tag_size;
+    const bool has_group_tag =
+        (sm->frame_header.flags & AERON_STATUS_MESSAGE_HEADER_GROUP_TAG_FLAG &&
+        sm->frame_header.frame_length >= (int32_t)frame_length_with_group_tag) || // ATS frames
+        sm->frame_header.frame_length == (int32_t)frame_length_with_group_tag;    // Compatibility with existing behaviour
 
-    if (sm->frame_header.frame_length == (int32_t)frame_length_with_group_tag)
+    if (has_group_tag && sm->frame_header.frame_length >= (int32_t)frame_length_with_group_tag)
```
vyazelenko, line 34: The length check is redundant as it is performed when computing the `has_group_tag` variable.

mikeb01: This was there for backward compatibility with implementations that weren't setting the flag.  However, @tmontgomery and I have discussed an alternative that means this PR is redundent.

PR #1300, opened 2022-03-11
Split catchupPosition into catchupJoinPosition and catchupCommitPosition

mikeb01:
Split catchupPosition into catchupJoinPosition - based on logPosition from new leadership term and catchCommitPosition - based on commit position messages received from the leader during catchup.  Use catchupCommitPosition as the limitPosition for processing messages from the LogAdapter during catchup.  Use the max of the two to determine that the follower is close enough to join the live stream and leave the catchup state.

### DIFF
--- a/aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModuleAgent.java
+++ b/aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModuleAgent.java
@@ -1771,14 +1771,10 @@ final class ConsensusModuleAgent implements Agent, TimerService.TimerHandler
                 throw new ClusterEvent(
                     "unexpected image close during catchup: position=" + logAdapter.image().position());
             }
-        }
 
-        final long appendPosition = logAdapter.position();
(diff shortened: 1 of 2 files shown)


aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModuleAgent.java
```diff
@@ -1771,14 +1771,10 @@ int catchupPoll(final long limitPosition, final long nowNs)
...
             }
-        }
 
-        final long appendPosition = logAdapter.position();
-        if (appendPosition > lastAppendPosition || nowNs > (timeOfLastAppendPositionSendNs + leaderHeartbeatIntervalNs))
-        {
-            commitPosition.proposeMaxOrdered(appendPosition);
             final ExclusivePublication publication = election.leader().publication();
-            tryUpdateAppendPosition(publication, nowNs, replayLeadershipTermId, appendPosition);
+            tryUpdateAppendPosition(publication, nowNs, replayLeadershipTermId, appendPosition.get());
```
mjpt777, line 1776: If no progress then we should only pulse out append position on an interval otherwise we could fill the window and prevent other messages getting out in a timely manner which may be more important.

PR #1302, opened 2022-03-18
[Java] Add `trace` alias to all logging contexts and change the meaning of `all`.

vyazelenko:
- Adds `trace` as an alias for all logging events.
- Changes `all` for the driver to mean the same as `admin`.
- Changes `all` for the cluster to mean all events excluding the `APPEND_POSITION`, `COMMIT_POSITION` and `CATCHUP_POSITION`.


mikeb01:
Is there a name that we can come up with other than "admin" to represent this group of events?

vyazelenko:
@mikeb01 I agree that the name is bad. Perhaps something like `no-position-events` or `position-less`?
I first implemented the opposite approach, i.e. I added the new `trace` event which included everything and had `all` be `trace - position_events`. But it felt weird as `all` means all existing events for driver and archive so I decided to go with the `admin` approach instead.

vyazelenko:
Renamed `admin` to `no-position-events`.

mikeb01:
I think we should alias "all" to this configuration and add a "trace" configuration into which we put everything including any new events.  We should do this for archive and driver too.  This means it shouldn't cause any problems with existing users in production and we can also let people know that trace will include lots of very verbose information and to be careful using it in production.

vyazelenko:
@mjpt777 I've considered this as well but it "breaks" if the end user specifies the disabled list explicitly.

For example our slow tests define:
```
aeron.event.cluster.log=all
aeron.event.cluster.log.disable=CANVASS_POSITION
```
Which means that the default value for `aeron.event.cluster.log.disable` will not apply. In this case this will be good (as we want to see all of the events in the tests) but it might be surprising to some users that might have defined `aeron.event.cluster.log.disable` in their env and will start getting the high volume events on top.

mikeb01:
> If we have an `all` group then it should have all events. How about we have the high volume event codes in the disabled list by default?

If we are unable to change `all` to include everything, including new, potentially high volume events as they are added to the logging, then we could effectively "deprecate" it.  By that I mean we still support it, but it only contains what is currently contains and we stop recommending its use.  We add `trace` with adequate warnings that it will log potentially lots of stuff and new events will be added to it.  Then we come up with some better groupings for convenience, e.g. `admin`, `election`, `position`.

PR #1303, opened 2022-03-21
[Java] Add flags to appendPosition message. 

mikeb01:
Set flag if appendPosition occurs during catchup.  Resend stopCatchup if catchup flag set on appendPosition even if replay has already been stopped.

### DIFF
--- a/aeron-agent/src/main/java/io/aeron/agent/ClusterEventDissector.java
+++ b/aeron-agent/src/main/java/io/aeron/agent/ClusterEventDissector.java
@@ -15,11 +15,11 @@
  */
 package io.aeron.agent;
 
+import io.aeron.protocol.HeaderFlyweight;
 import org.agrona.MutableDirectBuffer;
 
 import static io.aeron.agent.ClusterEventCode.ELECTION_STATE_CHANGE;
 import static io.aeron.agent.ClusterEventCode.NEW_LEADERSHIP_TERM;
-import static io.aeron.agent.CommonEventDissector.appendFlagsAsChars;
 import static io.aeron.agent.CommonEventDissector.dissectLogHeader;
 import static java.nio.ByteOrder.LITTLE_ENDIAN;
 import static org.agrona.BitUtil.SIZE_OF_INT;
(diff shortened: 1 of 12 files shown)


aeron-agent/src/main/java/io/aeron/agent/ClusterEventDissector.java
```diff
@@ -336,10 +336,13 @@ public static void dissectAppendPosition(
         final long logPosition = buffer.getLong(absoluteOffset, LITTLE_ENDIAN);
         absoluteOffset += SIZE_OF_LONG;
         final int followerMemberId = buffer.getInt(absoluteOffset, LITTLE_ENDIAN);
+        absoluteOffset += SIZE_OF_INT;
+        final int flags = buffer.getInt(absoluteOffset, LITTLE_ENDIAN);
 
         builder.append(": leadershipTermId=").append(leadershipTermId);
         builder.append(" logPosition=").append(logPosition);
         builder.append(" followerMemberId=").append(followerMemberId);
+        builder.append(" flags=0b").append(Integer.toBinaryString(flags));
```
mjpt777, line 345: It would be more efficient to use something like `HeaderFlyweight.appendFlagsAsChars`.

mjpt777:
On the whole I think this is a good approach.

mikeb01:
I'm wondering if the flags field should be smaller?  We've aimed to keep the appendPosition message <= 24 bytes.  This uses the last 4 bytes.  I could change it to be a byte instead that would allow for repurposing the remaining 3 bytes for other things as we may need them?

aeron-agent/src/main/java/io/aeron/agent/CommonEventDissector.java
```diff
@@ -144,4 +146,23 @@ else if (16 == addressLength)
 
         return encodedLength;
     }
+
+    static void appendFlagsAsChars(final short flags, final Appendable appendable)
```
vyazelenko, line 150: You can simply use `io.aeron.protocol.HeaderFlyweight#appendFlagsAsChars` method. Agent already pulls all of the decoders for the dissectors.

PR #1310, opened 2022-04-06
[C] Remove loop through destinations nested within receive_channel_endpoint_send

mikeb01:
Looping is already covered at a higher point in the call stack (generally by connection per image).  Use specific sends for setup eliciting SM so that only the correct destination is targeted.

### DIFF
--- a/aeron-driver/src/main/c/aeron_driver_receiver.c
+++ b/aeron-driver/src/main/c/aeron_driver_receiver.c
@@ -227,8 +227,10 @@ int aeron_driver_receiver_do_work(void *clientd)
             }
             else if (aeron_receive_channel_endpoint_should_elicit_setup_message(entry->endpoint))
             {
+                // TODO (MB): Check pending entries control_addr is populated correctly...
                 if (aeron_receive_channel_endpoint_send_sm(
                     entry->endpoint,
+                    entry->destination,
                     &entry->control_addr,
                     entry->stream_id,
                     entry->session_id,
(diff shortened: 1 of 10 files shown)


aeron-driver/src/main/c/aeron_driver_receiver.c
```diff
@@ -415,6 +417,8 @@ void aeron_driver_receiver_on_add_destination(void *clientd, void *item)
 
     if (destination->conductor_fields.udp_channel->has_explicit_control)
     {
+        printf("Adding destination: %s\n", destination->conductor_fields.udp_channel->original_uri);
```
mjpt777, line 420: Debug left in.

aeron-driver/src/test/c/aeron_c_multi_destination_test.cpp
```diff
@@ -0,0 +1,167 @@
...
+    ASSERT_TRUE(pub1) << aeron_errmsg();
+
+    std::int32_t initialTermId;
+    std::int32_t termId;
+    std::int32_t termOffset;
+    std::int32_t sessionId;
+    mdsParameters(pub1, &initialTermId, &termId, &termOffset, &sessionId);
+
+    auto pubUri2 = std::string(PUB_URI_2)
+        .append("|init-term-id=" + std::to_string(initialTermId))
```
mjpt777, line 89: More efficient to chain the appends as follows:
``` .append("|init-term-id=").append(std::to_string(initialTermId))```

PR #1325, opened 2022-05-24
[Cpp] Added tryClaim method to concurrent ringbuffers.

im95able:
In the codebase my team is working on we would like to use `concurrent::ringbuffers`. For us, `tryClaim` interface is preferred. I tried following you're code convetion as much as possible but reused the `tryClaim` implementaion for `write` methods(that way tryClaim also gets tested with `write`).
Unfortunately order of atomic writes for `oneToOneRingBuffer` was changed. `Tail` gets written first and only after does `messageHeader` get written. This could affect performance. 
In our implementation of `spsc` channel we provide the callback to the `tryClaim` method which gets called with the claimed buffer. That way `tail` gets written after the `messageHeader`; wanting to stay within the convention of your codebase this PR does it by introducing `RingBufferClaim` class.

Edit: As per @mjpt777's suggestion refactored the code to reflect Java `RingBuffer`s.


mjpt777:
It would be best if the implementations followed the same API semantics as we have for the Java `RingBuffer`s.

https://github.com/real-logic/agrona/tree/master/agrona/src/main/java/org/agrona/concurrent/ringbuffer

aeron-client/src/test/cpp/concurrent/OneToOneRingBufferTest.cpp
```diff
@@ -383,6 +383,63 @@ TEST_F(OneToOneRingBufferTest, shouldCopeWithExceptionFromHandler)
...
+    util::index_t length = 100;
+    util::index_t head = 0;
+    util::index_t tail =
+        head + (CAPACITY - util::BitUtil::align(length - RecordDescriptor::ALIGNMENT, RecordDescriptor::ALIGNMENT));
+
+    m_ab.putInt64(HEAD_COUNTER_INDEX, head);
+    m_ab.putInt64(TAIL_COUNTER_INDEX, tail);
+
+    EXPECT_EQ(OneToOneRingBuffer::INSUFFICIENT_CAPACITY, m_ringBuffer.tryClaim(MSG_TYPE_ID, length));
+    EXPECT_EQ(m_ab.getInt32(RecordDescriptor::lengthOffset(0)), -recordLength);
```
mikeb01, line 438: `recordLength` and `alignedRecordLength` do not appear to be define in this method.  Causing compile errors in some of the CI builds.

im95able:
> It would be best if the implementations followed the same API semantics as we have for the Java `RingBuffer`s.
> 
> https://github.com/real-logic/agrona/tree/master/agrona/src/main/java/org/agrona/concurrent/ringbuffer

Will do

aeron-client/src/test/cpp/concurrent/OneToOneRingBufferTest.cpp
im95able: Refactored both the implementation and tests

im95able:
@mjpt777 could there be a review of this ?

vyazelenko (changes requested):
There are two things that stand out:
- Usage of the `RecordDescriptor::makeHeader` to combine `length` and `type` fields together.
  We have decided that it would be better if the C++ version was aligned with the Java and C implementations, i.e. it should read/write `length` and `type` as separate fields.
- `OneToOneRingBuffer::claimCapacity` is not aligned with the Java implementation, i.e. in [Java](https://github.com/real-logic/agrona/blob/ace2638711c72780e8a714f5aba49cf7c270f13a/agrona/src/main/java/org/agrona/concurrent/ringbuffer/OneToOneRingBuffer.java#L412-L413) and in C we add an additional `HEADER_LENGTH` when computing the `requiredCapacity`. This allows for pre-zeroing of the next header upon claim and thus we don't need to zero the memory in the `read/controlledRead` methods.

Another thing to note is that the C++ and C implementations don't have the `controlledRead` API.

im95able:
> There are two things that stand out:
> 
> * Usage of the `RecordDescriptor::makeHeader` to combine `length` and `type` fields together.
>   We have decided that it would be better if the C++ version was aligned with the Java and C implementations, i.e. it should read/write `length` and `type` as separate fields.
> * `OneToOneRingBuffer::claimCapacity` is not aligned with the Java implementation, i.e. in [Java](https://github.com/real-logic/agrona/blob/ace2638711c72780e8a714f5aba49cf7c270f13a/agrona/src/main/java/org/agrona/concurrent/ringbuffer/OneToOneRingBuffer.java#L412-L413) and in C we add an additional `HEADER_LENGTH` when computing the `requiredCapacity`. This allows for pre-zeroing of the next header upon claim and thus we don't need to zero the memory in the `read/controlledRead` methods.
> 
> Another thing to note is that the C++ and C implementations don't have the `controlledRead` API.

Will fix this probably some time next week. Thanks for the input.

mjpt777:
Is this progressing?

PR #1328, opened 2022-06-07
Prevent timeout warnings with dynamic nodes and log replication

mikeb01:
Remove replicationCommitPosition which was used to verify a leader had committed a position equal or higher than the log replicated position. This is not required as replication should only proceed to the end of previous term which will already be committed. It allows dynamic nodes to work more cleanly with log replication as they don't receive commit position messages until after catch up and they have offically joined the cluster.

### DIFF
--- a/aeron-cluster/src/main/java/io/aeron/cluster/Election.java
+++ b/aeron-cluster/src/main/java/io/aeron/cluster/Election.java
@@ -1328,6 +1328,31 @@ class Election
         final long nowNs)
     {
         final long recordingId = consensusModuleAgent.logRecordingId();
+        final long initialTermBaseLogPosition = this.initialTermBaseLogPosition;
+        final long initialLogLeadershipTermId = this.initialLogLeadershipTermId;
+        final ConsensusModule.Context ctx = this.ctx;
+
+        ensureRecordingLogCoherent(
+            ctx,
+            recordingId,
+            initialLogLeadershipTermId,
+            initialTermBaseLogPosition,
+            leadershipTermId,
+            logTermBasePosition,
+            logPosition,
+            nowNs);
+    }
+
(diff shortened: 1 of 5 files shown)


aeron-cluster/src/main/java/io/aeron/cluster/Election.java
```diff
@@ -526,11 +519,10 @@ void onCommitPosition(final long leadershipTermId, final long logPosition, final
         {
             catchupCommitPosition = max(catchupCommitPosition, logPosition);
         }
-        else if (FOLLOWER_LOG_REPLICATION == state && leaderMemberId == leaderMember.id())
-        {
-            replicationCommitPosition = max(replicationCommitPosition, logPosition);
-            replicationDeadlineNs = ctx.clusterClock().timeNanos() + ctx.leaderHeartbeatTimeoutNs();
-        }
+//        else if (FOLLOWER_LOG_REPLICATION == state && leaderMemberId == leaderMember.id())
```
mjpt777, line 522: Why leave commented code?

aeron-cluster/src/main/java/io/aeron/cluster/Election.java
```diff
@@ -1328,6 +1328,31 @@ private void ensureRecordingLogCoherent(
         final long nowNs)
     {
         final long recordingId = consensusModuleAgent.logRecordingId();
+        final long initialTermBaseLogPosition = this.initialTermBaseLogPosition;
```
mjpt777, line 1331: Why create local vars here?

mikeb01: That was part of a extract method refactor and I forgot to inline the fields back into the method call.

PR #1332, opened 2022-06-21
Allow for replication and replay requests to specify the file I/O max length

mikeb01:
This can give a degree of control around priority of replay operations to prevent them from swamping a latency sensitive archive.

To make the addition of this parameter smoother, this PR introduces a different API for replay and replication requests.  A fluent Params object is supplied to the replay and replication request methods (including corresponding methods on the proxy).  This allows for a more extensible API to prevent more overloads being added to the AeronArchive client.

Includes checking the I/O operation length against the MTU.  Values less than the MTU length will cause issues.


mjpt777:
The approach of adding a params object is nice and good for extension.

The minor version should be increased to support the API extension.  https://github.com/real-logic/aeron/blob/master/aeron-archive/src/main/java/io/aeron/archive/client/AeronArchive.java#L2325

mjpt777 (approved):
LGTM

PR #1350, opened 2022-08-31
egress open message should be published before container messages

vdaniloff:
Right now it is possible first message on session open is sent to cluster, then it is responded within service, and only then cluster sends egress event message. In this case the message sent from service is lost, because AeronCluster will poll everything before the egress event.


vdaniloff:
well, I see this will not help, changing

PR #1355, opened 2022-09-06
prevent hanging when up to 0 fragments in ArchiveTool

vdaniloff:


vyazelenko:
@vdaniloff I changed how the `ArchiveTool#dump` method works in the https://github.com/real-logic/aeron/commit/6ea27ba5d7b54435141de95f0ed5740f662c433a. It is no longer possible to specify the  `fragmentCountLimit` that is negative or zero.

Was the problem you were trying to solve related to the zero limit or did you see an infinite loop with the non-empty archive recording when the `fragmentCountLimit > 0`?

vdaniloff:
no, this was with 0 passed as a limit, so I think this is resolved now, thanks

PR #1357, opened 2022-09-09
[Java] Fix pending service messages snapshot state.

vyazelenko:
This PR adds a new tool `ConsensusModuleSnapshotPendingServiceMessagesPatch` that can fix consensus module snapshot state if the divergence in pending service messages is detected.

### DIFF
--- a/aeron-samples/src/main/java/io/aeron/cluster/ConsensusModuleSnapshotPendingServiceMessagesPatch.java
+++ b/aeron-samples/src/main/java/io/aeron/cluster/ConsensusModuleSnapshotPendingServiceMessagesPatch.java
@@ -525,7 +525,10 @@ public class ConsensusModuleSnapshotPendingServiceMessagesPatch
             final long position,
             final RecordingSignal signal)
         {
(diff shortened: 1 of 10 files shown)


aeron-samples/src/main/java/io/aeron/cluster/ConsensusModuleSnapshotPendingServiceMessagesPatch.java
```diff
@@ -0,0 +1,570 @@
...
+
+        public void onSignal(
+            final long controlSessionId,
+            final long correlationId,
+            final long recordingId,
+            final long subscriptionId,
+            final long position,
+            final RecordingSignal signal)
+        {
+            this.recordingId = recordingId;
```
mjpt777, line 528: Be careful of overwriting existing capture `recordingId` as some signals have `-1` when not is relevant.

vyazelenko: Fixed.

PR #1359, opened 2022-09-13
[Java] Allow `replicate` to overwrite an empty recording.

vyazelenko:
This PR changes how the `replicate` Archive API will treat an empty destination recording (i.e. whose `startPosition == stopPosition`). In such a case the source recording will completely overwrite such recording.

### DIFF
--- a/aeron-archive/src/main/java/io/aeron/archive/Catalog.java
+++ b/aeron-archive/src/main/java/io/aeron/archive/Catalog.java
@@ -687,39 +687,21 @@ final class Catalog implements AutoCloseable
         {
             length = checksumLength = newFrameLength - DESCRIPTOR_HEADER_LENGTH;
 
-            final int numExtraBytes = newFrameLength - oldFrameLength;
+            final int shiftBytes = newFrameLength - oldFrameLength;
             final int endOfLastRecording = nextRecordingDescriptorOffset;
-            if (endOfLastRecording + numExtraBytes > capacity)
+            if (endOfLastRecording + shiftBytes > capacity)
             {
-                growCatalog(MAX_CATALOG_LENGTH, numExtraBytes);
+                growCatalog(MAX_CATALOG_LENGTH, shiftBytes);
             }
 
-            nextRecordingDescriptorOffset += numExtraBytes;
+            nextRecordingDescriptorOffset += shiftBytes;
 
             final long[] index = catalogIndex.index();
             final int lastPosition = catalogIndex.lastPosition();
             if (recordingId != index[lastPosition])
             {
-                //  if not the last recording then shift existing data to the end of the file
-                fieldAccessBuffer.putBytes(
-                    recordingOffset + newFrameLength,
-                    fieldAccessBuffer,
-                    recordingOffset + oldFrameLength,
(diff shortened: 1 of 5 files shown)


aeron-archive/src/test/java/io/aeron/archive/CatalogTest.java
```diff
@@ -847,19 +931,362 @@ void extendRecordingShouldUpdateChecksum()
...
+                TERM_LENGTH * 4,
+                1372,
+                21,
+                8,
+                "A",
+                "B",
+                "C");
+            final int newFrameLength = 96;
+            assertChecksum(catalog, recordingOneId, newFrameLength, 1488471744, checksum);
+            // verify that the excessive old data was erased
```
mjpt777, line 1227: Best to pull out as a named method so use of comment is not needed.

aeron-archive/src/main/java/io/aeron/archive/Catalog.java
```diff
@@ -656,6 +656,108 @@ void extendRecording(
...
+            if (recordingId != index[lastPosition])
+            {
+                //  if not the last recording then shift existing data to the end of the file
+                fieldAccessBuffer.putBytes(
+                    recordingOffset + newFrameLength,
+                    fieldAccessBuffer,
+                    recordingOffset + oldFrameLength,
+                    endOfLastRecording - (recordingOffset + oldFrameLength));
+
+                // fixup the index of the shifted data
```
mjpt777, line 710: Avoid use of comments by extracting blocks to named methods.

mjpt777 (changes requested):
Replace comments with extracted methods.

aeron-archive/src/test/java/io/aeron/archive/CatalogTest.java
vyazelenko: Done.

aeron-archive/src/main/java/io/aeron/archive/Catalog.java
vyazelenko: Done.

vyazelenko:
Changes implemented. Note, I rebased the code on top of current `HEAD`.

aeron-archive/src/main/java/io/aeron/archive/ReplicationSession.java
```diff
@@ -57,6 +57,7 @@ class ReplicationSession implements Session, RecordingDescriptorConsumer
     private long srcStopPosition = NULL_POSITION;
     private long srcRecordingPosition = NULL_POSITION;
     private final long dstStopPosition;
+    private final boolean dstRecordingIsEmpty;
```
mjpt777, line 60: `isDestinationRecordingEmpty` is a better. Booleans starting with `isX` or `hasX` is common coding standard.

vyazelenko: Originally I added a method name `isDestinationRecordingEmpty` to the `RecordingSummary` and used in several places. But then the only consumer of it was the `ReplicationSession` so I left the poorly named variable instead. Fixed now.

PR #1364, opened 2022-09-27
stopReplay when backupAgent onclose

WorkingChen:


mjpt777:
Why does this need to be stopped? It will stop naturally if not complete when it is no longer connected?

WorkingChen:
> Why does this need to be stopped? It will stop naturally if not complete when it is no longer connected?

There is a problem restarting the service。
so I think it would be reasonable to stopReplay :)

image

aeron-cluster/src/main/java/io/aeron/cluster/ClusterBackupAgent.java
```diff
@@ -306,6 +310,19 @@ private void reset()
         CloseHelper.close(multiSnapshotReplication);
         multiSnapshotReplication = null;
 
+        if (NULL_VALUE != liveLogReplaySessionId)
+        {
+            try
+            {
+                clusterArchive.stopReplay(liveLogReplaySessionId);
```
mjpt777, line 317: This will not work. You need the full replay session id and not just the lower bits for the stream session.

mjpt777:
Better solution with this commit. https://github.com/real-logic/aeron/commit/5748fada7ad0a947b967b6cc175840b543d03c98

PR #1370, opened 2022-10-21
[Java/C] Improve handling of the out of disc space errors.

vyazelenko:
* Add `StorageSpaceException` that is thrown when there is not enough space to create a log buffer.
* Translate the above exception into the client visible error code `STORAGE_SPACE`.
* Update the C driver to return `AERON_ERROR_CODE_STORAGE_SPACE` if insufficient storage space when creating log buffers.
* [C] Fix storage space calculation on MacOS.
* [C] Add support for sparse files on Windows.
* [C] Allocate storage space for non-sparse files.
* [C] Recognize OS-specific out of disc space errors.
* [C] Add OS-specific error handling for the file system calls.

### DIFF
--- a/aeron-archive/src/main/java/io/aeron/archive/RecordingWriter.java
+++ b/aeron-archive/src/main/java/io/aeron/archive/RecordingWriter.java
@@ -239,7 +239,7 @@ final class RecordingWriter implements BlockHandler, AutoCloseable
 
         try
         {
-            isLowStorageSpace = StorageSpaceException.isOutOfDiscError(ex) ||
+            isLowStorageSpace = StorageSpaceException.isStorageSpaceError(ex) ||
                 ctx.archiveFileStore().getUsableSpace() < writeLength;
         }
         catch (final IOException ex2)
--- a/aeron-archive/src/main/java/io/aeron/archive/RecordingWriter.java
+++ b/aeron-archive/src/main/java/io/aeron/archive/RecordingWriter.java
@@ -18,11 +18,11 @@ package io.aeron.archive;
 import io.aeron.Image;
 import io.aeron.archive.checksum.Checksum;
 import io.aeron.archive.client.ArchiveException;
+import io.aeron.exceptions.StorageSpaceException;
 import io.aeron.logbuffer.BlockHandler;
 import org.agrona.CloseHelper;
 import org.agrona.DirectBuffer;
 import org.agrona.LangUtil;
-import org.agrona.Strings;
 import org.agrona.concurrent.CountedErrorHandler;
 import org.agrona.concurrent.UnsafeBuffer;
 
@@ -234,13 +234,12 @@ final class RecordingWriter implements BlockHandler, AutoCloseable
 
     private void checkErrorType(final IOException ex, final int writeLength)
     {
-        final String msg = ex.getMessage();
         boolean isLowStorageSpace = false;
         IOException suppressed = null;
 
         try
         {
-            isLowStorageSpace = (!Strings.isEmpty(msg) && msg.contains("No space left on device")) ||
+            isLowStorageSpace = StorageSpaceException.isOutOfDiscError(ex) ||
                 ctx.archiveFileStore().getUsableSpace() < writeLength;
         }
         catch (final IOException ex2)
@@ -249,7 +248,7 @@ final class RecordingWriter implements BlockHandler, AutoCloseable
         }
 
         final int errorCode = isLowStorageSpace ? ArchiveException.STORAGE_SPACE : ArchiveException.GENERIC;
-        final ArchiveException error = new ArchiveException("java.io.IOException - " + msg, ex, errorCode);
+        final ArchiveException error = new ArchiveException("java.io.IOException - " + ex.getMessage(), ex, errorCode);
 
         if (null != suppressed)
         {
--- a/aeron-client/src/main/c/util/aeron_fileutil.c
+++ b/aeron-client/src/main/c/util/aeron_fileutil.c
@@ -376,7 +376,7 @@ int aeron_create_file(const char *path, size_t length, bool sparse_file)
             AERON_SET_ERR(errno, "Failed to allocate file space: %s", path);
             goto error;
         }
-#elif defined(__APPLE__)
+#elif HAVE_F_PREALLOCATE
         fstore_t flags = {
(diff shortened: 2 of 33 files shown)


mikeb01:
It would be useful to have a system test, so that we can check that both the C driver and Java driver send the appropriate storage errors.  I think if we get a `AERON_ERROR_CODE_STORAGE_SPACE` I think it will come through as `AERON_ERROR_CODE_GENERIC_ERROR`.

I think the best would be to have the CI scripts create a tmpfs file system will a much smaller size (will need sudo access) such that it will trigger the error.  The test could have an assumption to check for the presence of that file system.  We'd only need to run it on Linux.

vyazelenko:
@mikeb01 The error translation happens in the `aeron_driver_conductor_on_command`:
```c
int os_errno = aeron_errcode();
int code = os_errno < 0 ? -os_errno : AERON_ERROR_CODE_GENERIC_ERROR;
aeron_driver_conductor_on_error(conductor, code, aeron_errmsg(), strlen(aeron_errmsg()), correlation_id);
aeron_driver_conductor_log_error(conductor);
```
and because the check emits `-AERON_ERROR_CODE_STORAGE_SPACE` it will come out as `AERON_ERROR_CODE_STORAGE_SPACE`.

I was considering to convert the raw `ENOSPC` into the `AERON_ERROR_CODE_STORAGE_SPACE` instead. But that would be different from all other custom errors codes we emit and also will not match Java which only handle insufficient space errors for the log files.

Thanks for the test I'll add it to the PR.

aeron-client/src/main/c/util/aeron_fileutil.c
```diff
@@ -66,8 +68,7 @@ static int aeron_mmap(aeron_mapped_file_t *mapping, int fd, bool pre_touch)
 
     if (!hmap)
     {
-        AERON_SET_ERR(errno, "%s", "CreateFileMapping");
-        close(fd);
+        AERON_SET_ERR_WIN(GetLastError(), "%s", "CreateFileMapping");
```
mikeb01, line 71: I like the use of the `AERON_SET_ERR_WIN` here I think that it is the right approach, but we need to make sure that the error handling is consistent on other OSs and that the call chain respects it.  We should do `AERON_SET_ERR` in the Linux/OSX version of aeron_mmap and all of the places that call this function (`aeron_map_existing_file`, `aeron_map_new_file`, `aeron_raw_log_map`, `aeron_raw_log_map_existing`) should be changed to use `AERON_APPEND_ERR` if the call to `aeron_mmap` fails.

aeron-client/src/main/c/util/aeron_fileutil.c
```diff
@@ -364,9 +370,8 @@ int aeron_map_new_file(aeron_mapped_file_t *mapped_file, const char *path, bool
         return -1;
     }
 
-    if (0 != aeron_mmap(mapped_file, fd, fill_with_zeroes))
+    if (0 != aeron_mmap(mapped_file, fd, fill_with_zeroes, path))
```
mikeb01, line 362: There is no need to pass the path in here for the purposes of logging.  The better approach is to use `AERON_APPEND_ERR`, e.g:

```java
if (0 != aeron_mmap(mapped_file, fd, fill_with_zeroes))
{
    AERON_APPEND_ERR("path=%s", path);
    close(fd);
    if (-1 == remove(path))
    {
```
This is the main purpose of `AERON_APPEND_ERR`, we shouldn't need to pass values up/down the call stack just to log them.  We can append context to the error stack at the point where we have the information.  This also provides additional trace information (file/line number) so that we can see the path through the stack where the error occurred.  Even if there is no additional context information to include in the log, we should do `AERON_APPEND_ERR("%", "")` anyway, so that we capture the trace information.

vyazelenko: Makes sense. I'll update the code accordingly.

aeron-client/src/main/c/util/aeron_fileutil.c
vyazelenko: Done.

aeron-client/src/main/c/util/aeron_fileutil.c
vyazelenko: Done.

mikeb01:
LGTM

aeron-archive/src/main/java/io/aeron/archive/RecordingWriter.java
```diff
@@ -234,13 +234,12 @@ private void onFileRollOver()
...
     private void checkErrorType(final IOException ex, final int writeLength)
     {
-        final String msg = ex.getMessage();
         boolean isLowStorageSpace = false;
         IOException suppressed = null;
 
         try
         {
-            isLowStorageSpace = (!Strings.isEmpty(msg) && msg.contains("No space left on device")) ||
+            isLowStorageSpace = StorageSpaceException.isOutOfDiscError(ex) ||
```
mjpt777, line 242: `isOutOfDiscError` -> `isStorageSpaceError`

aeron-driver/src/main/c/aeron_driver_conductor.c
```diff
@@ -2743,7 +2743,19 @@ aeron_rb_read_action_t aeron_driver_conductor_on_command(
...
+        {
+            code = -os_errno;
+        }
+        else if (AERON_FILEUTIL_ERROR_ENOSPC == os_errno)
+        {
+            code = AERON_ERROR_CODE_STORAGE_SPACE;
+        }
+        else
+        {
+            code = AERON_ERROR_CODE_GENERIC_ERROR;
```
mjpt777, line 2757: Better to initialise with this value then reset based on conditions.

mjpt777 (changes requested):
LGTM. Minor changes and I'm happy for a merge.

aeron-archive/src/main/java/io/aeron/archive/RecordingWriter.java
vyazelenko: Done.

aeron-driver/src/main/c/aeron_driver_conductor.c
vyazelenko: Done.

mjpt777 (approved):

PR #1375, opened 2022-10-31
Support importing ConsensusModule state to support transitioning.

mikeb01:
- Define structures to import state into the ConsensusModule.
- Support starting the ConsensusModuleAgent from existing state.
- Add more Standby/TransitionModule counters.
- Small testing updates and utilities.

### DIFF
--- a/aeron-client/src/main/java/io/aeron/AeronCounters.java
+++ b/aeron-client/src/main/java/io/aeron/AeronCounters.java
@@ -280,6 +280,17 @@ public final class AeronCounters
     public static final int CLUSTER_STANDBY_CONTROL_TOGGLE_TYPE_ID = 223;
 
     /**
+     * The type id of the {@link Counter} used for keeping track of the max duty cycle time of the cluster standby.
+     */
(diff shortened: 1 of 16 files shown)


aeron-client/src/main/java/io/aeron/AeronCounters.java
```diff
@@ -279,6 +279,43 @@
...
+
+    /**
+     * Counter type id for the transition module error count.
+     */
+    public static final int TRANSITION_MODULE_ERROR_COUNT_TYPE_ID = 226;
+
+    /**
+     * The type if of the {@link Counter} used for transition module state
+     */
+    public static final int TRANSITION_MODULE_STATE_TYPE_ID = 224;
```
vyazelenko, line 301: Why are ids are not sorted? If one scans from top to bottom it looks like some are missing or have been deleted.

mikeb01: I've grouped them by name rather than in counter id order.  This seemed tidier to me.  I've added a test to ensure that we don't have accidentally add clashing ids.

Issue #1376, opened 2022-11-02
aeronmd gets stuck with two full proxy queues, and both agents trying to offer


reissGRVS:
Running in shared network mode it is possible to completely lock aeronmd. This seems to be due to both the receiver proxy command queue and the conductor proxy command queue being full, and both the conductor and receiver being stuck in a loop attempting to offer onto the opposite queue.

I reproduced this with a slightly modified version of BasicSubscriber, that opens 150 subscriptions instead of 1.

Process:
1. Start aeronmd with THREADING_MODE_SHARED_NETWORK
2. Start 150 publishers `seq 1001 1151  | xargs -n 1 -I{} -P 150 ./BasicPublisher -s {} > /dev/null`
3. Run the modified BasicSubscriber which opens subscriptions to all of the above stream ids
4. Stop BasicSubscriber
5. Repeat steps 3 and 4 until the driver keepalive times out consistently, for even a single subscriber or publisher.

I expect the same issue can be run into in all threading modes that use the proxy command queues.
```
(gdb) bt
#0  0x00007f43dab6b71b in sched_yield ()
#1  0x00007f43dacec7f9 in aeron_driver_receiver_proxy_offer (receiver_proxy=0x55afa3ef8eb0, cmd=0x55afa3fc7a30)
    ~/aeron/aeron-driver/src/main/c/aeron_driver_receiver_proxy.c:27
#2  0x00007f43daced2a2 in aeron_driver_receiver_proxy_on_remove_publication_image (receiver_proxy=0x55afa3ef8eb0, 
    image=0x55afa3f4d910)
    ~/aeron/aeron-driver/src/main/c/aeron_driver_receiver_proxy.c:388
#3  0x00007f43dacf9a40 in aeron_publication_image_on_time_event (conductor=0x55afa3ef7fb8, image=0x55afa3f4d910, 
    now_ns=3679428801108610, now_ms=1667385937553)
    ~/aeron/aeron-driver/src/main/c/aeron_publication_image.c:913
#4  0x00007f43dacd623e in aeron_publication_image_entry_on_time_event (conductor=0x55afa3ef7fb8, entry=0x55afa3f941d8, 
    now_ns=3679428801108610, now_ms=1667385937553)
    ~/aeron/aeron-driver/src/main/c/aeron_driver_conductor.c:1276
#5  0x00007f43dacd6d31 in aeron_driver_conductor_on_check_managed_resources (conductor=0x55afa3ef7fb8, 
[...]

PR #1377, opened 2022-11-03
Use ring buffers instead of queues for messaging between agents in the driver.

mikeb01:
Replace queues with ring buffers for sending commands between the driver agents (sender, receiver, conductor). Remove the need to queue a message back to the conductor to free allocated messages.  #1376.

### DIFF
--- a/aeron-client/src/main/c/aeron_alloc.c
+++ b/aeron-client/src/main/c/aeron_alloc.c
@@ -128,7 +128,6 @@ int aeron_reallocf(void **ptr, size_t size)
     return 0;
 }
 
-#include <stdio.h>
 void aeron_free(void *ptr)
 {
     free(ptr);
--- a/aeron-client/src/main/c/aeron_alloc.c
+++ b/aeron-client/src/main/c/aeron_alloc.c
@@ -128,6 +128,7 @@ int aeron_reallocf(void **ptr, size_t size)
     return 0;
 }
 
+#include <stdio.h>
 void aeron_free(void *ptr)
 {
     free(ptr);
--- a/aeron-client/src/main/c/concurrent/aeron_mpsc_rb.c
+++ b/aeron-client/src/main/c/concurrent/aeron_mpsc_rb.c
@@ -356,8 +356,6 @@ bool aeron_mpsc_rb_unblock(aeron_mpsc_rb_t *ring_buffer)
     return unblocked;
 }
 
-extern void *aeron_mpsc_rb_message(aeron_mpsc_rb_t *ring_buffer, int32_t offset);
-
 extern int64_t aeron_mpsc_rb_consumer_position(aeron_mpsc_rb_t *ring_buffer);
 
 extern int64_t aeron_mpsc_rb_producer_position(aeron_mpsc_rb_t *ring_buffer);
--- a/aeron-client/src/main/c/concurrent/aeron_mpsc_rb.c
+++ b/aeron-client/src/main/c/concurrent/aeron_mpsc_rb.c
@@ -356,6 +356,8 @@ bool aeron_mpsc_rb_unblock(aeron_mpsc_rb_t *ring_buffer)
     return unblocked;
 }
 
+extern void *aeron_mpsc_rb_message(aeron_mpsc_rb_t *ring_buffer, int32_t offset);
+
 extern int64_t aeron_mpsc_rb_consumer_position(aeron_mpsc_rb_t *ring_buffer);
 
 extern int64_t aeron_mpsc_rb_producer_position(aeron_mpsc_rb_t *ring_buffer);
--- a/aeron-driver/src/main/c/aeron_driver_conductor_proxy.c
+++ b/aeron-driver/src/main/c/aeron_driver_conductor_proxy.c
@@ -21,10 +21,17 @@
 
 void aeron_driver_conductor_proxy_offer(aeron_driver_conductor_proxy_t *conductor_proxy, void *cmd, size_t length)
 {
-    while (aeron_mpsc_rb_write(conductor_proxy->command_queue, 1, cmd, length) < 0)
+    aeron_rb_write_result_t result;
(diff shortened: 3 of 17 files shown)


aeron-driver/src/main/c/aeron_driver_conductor_proxy.c
```diff
@@ -19,31 +19,15 @@
 #include "aeron_alloc.h"
 #include "aeron_driver_conductor.h"
 
-void aeron_driver_conductor_proxy_offer(aeron_driver_conductor_proxy_t *conductor_proxy, void *cmd)
+void aeron_driver_conductor_proxy_offer(aeron_driver_conductor_proxy_t *conductor_proxy, void *cmd, size_t length)
 {
-    while (aeron_mpsc_concurrent_array_queue_offer(conductor_proxy->command_queue, cmd) != AERON_OFFER_SUCCESS)
+    while (aeron_mpsc_rb_write(conductor_proxy->command_queue, 1, cmd, length) < 0)
```
reissGRVS, line 24: Handling AERON_RB_ERROR properly would give some peace of mind here. Currently if one of the proxy_offer functions somehow tries to write a message over the length limit, they end up spinning on this error.

This applies to all the proxy_offer implementations.

aeron-client/src/main/c/concurrent/aeron_mpsc_rb.c
```diff
@@ -356,6 +356,8 @@ bool aeron_mpsc_rb_unblock(aeron_mpsc_rb_t *ring_buffer)
     return unblocked;
 }
 
+extern void *aeron_mpsc_rb_message(aeron_mpsc_rb_t *ring_buffer, int32_t offset);
```
reissGRVS, line 359: Leftover?

aeron-client/src/main/c/aeron_alloc.c
```diff
@@ -128,6 +128,7 @@ int aeron_reallocf(void **ptr, size_t size)
     return 0;
 }
 
+#include <stdio.h>
```
reissGRVS, line 131: Leftover?

aeron-driver/src/main/c/aeron_driver_sender.c
```diff
@@ -115,18 +115,14 @@ int aeron_driver_sender_init(
...
-    aeron_command_base_t *cmd = (aeron_command_base_t *)item;
-    bool is_delete_cmd = cmd->func == aeron_command_on_delete_cmd;
-
+    aeron_command_base_t *cmd = (aeron_command_base_t *)message;
     cmd->func(clientd, cmd);
-
-    if (!is_delete_cmd)
-    {
-        aeron_driver_conductor_proxy_on_delete_cmd(sender->context->conductor_proxy, cmd);
-    }
```
reissGRVS, line 129: This section seems like a worthwhile change. 

The rest of the change seems nice for tidying up the proxy files but I don't think it truly solves the issue.
Afaict the only reason its harder to reproduce now is that you've effectively increased the size of the command queue massively.
In 1.40.0 it was 128 commands long, in this change its 128KB long, Even the largest command doesn't look to be larger than 100 bytes with the ring buffer header, so the queue is holding a minimum of 1000 commands now.

I see that today the queue size has been increased in master anyway https://github.com/real-logic/aeron/commit/1f7ff46d37699e2a8e62844ee7cab5b0c8dfcb63.

Either way it seems like its still technically possible to end up in this locked queue situation, it just takes a much higher load to get there.

mikeb01: > Afaict the only reason its harder to reproduce now is that you've effectively increased the size of the command queue massively.

While this change does increase the buffer space, that is not the primary fix.  Prior to implementing this change, I experimented with turning the queue size back to 128 and replacing the `*_proxy_on_delete_command` messages with an inline free.  I was unable to replicate this issue with that change in place.

Using a ring buffer instead of a queue was mostly down to coming up with an implementation that avoided the need for `alloc`/`free` of the messages between the agents (thereby removing the need for the `*_proxy_on_delete_command` messages altogether).  We didn't want to go with the inline `free` as the fix because (depending on allocator) you can get jitter in the call to `free`, especially if the memory moves between threads.  This is something that we actively try to avoid in the sender/receiver.

As you mention it is still technically possible to get a deadlock situation, but it is not a factor of load.  It requires a very specific operation and a lot of load, to point where I think that it is highly unlikely in practical terms (i.e. to trigger it would probably require artificial load rather than that of a real application).  That's not to say that we won't fix it, but a more general solution requires more thought as it may result in a behaviour change that impacts users.  Even with a general solution in place we wanted to make this change anyway as it simplifies the code.

reissGRVS: Interesting, that gives me more confidence. I would've liked to try reproducing with the inline free version as well (if you still have this I'm happy to do so now).

Regardless of that, thanks for the full explanation of the change, and the speed of your responses/fix!

aeron-driver/src/main/c/aeron_driver_context.h
```diff
@@ -32,7 +32,7 @@
 #include "aeron_cnc_file_descriptor.h"
 #include "aeron_duty_cycle_tracker.h"
 
-#define AERON_COMMAND_QUEUE_CAPACITY (512)
+#define AERON_COMMAND_QUEUE_CAPACITY (128)
```
mjpt777, line 35: Is this still used?

mikeb01: Yes, but I need to change the way it is used. Will pick that up next week.

aeron-driver/src/main/c/aeron_driver_sender.c
mikeb01: I don't, but you can try it by replacing the implementation of `aeron_driver_conductor_proxy_on_delete_cmd` with a free instead of sending the message.  We are also planning on some changes to that will back pressure the client as the queues to get full.

PR #1383, opened 2022-11-17
[Java] Delay election state transitions if there is an active leader.

vyazelenko:
The transition from `CANVASS` to `NOMINATE` and from `CANDIDATE_BALLOT` to `LEADER_LOG_REPLICATION` in case the corresponding vote/candidacy check is not unanimous is done with a timeout that is based on the time of the last message received from the leader. Which is tracked upon receiving the `onCommitPosition` and `onNewLeadershipTerm` messages.
The delay is necessary to prevent the non-eligible majority of forming a quorum and electing a new leader.

### DIFF
--- a/aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModule.java
+++ b/aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModule.java
@@ -1371,8 +1371,6 @@ public final class ConsensusModule implements AutoCloseable
         @SuppressWarnings("MethodLength")
         public void conclude()
         {
-            final ExpandableArrayBuffer buffer = new ExpandableArrayBuffer();
-
             if (0 != IS_CONCLUDED_UPDATER.getAndSet(this, 1))
             {
                 throw new ConcurrentConcludeException();
@@ -1398,6 +1396,12 @@ public final class ConsensusModule implements AutoCloseable
             if (!clusterDir.exists() && !clusterDir.mkdirs())
             {
                 throw new ClusterException("failed to create cluster dir: " + clusterDir.getAbsolutePath());
+            }
+
+            if (startupCanvassTimeoutNs <= leaderHeartbeatTimeoutNs)
+            {
+                throw new ClusterException("startupCanvassTimeoutNs=" + startupCanvassTimeoutNs +
+                    " must be greater than leaderHeartbeatTimeoutNs=" + leaderHeartbeatTimeoutNs);
             }
 
             if (null == clusterClock)
@@ -1499,6 +1503,8 @@ public final class ConsensusModule implements AutoCloseable
                     aeron.context().errorHandler(countedErrorHandler);
                 }
             }
+
+            final ExpandableArrayBuffer buffer = new ExpandableArrayBuffer();
 
             if (null == moduleStateCounter)
             {
--- a/aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModule.java
+++ b/aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModule.java
@@ -1401,7 +1401,7 @@ public final class ConsensusModule implements AutoCloseable
             if (startupCanvassTimeoutNs / leaderHeartbeatTimeoutNs < 2)
             {
                 throw new ClusterException("startupCanvassTimeoutNs=" + startupCanvassTimeoutNs +
-                    " must be multiples of the leaderHeartbeatTimeoutNs=" + leaderHeartbeatTimeoutNs);
+                    " must be a multiple of leaderHeartbeatTimeoutNs=" + leaderHeartbeatTimeoutNs);
             }
 
             if (null == clusterClock)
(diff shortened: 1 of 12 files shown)


aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModule.java
```diff
@@ -1400,6 +1398,12 @@ public void conclude()
                 throw new ClusterException("failed to create cluster dir: " + clusterDir.getAbsolutePath());
             }
 
+            if (startupCanvassTimeoutNs <= leaderHeartbeatTimeoutNs)
```
mikeb01, line 1401: Did we want `startupCanvassTimeoutNs` to be a multiple of `leaderHeartbeatTimeoutNs`?

aeron-cluster/src/main/java/io/aeron/cluster/Election.java
```diff
@@ -59,6 +59,7 @@
     private long timeOfLastStateChangeNs;
     private long timeOfLastUpdateNs;
     private long timeOfLastCommitPositionUpdateNs;
+    private long timeOfLastLeaderMessageReceivedNs;
```
mikeb01, line 62: Do we need to think about initialising this value when we start and reset an election?  What would be a sensible default when the election starts?  Should it be 0 as we are assuming that the leadership timeout has already expired to trigger the election or should it be initialised to the timestamp of when the election starts.

If the election resets, should we re-initialise this field?

aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModule.java
vyazelenko: Yes, you are right. I'll adjust the validation to ensure that canvass timeout is at least twice as large as the election timeout.

aeron-cluster/src/main/java/io/aeron/cluster/Election.java
vyazelenko: This can be solved if we move the field to the `ConsensusModuleAgent` and have it track the time instead. Then each election will automatically have access to the right value.

aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModule.java
vyazelenko: Done.

aeron-cluster/src/main/java/io/aeron/cluster/Election.java
vyazelenko: The time is tracked by the `ConsensusModuleAgent` now.
I don't think we need to do anything special if election resets.

mikeb01: I think you're correct.  If the election is reset, the last time that a message has been received from the leader doesn't change.

aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModule.java
```diff
@@ -1400,6 +1398,12 @@ public void conclude()
                 throw new ClusterException("failed to create cluster dir: " + clusterDir.getAbsolutePath());
             }
 
+            if (startupCanvassTimeoutNs / leaderHeartbeatTimeoutNs < 2)
+            {
+                throw new ClusterException("startupCanvassTimeoutNs=" + startupCanvassTimeoutNs +
+                    " must be multiples of the leaderHeartbeatTimeoutNs=" + leaderHeartbeatTimeoutNs);
```
mikeb01, line 1404: " must be a multiple of…"

aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModuleAgent.java
```diff
@@ -862,6 +863,9 @@ void onNewLeadershipTerm(
             throw new AgentTerminationException();
         }
 
+        final long nowNs = clusterClock.timeNanos();
+        timeOfLastLeaderMessageReceivedNs = nowNs;
```
mikeb01, line 867: One question for this logic; how protective should we be against byzantine failures?  Should we check that the newLeadershipTerm/commitPosition message received is from the node that we currently consider to be the leader?

aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModule.java
vyazelenko: Fixed.

aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModuleAgent.java
vyazelenko: It does not work, i.e. if we were to check for the `leadershipTermId` then we would not update the timeout and as a result there would be no delay applied in the election. Instead we trust any node that is currently a leader and update the time a message was received. Otherwise the case of an old valid leader with two empty follower nodes will not work.

mjpt777: timeOfLastLeaderMessageRecievedNs -> timeOfLastLeaderUpdateNs

mjpt777:
LGTM

aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModuleAgent.java
vyazelenko: Done.

PR #1392, opened 2022-12-15
[Java] Make Archive counters unique.

vyazelenko:
- Adds a notion of `archiveId` (`long`) which can be set on the `Archive.Context` programmatically or via the system property `aeron.archive.id`.
  If not set then it defaults to `Aeron.clientId()` of the corresponding `Aeron` instance.
- Uses `archiveId` to make Archive-specific counters unique, i.e. adds `archiveId` to the key and label (i.e. `` similar to what `ClusterCounters` does.
- Changes `RecordingPos` to include `archiveId` in the key and label.

Here a sample of how the counters look like after the change:
```java
...
 33:           21,365,164 - archive-conductor max cycle time in ns - archiveId=1
 34:                    0 - archive-conductor work cycle time exceeded count: threshold=1000000000ns - archiveId=1
 35:            2,268,616 - archive-recorder max cycle time in ns - archiveId=1
 36:                    0 - archive-recorder work cycle time exceeded count: threshold=1000000000ns - archiveId=1
 37:            1,476,860 - archive-replayer max cycle time in ns - archiveId=1
 38:                    0 - archive-replayer work cycle time exceeded count: threshold=1000000000ns - archiveId=1
 39:                    1 - Archive Control Sessions - archiveId=1
 40:            1,832,163 - archive-recorder max write time in ns - archiveId=1
 41:       10,716,987,392 - archive-recorder total write bytes - archiveId=1
 42:        2,977,878,326 - archive-recorder total write time in ns - archiveId=1
 43:                    0 - archive-replayer max read time in ns - archiveId=1
 44:                    0 - archive-replayer total read bytes - archiveId=1
 45:                    0 - archive-replayer total read time in ns - archiveId=1
...
 73:        4,327,473,152 - rec-pos: 1 994977179 1001 aeron:ipc - archiveId=1
```

### DIFF
--- a/aeron-archive/src/main/cpp/client/RecordingPos.h
+++ b/aeron-archive/src/main/cpp/client/RecordingPos.h
@@ -38,6 +38,9 @@ namespace aeron { namespace archive { namespace client
  *  |                                                              ...
  * ...                                                              |
  *  +---------------------------------------------------------------+
+ *  |                         Archive ID                            |
+ *  |                                                               |
+ *  +---------------------------------------------------------------+
  * </pre>
  */
 namespace RecordingPos
@@ -58,6 +61,11 @@ struct RecordingPosKeyDefn
 };
 #pragma pack(pop)
 
+static const util::index_t RECORDING_ID_OFFSET = offsetof(RecordingPosKeyDefn, recordingId);
+static const util::index_t SESSION_ID_OFFSET = offsetof(RecordingPosKeyDefn, sessionId);
+static const util::index_t SOURCE_IDENTITY_LENGTH_OFFSET = offsetof(RecordingPosKeyDefn, sourceIdentityLength);
+static const util::index_t SOURCE_IDENTITY_OFFSET = SOURCE_IDENTITY_LENGTH_OFFSET + sizeof(int32_t);
+
 /**
  * Find the active counter id for a stream based on the recording id.
  *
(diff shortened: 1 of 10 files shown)


aeron-archive/src/main/cpp/client/RecordingPos.h
```diff
@@ -29,6 +29,9 @@ namespace aeron { namespace archive { namespace client
  *   0                   1                   2                   3
  *   0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
  *  +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+ *  |                         Archive ID                            |
```
mjpt777, line 32: It would be better if archive id came at the end to not break backwards compatibility.

vyazelenko: The key ends with vardata (`sourceInformation`) that's why I added `archiveId` first to avoid having fixed size data after the vardata. But doing so breaks existing clients which expect the key fields to be present at certain offsets.

vyazelenko: I've moved the `archiveId` to the end of the key for the `RecordingPos` counter. I've also used the same label suffix as for other counters.

PR #1400, opened 2022-12-23
[java] leader transfer manually for aeron cluster

spiritlcx:
When releasing new version of application code, we need to shut down leader node.
And it will take 10s to have a new leader, which means the service will stop for at least 10s.

If we have leader transfer functionality, this will reduce to 200ms.

The code has been tested, please have a look, thanks very much.


mjpt777:
This is not the type of approach we plan to take. A separate stream of development is using different techniques to achieve a similar goal.

spiritlcx:
> This is not the type of approach we plan to take. A separate stream of development is using different techniques to achieve a similar goal.

Cool, looking forward to it. Will it be available in the open source version. If yes, is there an estimate time when it will be ready?

PR #1407, opened 2023-01-11
Queue Resource Free Operations

### LINKED ISSUE #1388: MediaDriver keepalive timeout due to driver being slow on munmaps
In the case where we have an application subscribing to a large number of streams, if we then shut that application down we have a massive influx of freeing the publication images.
Due to munmap being potentially slow (in the milliseconds) and the large number of streams, we can see a single driver_conductor work loop take over 1 second. This then triggers media driver keepalive timeouts for clients.

Would it be possible to queue up the munmaps so that other driver_conductor work isn't entirely blocked by a large number of munmaps? i.e. only free N publication images per driver_conductor work loop?

mikeb01:
- Separate the close and free operations on `DriverManagedResource`s
- When a resource reaches end of life, the resource is added to a separate queue.
- The conductor duty cycle will free a configurable number of the resources from the queue.  This allows the free operations to interleave with additional conductor actions to prevent the driver from blocking until resources are freed.
- Allow configuration of the number of resources freed per duty cycle (default 10).
- Intended to fix #1388


aeron-driver/src/main/java/io/aeron/driver/DriverConductor.java
```diff
@@ -1968,21 +1972,54 @@ private <T extends DriverManagedResource> void checkManagedResources(
...
+            }
+
+            if (resource.free())
+            {
+                workCount++;
+            }
+            else
+            {
+                ctx.systemCounters().get(FREE_FAILS).incrementOrdered();
+                endOfLiveResources.addLast(resource);
```
vyazelenko, line 2016: This changes the order in which the resources are closed, is this intentional?

TestTimestamps.java
```diff
@@ -0,0 +1,18 @@
+/*
```
vyazelenko, line 1: What is this file for?

aeron-driver/src/main/java/io/aeron/driver/DriverManagedResource.java
```diff
@@ -18,7 +18,7 @@
 /**
  * Common behaviour supported for driver resources such as publications and images.
  */
-public interface DriverManagedResource extends AutoCloseable
+public interface DriverManagedResource
```
vyazelenko, line 21: Why remove `AutoCloseable`?

aeron-driver/src/main/java/io/aeron/driver/DriverConductor.java
mikeb01: This only changes the ordering of the `free` operations.  This is now independent of the `close`.  Frees should be able to happen in any order (they are just unmapping and removing files).  Closes are decoupled from the frees and have already happened in the appropriate order.

aeron-driver/src/main/java/io/aeron/driver/DriverManagedResource.java
mikeb01: Because I removed the free operation from the close within the various driver managed resources I wanted to discourage people from thinking that something like the following would be valid:
```
try (final NetworkPublication publication = ....)
{
    // Do stuff
}
```
This is not really valid now as it will leak file resources.  I would like the change the name of the `close` operation, but not sure what would be correct.

vyazelenko: One consequence of making `DriverManagedResource` not `AutoCloseable` is that the actual invocation of the `close` operation now allocates:
```java
CloseHelper.close(ctx.errorHandler(), resource::close);
```

mikeb01: Yes, I'm conflicted on that one.  Allocation vs indicating that a language feature applies when it doesn't.  The other option would be to inline the behaviour from the CloseHelper.  I'll have a look at what feels best.

vyazelenko: If the intent is to remove the `AutoClosable` then inlining the `CloseHelper` logic seems the most straightforward to me.

vyazelenko:
LGTM.

vyazelenko (approved):

aeron-driver/src/main/java/io/aeron/driver/DriverManagedResource.java
mikeb01: I ran the driver through the async-profiler's allocation profiler and it didn't show any allocations for that section of the code.  It is possible in some cases that a method reference can allocate, but in this particular case I can't see any evidence of it.  While adding and closing 20,000 exclusive publications, all of the allocations within the `DriverConductor::checkManagedResources` call tree come from `ReceiverProxy::removeCoolDown`, which has a capturing lambda.  I'm going to leave the code as is.

vyazelenko: Actually those lambdas do allocate and you can see it if you run with the `-XX:+UseG1GC -XX:-UseTLAB` options (which disables TLAB and allows async-profiler to see every allocation).
Allocating_lambdas_when_closing
As you can see from the above the line [DriverConductor.java:1977](https://github.com/mikeb01/Aeron/blob/7469a86ae3423a8464a04d8e657cf29734c37f1b/aeron-driver/src/main/java/io/aeron/driver/DriverConductor.java#L1977) does allocate.

mjpt777: Allocation here is not a big issue. The resources have to be allocated and don't churn that often.

aeron-driver/src/main/c/aeron_driver_context.c
```diff
@@ -973,6 +977,20 @@ int aeron_driver_context_init(aeron_driver_context_t **context)
...
         AERON_NETWORK_PUBLICATION_MAX_MESSAGES_PER_SEND);
 
+    _context->resource_free_limit = aeron_config_parse_uint32(
+        AERON_DRIVER_RESOURCE_FREE_LIMIT_ENV_VAR,
+        getenv(AERON_DRIVER_RESOURCE_FREE_LIMIT_ENV_VAR),
+        (uint32_t)_context->resource_free_limit,
+        1,
+        UINT32_MAX);
+
+    _context->resource_free_limit = aeron_config_parse_uint32(
```
vyazelenko, line 987: Config option `resource_free_limit` is assigned twice. Should it be `resource_free_queue_length` instead?

aeron-driver/src/main/c/aeron_publication_image.c
```diff
@@ -272,16 +272,31 @@ int aeron_publication_image_close(aeron_counters_manager_t *counters_manager, ae
...
+    {
+        return false;
+    }
+
+    aeron_free(image->log_file_name);
+    aeron_free(image);
+    return true;
+}
+
+extern bool aeron_publication_image_free_voidp(void *image);
```
vyazelenko, line 298: Could we avoid having an extra wrapper function by using `void *` in the actual function definition?
```C
bool aeron_publication_image_free(void *image)
```

aeron-driver/src/main/c/aeron_driver_context.h
```diff
@@ -134,6 +134,8 @@ typedef struct aeron_driver_context_stct
     uint32_t receiver_io_vector_capacity;                   /* aeron.receiver.io.vector.capacity = 2 */
     uint32_t sender_io_vector_capacity;                     /* aeron.sender.io.vector.capacity = 2 */
     uint32_t network_publication_max_messages_per_send;     /* aeron.network.publication.max.messages.per.send = 2 */
+    uint32_t resource_free_limit;                           /* aeron.driver.resource.free.limit = 10 */
+    uint32_t resource_free_queue_length;                    /* aeron.driver.resource.free.queue.length = 16384 */
```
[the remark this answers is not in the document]
mikeb01: Yes potentially.  Because we are using a spsc queue the size is fixed.  In the Java implementation we have a deque which grows.

aeron-driver/src/main/c/aeron_driver_context.c
mikeb01: Yes it should.

aeron-driver/src/main/c/aeron_publication_image.c
mikeb01: My preference would be for the typesafe version the voidp wrapper only being a necessity for the internals of the conductor.  Perhaps it would be better there.

aeron-driver/src/main/c/aeron_driver_context.h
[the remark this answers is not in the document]
mikeb01: I've added a very simple growable deque to the C code base.  It is incomplete (only had add_list, remove_first), but that can be expanded later.  This removes the need for the additional configuration option and is more consistent with the Java approach.

Issue #1411, opened 2023-01-16
[C++] Possible race condition in AtomicArrayUpdater


StephanDollberg:
Hi,

Please have a look at the following code using the C++ `AtomicArrayUpdater`. 

``` cpp
int main() {

    std::int64_t counter = 0;

    aeron::concurrent::AtomicArrayUpdater<std::int64_t> aau;
    aau.addElement(1);
    aau.addElement(2);
    aau.addElement(3);
    aau.addElement(4);
    aau.addElement(5);

    const std::int64_t max = 100000000;

    std::thread writer([&] () {
        std::int64_t i = 1;
        while (i < max) {
            ++i;

            auto res = aau.removeElement([&] (auto ele) {
                return ele == i;
            });
            aau.addElement(i + 5);
        }
    });

    std::thread reader([&] () {
        std::int64_t i = 0;
        while (i < max * 100)
        {
            ++i;
            auto res = aau.load();
            if (res.first == nullptr)
            {
                continue;
            }
            // std::printf("pop %p %lu\n", res.first, res.second);
            counter += res.first[res.second - 1];
        }
    });

    writer.join();
    reader.join();

    std::cout << counter << std::endl;
}
```

There is a fully compilable example here: https://gist.github.com/StephanDollberg/5587478f958d759a8c1d322a2617cc2a#file-test-cpp

Basically we are having two threads one constantly adds and removes elements while the other one just tries to read them (so in between the remove and add the size is different). Note that this has came up as part of normal Aeron client usage but is easier to reproduce as above.

Compiling above with clang 15 (haven't tested other versions yet) results in the following ASAN error:

```
clang++ -g -fno-omit-frame-pointer -O1 -fsanitize=address -std=c++2b test.cpp -o test
```

```
=================================================================
==1053728==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x6040000133f8 at pc 0x000000519676 bp 0x7fdf08bfddf0 sp 0x7fdf08bfdde8
READ of size 8 at 0x6040000133f8 thread T2
    #0 0x519675 in main::$_1::operator()() const /home/inf/build/aeron/test.cpp:167:24
    #1 0x519675 in void std::__invoke_impl<void, main::$_1>(std::__invoke_other, main::$_1&&) /usr/bin/../lib/gcc/x86_64-redhat-linux/12/../../../../include/c++/12/bits/invoke.h:61:14
[...]

PR #1413, opened 2023-01-19
[C++] AtomicArrayUpdater concurrency fixes.

vyazelenko:
This PR fixes concurrency issues in the `AtomicArrayUpdater` that manifest themselves when compiled with a recent clang compiler (version >= 14) and `-O1 -fsanitize=address` compilation options. The fix is based on using `atomic` fields instead of plain fields and the fences. See [Can Seqlocks Get Along With Programming Language Memory Models?](https://www.hpl.hp.com/techreports/2012/HPL-2012-68.pdf) for information on correctness and why plain fields are not enough.

The new `AtomicArrayUpdaterTest` contains concurrent tests for the `AtomicArrayUpdater` class one of which fails on clang when `cppbuild/cppbuild --compiler-optimization-level 1 --sanitise-build ...` (which is similar to the `-O1 -fsanitize=address`).

Fixes issue #1411.


aeron-client/src/main/cpp/concurrent/AtomicArrayUpdater.h
```diff
@@ -91,52 +93,68 @@ class AtomicArrayUpdater
...
+    {
+        const std::uint64_t newChangeNumber = m_beginChange.load(std::memory_order_relaxed) + 1;
+        m_beginChange.store(newChangeNumber, std::memory_order_release);
+
+        // The `release` fence which makes two following `relaxed` stores into the `release` stores and synchronizes
+        // with the `acquire` fence from the `load` method.
+        aeron::concurrent::atomic::release();
+        m_array.store(newArray, std::memory_order_relaxed);
+        m_length.store(newLength, std::memory_order_relaxed);
+
```
mjpt777, line 155: Release fence required here.

mjpt777 (changes requested):
Looks good. Needs the release fence mentioned.

aeron-client/src/main/cpp/concurrent/AtomicArrayUpdater.h
mjpt777: Sorry I missed the fence above.

mjpt777 (approved):

PR #1414, opened 2023-01-24
ArchiveTool with separate mark file.

mikeb01:
Support using ArchiveTool when the archive directory and mark file directory are different.


aeron-archive/src/test/java/io/aeron/archive/ArchiveToolSeparateMarkFileTest.java
```diff
@@ -0,0 +1,86 @@
...
+import io.aeron.test.TestContexts;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+
+import java.io.ByteArrayOutputStream;
+import java.io.File;
+import java.io.PrintStream;
+import java.util.Objects;
+
+public class ArchiveToolSeparateMarkFileTest
```
mjpt777, line 28: Tests do not need to be public.

mjpt777:
On the whole LGTM.

PR #1415, opened 2023-01-24
Static analysis for aeron

Nonkle:
I ran some static analysis locally and found some problems. I'm happy to post a PR with some fixes - but ideally some analysis could be enabled on the repo to prevent problems creeping in again.

Does the `aeron` project have any interest in pursuing this? If so we can continue to refine this setup ... a first step might be to restrict the analysis to a single folder and filter only the queries we think are actually urgent.

(note) I had to remove the existing workflows so the fork runs only the analysis... the PR cannot be merge 'as-is'

----------------

I tried enabling CodeQL in Github - built-in static analysis. The results are at https://github.com/Nonkle/aeron/security/code-scanning

The Github UI is not great - it is very hard to dig into that list and see what is important. But there are some real problems hiding (below is output from Coverity that I ran locally):

eg, off-by-one errors overwriting arrays:

aeron/aeron-client/src/main/c/util/aeron_http_util.c

```
(38) Event cond_false:Condition "length > 512 /* (int)sizeof (parsed_url->path_and_query) */", taking false branch.(40) Event cond_at_most:Checking "length > 512" implies that "length" may be up to 512 on the false branch.Also see events:[overrun-local]
69   	        if (length > (int)sizeof(parsed_url->path_and_query))
70   	        {
71   	            AERON_SET_ERR(EINVAL, "URL %s has too long path", url);
72   	            return -1;
(39) Event if_end:End of if statement.
73   	        }
74   	
75   	        memcpy(parsed_url->path_and_query, pb + first_slash_index, (size_t)length);
(41) Event overrun-local:Overrunning array "parsed_url->path_and_query" of 512 bytes at byte offset 512 using index "length" (which evaluates to 512).Also see events:[cond_at_most]
76   	        parsed_url->path_and_query[length] = '\0';
77   	        end_index = first_slash_index;
78   	    }
```


Nonkle:
@vyazelenko I see you enabled this _yesterday_, what are the odds? :) 

Are you sure that configuration is actually finding any problems? I had to add extra queries to get any results.

I'm also comparing to Coverity I run locally. Perhaps the best approach at this point is I just submit PR with fixes to our Coverity problems... the CodeQL analysis can be left to the aeron team.

vyazelenko:
@Nonkle Why did you delete the other workflows?

Nonkle:
@vyazelenko I didn't know how to set up CodeQL, deleting the workflows was needed to get the "Setup automatically ... " link to work.

I didn't try bringing them back as I didn't want to run them all.

At this point there's not much point to my PR; you could just add the line `queries: security-extended,security-and-quality`

vyazelenko:
According to the [Using queries in QL packs](https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/customizing-code-scanning#using-queries-in-ql-packs) page `security-and-quality` is a superset of `security-extended` and we already have the former enabled.

PR #1417, opened 2023-02-01
[Java] Allow running Archive in IPC-only mode.

### LINKED ISSUE #1416: [AeronArchive] Allow disabling UDP aeron.archive.control.response.channel
Hi,

The defaul value for `aeron.archive.control.response.channel` creates a localhost subscription. This is in addition to an `aeron:ipc` one that is also created by default.

Equally it doesn't seem to be possible to disable the UDP channel altogether. Set `aeron.archive.control.response.channel` to something non-udpish makes it error out.

This is an issue because a localhost socket can't be accelerated by most kernel bypass LD_PRELOAD solutions which in turn means that the default forces the Aeron receive thread to poll the system even if kernel bypassed.

My understanding of aeron archive might be lacking but I don't understand why you'd want a localhost subscription in addition to an `ipc` one anyway.

vyazelenko:
This PR adds a new configuration property `aeron.archive.control.channel.enabled` which allows disabling control channel (UDP) when set to `false`. In which case the Archive only listens to the IPC requests.
Setting this property to `false` also requires that the `Archive.Context.archiveClientContext.controlResponseChannel` is set as it won't be derived from the `Archive.Context.controlChannel`.

Fixes #1416.

### DIFF
--- a/aeron-archive/src/main/java/io/aeron/archive/Archive.java
+++ b/aeron-archive/src/main/java/io/aeron/archive/Archive.java
@@ -504,14 +504,12 @@ public final class Archive implements AutoCloseable
         public static final String ARCHIVE_ID_PROP_NAME = "aeron.archive.id";
 
         /**
-         * Is control channel enabled.
+         * Is control channel enabled. Defaults to {@code true}.
+         * <p>
+         * If set to anything other than {@code true} then control channel is disabled, i.e. the Archive will run in
+         * IPC-only mode.
          */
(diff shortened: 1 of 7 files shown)


mikeb01:
The code looks good.  I am wondering if we need a test that to show that replication still works when running in IPC mode?

vyazelenko:
@mikeb01 I've added a new system test for that. See https://github.com/real-logic/aeron/pull/1417/commits/60c93955f914eeee12f845b21d21d539ea5498eb.

PR #1421, opened 2023-02-07
Set max of resource free queue length and resource free limit to INT32_MAX

reissGRVS:
This stops them being incorrectly set to 0 by aeron_config_parse_uint32 when comparing against int32 0
Current behaviour will always set the value to 0 if the environment variable is specified.


reissGRVS:
Might make sense to fix this in aeron_config_parse_uint32 instead.

mikeb01:
I'm going to merge this and I'll fix up `aeron_config_parse_uint32` too.

PR #1422, opened 2023-02-13
Refactored Cluster Replication

mikeb01:
Unify LogReplication and SnapshotReplication in a single class so that progress checking can be used wherever replication is required within Cluster, (log replication, dynamic join, backup).


mikeb01:
Merged directly.

PR #1423, opened 2023-02-13
[C] Set URI for receiver counters based on matching subscription

reissGRVS:
Currently the URI of the first subscription is used for all other subscriptions.
If you create two subscriptions, one on channel A and the next on channel B the receiver counters (rcv-pos and rcv-hwm) will both report their channel as A for both subscriptions.

I think this fixes that.


mikeb01:
Thanks, I've pull this change in here: 6d91a68.  I've also added a test to verify that the counter URIs were formatted correctly.

PR #1424, opened 2023-02-14
[C] Allow for preallocation of pool of buffers

reissGRVS:
The motivation behind this change is to spread out the cost of mmaps across work cycles and/or pay the cost at start up.

 I'd like confirmation this is something that would actually be merged before spending more time tidying up the global storage and adding proper tests. I've run this manually myself and it seems to work well.


mikeb01:
Thank you for the PR, we will have a look at it.  To confirm the purpose of this, is it to improve responsiveness when creating publications or when large numbers of images become available?

mjpt777:
Have you considered how this is not very effective when streams use various term buffer lengths?

reissGRVS:
> Have you considered how this is not very effective when streams use various term buffer lengths?

I figure its not a problem as this is completely opt in. 
For our use case we widely use two larger term-lengths where this is useful. 
I can't pretend to know how other people are using aeron but I wouldn't be surprised that other users are also using only a few discrete term-lengths across all streams.

tmontgomery:
In most usages we see, there is a clear separation of buckets of sizes—usually the default size (or the configured size) and 64K. So, I think to be useful, it would be good to have at least 2 buckets.

reissGRVS:
> In most usages we see, there is a clear separation of buckets of sizes—usually the default size (or the configured size) and 64K. So, I think to be useful, it would be good to have at least 2 buckets.

In my case this is exactly what I'm doing, 2 calls to aeron_driver_context_add_raw_log_pool_config to set up different pools for different term lengths. It's been working well in my usage so far

PR #1426, opened 2023-02-17
Store cluster membership changes in node state file

mikeb01:
- Support dynamically sized records in the node state file.  Handle this by moving trailing records up/down the file as required.
- Add a footer record for tracking the end of the file.
- Add cluster members record.
- Remove cluster member state from the consensus module snapshot.


tmontgomery (approved):
LGTM

PR #1428, opened 2023-02-22
[Java] Cluster tool is leader

eleventy7:
Adds an `is-leader` command to the Cluster Tool. This returns zero if the cluster node is the leader and non-zero if not.

### DIFF
--- a/aeron-cluster/src/main/java/io/aeron/cluster/ClusterTool.java
+++ b/aeron-cluster/src/main/java/io/aeron/cluster/ClusterTool.java
@@ -324,7 +324,7 @@ public class ClusterTool
     public static void recoveryPlan(final PrintStream out, final File clusterDir, final int serviceCount)
     {
         try (AeronArchive archive = AeronArchive.connect();
-             RecordingLog recordingLog = new RecordingLog(clusterDir, false))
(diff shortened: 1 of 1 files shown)


aeron-cluster/src/main/java/io/aeron/cluster/ClusterTool.java
```diff
@@ -613,6 +620,66 @@ public static void describe(final PrintStream out, final ClusterMarkFile[] servi
...
+     *                   -1 if the mark file does not exist.
+     */
+    public static int isLeader(final PrintStream out, final File clusterDir)
+    {
+        if (markFileExists(clusterDir))
+        {
+            try (ClusterMarkFile markFile = openMarkFile(clusterDir, System.out::println))
+            {
+                final String aeronDirectoryName = markFile.decoder().aeronDirectory();
+                final AtomicLong nodeRoleCounter = new AtomicLong(-1);
```
mjpt777, line 638: `MutableLong` is sufficient for these variables.

eleventy7: Thanks; I've pushed a change, switching to `MutableLong`.

PR #1429, opened 2023-02-23
[C++] Make MediaDriver keepalive exception consistent

reissGRVS:
Makes the keepalive timeout exception consistent with implementation in aeron_client_conductor_check_liveness
https://github.com/real-logic/aeron/blob/master/aeron-client/src/main/c/aeron_client_conductor.c#L381-L393

### DIFF
--- a/aeron-client/src/main/cpp/ClientConductor.h
+++ b/aeron-client/src/main/cpp/ClientConductor.h
@@ -446,13 +446,14 @@ private:
 
         if (nowMs > (m_timeOfLastKeepaliveMs + KEEPALIVE_TIMEOUT_MS))
         {
-            if (nowMs > (m_driverProxy.timeOfLastDriverKeepalive() + m_driverTimeoutMs))
+            
+            int64_t lastKeepaliveMs = m_driverProxy.timeOfLastDriverKeepalive();
+            if (nowMs > (lastKeepaliveMs + m_driverTimeoutMs))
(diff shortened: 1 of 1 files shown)


reissGRVS:
I dont think my change could've broken atomicArrayUpdaterTest in C++ Xcode (macOS) build.
I'm guessing thats another issue?

aeron-client/src/main/cpp/ClientConductor.h
```diff
@@ -446,13 +446,14 @@ class CLIENT_EXPORT ClientConductor
...
         if (nowMs > (m_timeOfLastKeepaliveMs + KEEPALIVE_TIMEOUT_MS))
         {
-            if (nowMs > (m_driverProxy.timeOfLastDriverKeepalive() + m_driverTimeoutMs))
+            
+            int64_t lastKeepaliveMs = m_driverProxy.timeOfLastDriverKeepalive();
+            if (nowMs > (lastKeepaliveMs + m_driverTimeoutMs))
             {
                 m_driverActive = false;
 
-                int64_t age = m_driverProxy.timeOfLastDriverKeepalive() + m_driverTimeoutMs;
```
reissGRVS, line 453: Another annoying thing about this was the two separate calls to timeOfLastDriverKeepalive().
In theory it could've changed in between and you end up logging a nonsense "age"

PR #1435, opened 2023-03-10
[C] Handle failed create_publication_image

reissGRVS:
Currently if create_publication_image fails:
1. It can fail silently without actually committing errors.
2. It will leave an entry in the data_packet_dispatcher for that session/stream in INIT_IN_PROGRESS state, for as long as there is a subscriber. This stops any potential for this stream/session to be subscribed to, even if the error was temporary.

This change removes the dead entry from the image_by_session_id_map in the packet_dispatcher, and ensures errors are logged if a failure is hit.

### DIFF
--- a/aeron-driver/src/main/c/aeron_driver_conductor.c
+++ b/aeron-driver/src/main/c/aeron_driver_conductor.c
@@ -4494,13 +4494,13 @@ void aeron_driver_conductor_on_create_publication_image(void *clientd, void *ite
         endpoint, (size_t)command->mtu_length, initial_window_length, conductor->context) < 0)
     {
         AERON_APPEND_ERR("%s", "");
-        aeron_driver_conductor_log_error(conductor);
-        return;
+        goto error_cleanup;
     }
 
     if (!aeron_driver_conductor_has_network_subscription_interest(
         conductor, endpoint, command->stream_id, command->session_id))
     {
+        aeron_driver_receiver_proxy_on_remove_init_in_progress(conductor->context->receiver_proxy, endpoint, command->session_id, command->stream_id);
         return;
     }
 
@@ -4508,7 +4508,7 @@ void aeron_driver_conductor_on_create_publication_image(void *clientd, void *ite
     AERON_ARRAY_ENSURE_CAPACITY(ensure_capacity_result, conductor->publication_images, aeron_publication_image_entry_t)
     if (ensure_capacity_result < 0)
     {
-        return;
+        goto error_cleanup;
     }
 
     const int64_t registration_id = aeron_mpsc_rb_next_correlation_id(&conductor->to_driver_commands);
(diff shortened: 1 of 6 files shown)


aeron-driver/src/main/c/aeron_driver_conductor.c
```diff
@@ -4495,20 +4495,21 @@ void aeron_driver_conductor_on_create_publication_image(void *clientd, void *ite
...
         aeron_driver_conductor_log_error(conductor);
-        return;
+        goto error_cleanup;
     }
 
     if (!aeron_driver_conductor_has_network_subscription_interest(
         conductor, endpoint, command->stream_id, command->session_id))
     {
-        return;
+        AERON_SET_ERR(AERON_ERROR_CODE_UNKNOWN_SUBSCRIPTION, "Image doesnt have network subscription interest", "");
```
reissGRVS, line 4504: Admittedly I'm not sure how this situation can occur, or if it is an error case really?

aeron-driver/src/main/c/aeron_driver_conductor.c
```diff
@@ -4635,11 +4636,17 @@ void aeron_driver_conductor_on_create_publication_image(void *clientd, void *ite
...
         }
     }
 
     aeron_driver_receiver_proxy_on_add_publication_image(conductor->context->receiver_proxy, endpoint, image);
+    return;
+
+error_cleanup:
+    AERON_APPEND_ERR("%s", "Failed aeron_driver_conductor_on_create_publication_image for %d %d", command->stream_id, command->session_id);
+    aeron_driver_conductor_log_error(conductor);
+    aeron_driver_receiver_proxy_on_remove_init_in_progress(conductor->context->receiver_proxy, endpoint, command->session_id, command->stream_id);
```
reissGRVS, line 4649: Potentially I should delay the actual removal, to stop excess failed attempts if the error condition is persistent

aeron-driver/src/main/c/aeron_driver_conductor.c
mikeb01: It isn't really an error condition.  This is to short circuit/no-op a case where the driver receives a setup for a stream that doesn't have any subscriptions.  I suspect that this code may even be redundant as the only code path that generates this is when we have a subscription that is awaiting setup, in which case it should have network subscription interest.

reissGRVS: Yeah I wasn't sure if it was possible to hit this with a small gap between subscription creation and the unsubscribe. Potentially if the conductor queue is a bit backed up?

reissGRVS: I've removed the error from here, but still clear the state in the receiver packet dispatcher

aeron-driver/src/main/c/aeron_driver_conductor.c
mikeb01: This is probably not too much of an issue if the error trace is the same for each error, i.e. the same logged parameters and call error stack, then the logger will collapse it into a single log entry and maintain a count.

PR #1438, opened 2023-03-14
[C] Fix rejoin logic by swapping cool down map insertion and removal.

reissGRVS:
Seems to have been broken by this commit: https://github.com/reissGRVS/aeron/commit/605a71feac0d101fb5006c52aea48e6ea1a844ec#r104348901

The cool down entry was being inserted after the remove attempt.

Remove attempt via aeron_driver_conductor_image_transition_to_linger https://github.com/real-logic/aeron/blob/f94e8e899ff5900556d7e6c5bab65fcd18784f89/aeron-driver/src/main/c/aeron_driver_conductor.c#L1351

Insert via aeron_driver_receiver_proxy_on_remove_publication_image->aeron_driver_receiver_on_remove_publication_image->aeron_receive_channel_endpoint_on_remove_publication_image->aeron_data_packet_dispatcher_remove_publication_image
https://github.com/real-logic/aeron/blob/f94e8e899ff5900556d7e6c5bab65fcd18784f89/aeron-driver/src/main/c/aeron_data_packet_dispatcher.c#L262

### DIFF
--- a/aeron-driver/src/main/c/aeron_publication_image.c
+++ b/aeron-driver/src/main/c/aeron_publication_image.c
@@ -921,10 +921,9 @@ void aeron_publication_image_on_time_event(
                 image->conductor_fields.state = AERON_PUBLICATION_IMAGE_STATE_LINGER;
                 image->conductor_fields.time_of_last_state_change_ns = now_ns;
 
-                aeron_driver_conductor_image_transition_to_linger(conductor, image);
-
                 aeron_receive_channel_endpoint_dec_image_ref_count(image->endpoint);
                 aeron_driver_receiver_proxy_on_remove_publication_image(conductor->context->receiver_proxy, image);
+                aeron_driver_conductor_image_transition_to_linger(conductor, image);
 
                 aeron_receive_channel_endpoint_try_remove_endpoint(image->endpoint);
             }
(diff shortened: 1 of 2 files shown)


aeron-driver/src/test/c/aeron_c_system_test.cpp
```diff
@@ -567,6 +567,87 @@ TEST_P(CSystemTest, shouldAllowImageToGoUnavailableWithPollAfter)
     EXPECT_EQ(aeron_subscription_close(subscription, nullptr, nullptr), 0);
 }
 
+
+TEST_P(CSystemTest, shouldAllowImageToGoUnavailableAndThenRejoin)
```
reissGRVS, line 571: This test blocks forever on current HEAD

aeron-driver/src/main/c/aeron_publication_image.c
```diff
@@ -921,10 +921,9 @@ void aeron_publication_image_on_time_event(
                 image->conductor_fields.state = AERON_PUBLICATION_IMAGE_STATE_LINGER;
                 image->conductor_fields.time_of_last_state_change_ns = now_ns;
 
-                aeron_driver_conductor_image_transition_to_linger(conductor, image);
-
                 aeron_receive_channel_endpoint_dec_image_ref_count(image->endpoint);
                 aeron_driver_receiver_proxy_on_remove_publication_image(conductor->context->receiver_proxy, image);
+                aeron_driver_conductor_image_transition_to_linger(conductor, image);
```
reissGRVS, line 927: On another note, I think COOL_DOWN is misleading now.
Before it made sense as it was in place for the entire DRAINING period, now its there either for a split second, or for the lifetime of the subscription set for this stream

tmontgomery: I think we might be able to eliminate COOL_DOWN. The original intent was to stop the image from flapping. However, the nature of a slow receiver that falls outside the window has a natural resonance and won't flap anyway. So, it's worth examining what the logic would look like refactored. But that is outside the scope of this PR.

PR #1442, opened 2023-03-28
Fix hash map remove if

reissGRVS:
aeron_int64_to_tagged_ptr_hash_map_remove_if would miss elements that were moved to the end of the array due to chain compaction.


reissGRVS:
Similar issue in aeron_int64_to_ptr_hash_map_remove_if, but it seems not to be used.

mikeb01:
Merged.

PR #1444, opened 2023-03-30
Support migrating segments from a source recording to the end of a (stopped) destination recording

ZachBray:
**Context:**

The aim of this PR is to support use cases where "appending migrations" are necessary, for example, as suggested recently, to fix snapshot recordings:

> truncate the original snapshot back to 0 length and then
> migrate the segments from the new recording to the old snapshot
> recording

**Main changes:**

- Compute whether to prepend or append based on the extent of the source and destination recordings
- Perform more checks, e.g., that all source segments are available and that no unattached destination segments conflict, before renaming any segments
    - Some cases that were previously partially successful are now failures, as a result
    - The conductor no longer attaches segments in order of closest to the "seam" to furthest away
- Added a new suite of tests to replace the existing `migrateSegments` tests, which comprises two parameterised tests:
    - `shouldMigrateSegments(TestCaseParameters)`
        - Takes a description of how to arrange the source and destination recordings
        - Computes a checksum from replaying the original source and destination parts
        - Calls `migrateSegments`
        - Computes another checksum from the migrated stream and verifies it matches the original
        - Checks the start and stop positions of the source and destination recordings match expectations
        - Checks the segments in the Archive directory are as expected
        - Checks both recordings are left in a valid state where they can later be extended
        - See [successful test cases](https://github.com/ZachBray/aeron/blob/f2430ccb9d4573ac8a5ed10cf01bd33beef5e5be/aeron-system-tests/src/test/java/io/aeron/archive/MigrateSegmentsTest.java#L204-L217)
    - `migrateSegmentsShouldThrow(FailureCaseParameters)`
        - Takes a description of how to arrange the source and destination recordings
        - Calls `migrateSegments` and catches the resulting exception
        - Verifies the error message matches expectations 
        - See [failure test cases](https://github.com/ZachBray/aeron/blob/f2430ccb9d4573ac8a5ed10cf01bd33beef5e5be/aeron-system-tests/src/test/java/io/aeron/archive/MigrateSegmentsTest.java#L219-L239)

**Questions:**

1. As mentioned above, some partial successes could now be failures, e.g., where there was a missing segment. Strictly speaking, this behaviour change is a breaking change. Is migrating with missing segments an expected/supported use case? Or is it reasonable for this to result in an error now?
2. Is "seam position" a suitable name for the position where source recordings are stitched onto a destination? Or should I (re)use "join position"?

**Please note** that the new `MigrateSegmentsTest` suite is collapsed by default in the diff viewer.


aeron-archive/src/main/java/io/aeron/archive/ArchiveConductor.java
```diff
@@ -1328,54 +1329,86 @@ void migrateSegments(
...
+                    " srcStartPosition=" + srcRecordingSummary.startPosition +
+                    " srcStopPosition=" + srcRecordingSummary.stopPosition +
+                    " dstStartPosition=" + dstRecordingSummary.startPosition +
+                    " dstStopPosition=" + dstRecordingSummary.stopPosition;
+                controlSession.sendErrorResponse(correlationId, msg, controlResponseProxy);
+                return;
+            }
+
+            if (isSeamPosSegmentUnaligned(controlSession, correlationId, "src", srcRecordingSummary, seamPosition) ||
+                isSeamPosSegmentUnaligned(controlSession, correlationId, "dst", dstRecordingSummary, seamPosition))
```
ZachBray, line 1371: I prefer `!positivePredicate` rather than `negativePredicate`, but the latter makes IntelliJ happy without suppressing warnings.

aeron-archive/src/main/java/io/aeron/archive/ArchiveConductor.java
```diff
@@ -1328,54 +1329,86 @@ void migrateSegments(
...
+            if (movedSegmentCount >= 0)
             {
-                catalog.startPosition(dstRecordingId, startPosition);
-                catalog.stopPosition(srcRecordingId, startPosition);
-                controlSession.sendOkResponse(correlationId, attachedSegmentCount, controlResponseProxy);
+                final int segmentToBeDeletedCount =
+                    addDeleteSegmentsSession(correlationId, srcRecordingId, controlSession, emptyFollowingSrcSegment);
 
-                if (files.isEmpty())
+                if (segmentToBeDeletedCount >= 0)
```
mjpt777, line 1391: Minor but `toBeDeletedSegmentCount` would be more consistent.

aeron-archive/src/main/java/io/aeron/archive/ArchiveConductor.java
```diff
@@ -80,6 +80,7 @@
     private volatile boolean isAbort;
 
     private final RecordingSummary recordingSummary = new RecordingSummary();
+    private final RecordingSummary otherRecordingSummary = new RecordingSummary();
```
mjpt777, line 83: Probably better to have this scoped to the call for simplicity.

mjpt777:
Besides minor comments this looks good.

aeron-archive/src/main/java/io/aeron/archive/ArchiveConductor.java
ZachBray: :+1: Fixed in latest commits.

aeron-archive/src/main/java/io/aeron/archive/ArchiveConductor.java
ZachBray: :+1: Fixed in latest commits.

PR #1455, opened 2023-04-27
[Java] Name resolution execution time tracking.

vyazelenko:
This PR adds a capability to track the execution of the `NameResolver#resolve` and `NameResolver#lookup` methods via a new set of counters. It also extends the existing logging events with the `durationNs` field and adds logging for the lookup method, e.g.:
```bash
[43598.051268] DRIVER: NAME_RESOLUTION_LOOKUP [74/74]: resolver=RedirectingNameResolver durationNs=6430 name=127.0.0.1:24326 isReLookup=false resolvedName=127.0.0.1:24326
[43598.051826] DRIVER: NAME_RESOLUTION_RESOLVE [57/57]: resolver=RedirectingNameResolver durationNs=21460 name=127.0.0.1 isReResolution=false address=127.0.0.1
```
There is also a new method `NameResolver#init(CounterProvider)` which was added to replace the `NameResolver#init(MediaDriver.Context)`. The latter could only be used in the driver while the former can be used in other places.

### DIFF
--- a/aeron-agent/src/main/java/io/aeron/agent/DriverComponentLogger.java
+++ b/aeron-agent/src/main/java/io/aeron/agent/DriverComponentLogger.java
@@ -250,7 +250,7 @@ public class DriverComponentLogger implements ComponentLogger
             NAME_RESOLUTION_RESOLVE,
             "DefaultNameResolver",
             DriverInterceptor.NameResolution.Resolve.class,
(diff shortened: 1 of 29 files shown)


tmontgomery (approved):
Looks good. I think lookup should also use the same tracking. Don't think it needs to be a separate counter, though.

PR #1457, opened 2023-05-09
[C] Implement name resolver time tracking.

vyazelenko:
* Adds time tracking for name resolver `resolve` and `lookup` functions. The same as the Java media driver.
* Adds `GENERIC_MESSAGE` logging event to ensure compatibility with the Java MD.
* Fixes a bug where the `aeron_duty_cycle_stall_tracker_t` was used instead of `aeron_duty_cycle_tracker_t`.
* Aligns `admin` events definition with the Java implementation, i.e. only exclude `FRAME_IN` and `FRAME_OUT` events.


mikeb01:
LGTM.

mjpt777:
LGTM. The only question I have is the naming of `GENERIC_MESSAGE`. Not sure if it can be better named.

vyazelenko:
@mjpt777 The `GENERIC_MESSAGE` was there in the Java driver first and I just copied that to the C side.

PR #1465, opened 2023-05-31
Allow the for the session-id for replication to be specified.

mikeb01:
- Set the session-id during log replication.


mjpt777:
On the whole it looks good. There are some formatting issues and commented code plus it needs a C++ client implementation.

tmontgomery (approved):

PR #1476, opened 2023-06-28
[C] Dont clean beyond rcv_position, when all subscribers are resting

reissGRVS:
Admittedly I'm not 100% sure about this change, but it seems like with the existing code the following situation can occur.

All subscribers are resting, this leads min_sub_pos = INT64_MAX, and ends up calling aeron_publication_image_clean_buffer_to with INT64_MAX - term length.
We wipe out the rest of the term, this could wipe out frames above the rcv position.
If we then had an active subscriber join again whilst rcv position is still on the same term, I think one of the two will happen (unsure which):
- end up sending unnecessary NAKs, repopulating the frames. And then as the clean position is already set to the end of the term, we wont reclean this section?
- client will get stuck waiting for frame data to be populated?

### DIFF
--- a/aeron-driver/src/main/c/aeron_publication_image.c
+++ b/aeron-driver/src/main/c/aeron_publication_image.c
@@ -385,7 +385,7 @@ void aeron_publication_image_track_rebuild(aeron_publication_image_t *image, int
         }
 
         const int64_t rcv_position = *image->rcv_pos_position.value_addr;
-        const int64_t consumed_position = rcv_position < min_sub_pos ? rcv_position : min_sub_pos;
+        const int64_t consumed_position = INT64_MAX == min_sub_pos ? rcv_position : min_sub_pos;
         const int64_t rebuild_position = rcv_position > max_sub_pos ? rcv_position : max_sub_pos;
 
         bool loss_found = false;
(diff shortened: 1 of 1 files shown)


aeron-driver/src/main/c/aeron_publication_image.c
```diff
@@ -384,8 +384,9 @@ void aeron_publication_image_track_rebuild(aeron_publication_image_t *image, int
             }
         }
 
-        const int64_t rebuild_position = *image->rcv_pos_position.value_addr > max_sub_pos ?
-            *image->rcv_pos_position.value_addr : max_sub_pos;
+        const int64_t rcv_position = *image->rcv_pos_position.value_addr;
+        const int64_t consumed_position = rcv_position < min_sub_pos ? rcv_position : min_sub_pos;
```
reissGRVS, line 388: Maybe this should be
```suggestion
        const int64_t consumed_position = INT64_MAX == min_sub_pos ? rcv_position : min_sub_pos;
```
So that if we have a valid min sub pos and its beyond the rcv position we use it instead

mikeb01:
We're currently looking at this one.  There may be some other subtle issues around this as well.

reissGRVS:
#1484 corrects this now I believe

PR #1492, opened 2023-08-15
[C] Only count active subscriptions in is drained check

reissGRVS:
If an image is in the draining state it does not call aeron_publication_image_check_untethered_subscriptions, so all subscriptions maintain their state forever.
If one of the subscriptions is in the linger state it will never progress to resting if it is completely blocked.
Therefore the aeron_publication_image_is_drained here https://github.com/real-logic/aeron/blob/master/aeron-driver/src/main/c/aeron_publication_image.c#L917 will never return true, and the image will never progress to linger or done.

### DIFF
--- a/aeron-driver/src/main/c/aeron_publication_image.h
+++ b/aeron-driver/src/main/c/aeron_publication_image.h
@@ -260,7 +260,7 @@ inline bool aeron_publication_image_is_drained(aeron_publication_image_t *image)
     {
         aeron_tetherable_position_t *tetherable_position = &image->conductor_fields.subscribable.array[i];
 
-        if (AERON_SUBSCRIPTION_TETHER_RESTING != tetherable_position->state)
+        if (AERON_SUBSCRIPTION_TETHER_ACTIVE = tetherable_position->state)
         {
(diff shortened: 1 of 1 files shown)


aeron-driver/src/main/c/aeron_publication_image.h
```diff
@@ -260,7 +260,7 @@ inline bool aeron_publication_image_is_drained(aeron_publication_image_t *image)
     {
         aeron_tetherable_position_t *tetherable_position = &image->conductor_fields.subscribable.array[i];
 
-        if (AERON_SUBSCRIPTION_TETHER_RESTING != tetherable_position->state)
+        if (AERON_SUBSCRIPTION_TETHER_ACTIVE = tetherable_position->state)
```
reissGRVS:
```suggestion
        if (AERON_SUBSCRIPTION_TETHER_ACTIVE == tetherable_position->state)
```

reissGRVS:
A very small edge case, maybe there is a better way to solve this.

mikeb01:
To be considered drained we also need to wait on subscription positions that are in the LINGER state.

PR #1495, opened 2023-08-28
Add ArchiveTool mark-valid, mark-invalid commands.

ethanf:
These provide a way to patch the catalog in an emergency.

Added the following ArchiveTool commands:

```
mark-invalid <recordingId>: marks a recording as invalid.

mark-valid <recordingId>: marks a previously invalidated recording as valid.
```


tmontgomery (approved):
LGTM. Will wait for checks.

PR #1496, opened 2023-08-29
[C] Separate out management of logbuffers to allow for preallocation and reuse

reissGRVS:
Re-work of https://github.com/real-logic/aeron/pull/1424

**Motivation** of this change is to reduce spikes in conductor_do_work cycles caused by slow mmap or munmap calls.
This is to allow for creation of new publications, subscriptions without affecting status message updates.
The benefit of this is that it reduces/removes the impact of new logbuffer allocations on the latency and throughput of high throughput streams.

I've tried to achieve this through a few methods:
* Configure pre allocated pools - allows multiple pools of different term-length logbuffers to be allocated at startup. A minimum reserve can also be configured to serve bursts of allocations after the initial capacity has been used.
* Reuse of logbuffers - upon a logbuffer being freed it is offered back to the pool, if the pool is not already at capacity the log manager will wipe this buffer and allow for it to be reused. Reducing mmap/munmap calls for short lived buffers.
* Dedicated thread for logbuffer management - moves the cost of prefaulting buffers to a separate thread. In combination with the minimum reserve this can help apps achieve similar benefits, without a large unused initial capacity.


reissGRVS:
Not sure exactly what's causing the test to die on windows. I don't have easy access to a windows environment to test in.

tmontgomery:
When reusing logbuffers, how do you plan to keep old clients accessing those buffers from writing/reading those buffers on new streams? One of the "advantages" of the current scheme is how Linux (& Windows) will handle the old file. When reusing, this becomes a little more nuanced. If it is handled, let us know how. Reviewing here, I need clarification on how it can be handled.

Windows and files are a combination that is annoying to handle with how mapping is done. When changing the lifetime of a file, it can quickly come down to how things are kept open. It's been annoying to get semi-right (very very "semi"). We've also been having some weird issues with Windows on GitHub actions. That doesn't explain the RawLogManagerTest failure. That is likely to be something else.

reissGRVS:
@tmontgomery I was under the impression that the aeron driver will only free a buffer when all of its publishers or subscribers have stopped writing to/reading from it. 
If I'm wrong about this then I don't see an obvious safe way to do buffer reuse, I'd remove that but leave the rest of the logic.

tmontgomery:
The Aeron client may write into publications or read from images if it takes a long pause causing the client to timeout publications to be closed and images to go unavailable. For how buffers work without reuse, this is somewhat "handled" on Linux (Windows is less clean, but is handled slightly differently). But with reuse, this would not be the case as the publication might now be an image, or write into some other part of the buffers, read from something that is not valid, etc.

I agree. Reuse would be good. But unless we add a means to coordinate all clients freeing a buffer and acknowledging it is not in use, reuse has a lot of edges to worry about.

reissGRVS:
I guess for this use case simply adding another client to driver event for confirming that a publication or image is discarded would do the trick.
In the event of client timeouts you can just munmap instead of returning to the pool.
I might have a play around with this if I find time

tmontgomery:
Client timeout is only one option. It may not timeout, just pause at the most inconvenient time. We've discussed in the past making it more of a back and forth to free quicker. But it does still need to handle a timeout case. Which in this area would just be an munmap and close.

I think that client close might be OK to reuse. But that might be the only real clean case. Which is not that useful.

reissGRVS:
Going to move this to draft for now, will put it back up once I either get rid of the reuse stuff entirely, or implement safe reuse for enough cases that it'd be beneficial.

Issue #1498, opened 2023-09-01
MDC subscriber ephemeral port increased network traffic


m4ce:
Been playing with Aeron for the last couple of weeks and I believe the MDC subscription could be improved when using a random ephemeral port and multiple MDC subscribers (on the same host) receiving data from the same MDC publisher.

Take the following example:

Publisher:
- channel: aeron:udp?control-mode=dynamic|control=192.168.0.100:9000 | streamId: 1

Subscribers (same host):
- channel: aeron:udp?endpoint=192.168.0.101:10000|control=192.168.0.100:9000 | streamId: 1
- channel: aeron:udp?endpoint=192.168.0.101:10000|control=192.168.0.100:9000 | streamId: 1

In the above case, the media driver will bind to UDP 127.0.0.1:10000 which is where the MDC publisher will send data too. The second subscription will work too as it seems Aeron understands there's already a socket bound to that ip/port and just reuse it.

However, reality is that most people won't probably feel like managing the port on the subscriber's side when dealing with MDC, which is likely the reason why you added support for a random ephemeral port.

Publisher:
- channel: aeron:udp?control-mode=dynamic|control=192.168.0.100:9000 | streamId: 1

Subscribers (same host):
- channel: aeron:udp?endpoint=192.168.0.101:0|control=192.168.0.100:9000 | streamId: 1
- channel: aeron:udp?endpoint=192.168.0.101:0|control=192.168.0.100:9000 | streamId: 1

In the above case, Aeron media driver will actually create two UDP sockets on two different randomly selected ephemeral ports. I am not sure if this makes sense? What happens now is that the MDC publisher will now have to send data to the same host **twice**, increasing network traffic for no apparent reason.

IMHO, the media driver should re-use the existing UDP socket that was created by the first subscriber, on the same randomly selected ephemeral port. Or at least there should be a simple URI option to support that, like a reuse port type of thing.

Look forward to hearing your feedback on the above.

m4ce:
I've also tried with tags, which seems like it does achieve some of what I outlined above but operationally is not great.

This works:

Publisher:
- channel: aeron:udp?control-mode=dynamic|control=192.168.0.100:9000 | streamId: 1

Subscribers (same host):
- channel: aeron:udp?endpoint=192.168.0.101:0|control=192.168.0.100:9000|tags=1001 | streamId: 1
- channel: aeron:udp?tags=1001 | streamId: 1

However, it fails if I run the two subscribers as follows:

[...]

PR #1500, opened 2023-09-07
Support specifying endpoints and tags in the same URI

mikeb01:
- Allow a channel URI to contain both tags and endpoint parameters even if there is already a send or receive channel endpoint for the tag.
- For existing endpoints, check that the control and endpoint values are the same and reject if they do not match.
- Potential Fix for #1498.

### DIFF
--- a/aeron-driver/src/main/c/media/aeron_udp_channel.h
+++ b/aeron-driver/src/main/c/media/aeron_udp_channel.h
@@ -71,8 +71,8 @@ inline bool aeron_udp_channel_is_wildcard(aeron_udp_channel_t *channel)
 
 inline int aeron_udp_channel_endpoints_match(aeron_udp_channel_t *channel, aeron_udp_channel_t *other, bool *result)
 {
-    bool cmp;
-    int rc;
+    bool cmp = false;
+    int rc = 0;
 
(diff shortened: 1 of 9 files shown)


aeron-driver/src/main/c/media/aeron_udp_channel.h
```diff
@@ -68,6 +69,35 @@ inline bool aeron_udp_channel_is_wildcard(aeron_udp_channel_t *channel)
         aeron_is_wildcard_addr(&channel->local_data) && aeron_is_wildcard_port(&channel->local_data);
 }
 
+inline int aeron_udp_channel_endpoints_match(aeron_udp_channel_t *channel, aeron_udp_channel_t *other, bool *result)
+{
+    bool cmp;
```
mjpt777, line 74: Best to initialise.

aeron-driver/src/main/c/media/aeron_udp_channel.h
```diff
@@ -68,6 +69,35 @@ inline bool aeron_udp_channel_is_wildcard(aeron_udp_channel_t *channel)
...
+    int rc;
```
mjpt777, line 75: Declare and assign so not in uninitialised state.

mjpt777 (approved):
Looks good.

PR #1510, opened 2023-10-02
[C] Refactor aeron_counters_reader_counter_type_id().

wojciech-adaptive:
I find accessing a field with a different field quite confusing.


mjpt777:
Not really a refactor, more a good fix ;-)

PR #1519, opened 2023-10-16
Remove dynamic join

mikeb01:
* Remove the implementation for Dynamic Join that was deprecated in 1.41.


aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModuleAgent.java
```diff
@@ -1014,10 +1001,6 @@ void onSnapshotRecordingQuery(final long correlationId, final int requestMemberI
 
     void onSnapshotRecordings(final long correlationId, final SnapshotRecordingsDecoder decoder)
```
mjpt777, line 1002: Can this method be removed?

mjpt777 (approved):
Looks good.

aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModuleAgent.java
mikeb01: Yes, there is still more to be removed as well.

PR #1522, opened 2023-10-25
[Java] Log an error instead of throwing an exception if the Clustered…

mikeb01:
…ServiceAgent receives a close for an unknown session.

### DIFF
--- a/aeron-cluster/src/main/java/io/aeron/cluster/service/ClusteredServiceAgent.java
+++ b/aeron-cluster/src/main/java/io/aeron/cluster/service/ClusteredServiceAgent.java
@@ -486,22 +486,15 @@ final class ClusteredServiceAgent extends ClusteredServiceAgentRhsPadding implem
     {
         this.logPosition = logPosition;
         clusterTime = timestamp;
-        final ContainerClientSession session = sessionByIdMap.remove(clusterSessionId);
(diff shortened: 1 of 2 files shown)


tmontgomery (approved):

aeron-cluster/src/main/java/io/aeron/cluster/service/ClusteredServiceAgent.java
```diff
@@ -486,22 +486,15 @@ void onSessionClose(
     {
         this.logPosition = logPosition;
         clusterTime = timestamp;
-        final ContainerClientSession session = sessionByIdMap.remove(clusterSessionId);
+        final ContainerClientSession session = removeSession(clusterSessionId);
 
         if (null == session)
         {
-            throw new ClusterException(
+            ctx.errorLog().record(new ClusterException(
```
mjpt777, line 493: What about updating the error counter?

mikeb01: Changed to counted error handler.

mjpt777:
Extracted to this commit https://github.com/real-logic/aeron/commit/752cc9cc0a9eddf6e4e34c385fdeeb954b2a645d

PR #1523, opened 2023-10-31
Stricter validation on control addresses

mikeb01:
This tidies up some of the validation around control-mode and control address.  It now requires that `control-mode=dynamic` have a control address and also be explicitly stated.  Just specifying a control address does not imply `control-mode=dynamic`.  Now specifying a `control` address requires a `control-mode` or an `endpoint`.  If an endpoint is specified then this is a plain channel with an explicit control address (i.e. not a wildcard).

It also validates how control-mode can be specified with respect to the use of tags.  I.e. the control-mode can be omitted or must match the control-mode of any existing channel with the same tag.

### DIFF
--- a/aeron-driver/src/main/c/aeron_driver_conductor.c
+++ b/aeron-driver/src/main/c/aeron_driver_conductor.c
@@ -379,6 +379,33 @@ static inline int aeron_driver_conductor_validate_endpoint_for_publication(aeron
     return 0;
 }
 
+static inline int aeron_driver_conductor_validate_control_for_publication(aeron_udp_channel_t *udp_channel)
+{
+    if (udp_channel->is_dynamic_control_mode && !udp_channel->has_explicit_control)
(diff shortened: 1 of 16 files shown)


aeron-driver/src/main/c/aeron_driver_conductor.c
```diff
@@ -2038,25 +2103,11 @@ aeron_send_channel_endpoint_t *aeron_driver_conductor_get_or_add_send_channel_en
 
     if (NULL != endpoint)
     {
-        bool endpoints_match = false;
```
mjpt777, line 2041: Minor point on naming. `endpoints_match` -> `has_matching_endpoints`.

mjpt777 (approved):

tmontgomery (approved):

tmontgomery:
LGTM

PR #1524, opened 2023-11-01
[C] Fix client image reference counting.

wojciech-adaptive:
Noticed that image reference count in the C client sometimes had funny values around 4294967296. Turns out that our macro with -1 literal gets compiled to:

```
58	    AERON_GET_AND_ADD_INT64(res, var, -1);
   0x000000000000c74b <+39>:	b8 ff ff ff ff	mov    $0xffffffff,%eax
   0x000000000000c750 <+44>:	f0 48 0f c1 45 e8	lock xadd %rax,-0x18(%rbp)
```

So instead of adding -1, we're adding 4294967295 (0xffffffff).

With this change, it looks like this:

```
51	    AERON_GET_AND_ADD_INT64(res, var, -1);
   0x000000000000c6c9 <+39>:	48 c7 c0 ff ff ff ff	mov    $0xffffffffffffffff,%rax
   0x000000000000c6d0 <+46>:	f0 48 0f c1 45 e8	lock xadd %rax,-0x18(%rbp)
```


mikeb01:
This is a good spot and we should implement the change.  We should also be more careful with our literals in the C code.  In this case we should also be changing the call site to:
```
AERON_GET_AND_ADD_INT64(res, var, INT64_C(-1));
```

mikeb01:
I'm wondering we could use `_Generic` to introduce some type safety here as well.

tmontgomery (approved):

PR #1532, opened 2023-11-14
[Java] Introduce SnapshotDurationTracker

eliquinox:
This PR introduces a set of counters that enable monitoring of snapshot duration within Cluster's Consensus Module:
* `AeronCounters#CLUSTER_CONSENSUS_MODULE_MAX_SNAPSHOT_DURATION_TYPE_ID` to track maximum snapshot duration.
* `AeronCounters#CLUSTER_CONSENSUS_MODULE_SNAPSHOT_DURATION_THRESHOLD_EXCEEDED_TYPE_ID` to track number of times a predefined threshold value (see below) has been exceeded.

Threshold value can be set via `aeron.cluster.consensus.module.snapshot.threshold` property or via `io.aeron.cluster.ConsensusModule.Context#snapshotDurationThresholdNs(long)`.

### DIFF
--- a/aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModule.java
+++ b/aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModule.java
@@ -752,6 +752,18 @@ public final class ConsensusModule implements AutoCloseable
         public static final long CYCLE_THRESHOLD_DEFAULT_NS = TimeUnit.MILLISECONDS.toNanos(1000);
 
         /**
+         * Property name for threshold value, which is used for tracking total snapshot duration breaches.
+         */
+        public static final String TOTAL_SNAPSHOT_DURATION_THRESHOLD_PROP_NAME =
+            "aeron.cluster.total.snapshot.threshold";
+
+        /**
+         * Default threshold value, which is used for tracking total snapshot duration breaches.
+         */
+        public static final long TOTAL_SNAPSHOT_DURATION_THRESHOLD_DEFAULT_NS =
+            TimeUnit.MILLISECONDS.toNanos(1000);
+
+        /**
          * Default timeout a leader will wait on getting termination ACKs from followers.
          */
         public static final long TERMINATION_TIMEOUT_DEFAULT_NS = TimeUnit.SECONDS.toNanos(10);
@@ -1101,6 +1113,18 @@ public final class ConsensusModule implements AutoCloseable
         }
 
         /**
+         * Get threshold value, which is used for monitoring total snapshot duration breaches of its predefined
+         * threshold.
+         *
+         * @return threshold value in nanoseconds.
+         */
+        public static long totalSnapshotDurationThresholdNs()
+        {
+            return getDurationInNanos(TOTAL_SNAPSHOT_DURATION_THRESHOLD_PROP_NAME,
+                TOTAL_SNAPSHOT_DURATION_THRESHOLD_DEFAULT_NS);
+        }
+
+        /**
          * Size in bytes of the error buffer in the mark file.
          *
          * @return length of error buffer in bytes.
@@ -1392,6 +1416,7 @@ public final class ConsensusModule implements AutoCloseable
         private long electionStatusIntervalNs = Configuration.electionStatusIntervalNs();
         private long terminationTimeoutNs = Configuration.terminationTimeoutNs();
         private long cycleThresholdNs = Configuration.cycleThresholdNs();
+        private long totalSnapshotDurationThresholdNs = Configuration.totalSnapshotDurationThresholdNs();
 
         private String agentRoleName = Configuration.agentRoleName();
         private ThreadFactory threadFactory;
@@ -1425,6 +1450,7 @@ public final class ConsensusModule implements AutoCloseable
         private LogPublisher logPublisher;
         private EgressPublisher egressPublisher;
         private DutyCycleTracker dutyCycleTracker;
+        private SnapshotDurationTracker totalSnapshotDurationTracker;
         private AppVersionValidator appVersionValidator;
         private boolean isLogMdc;
         private boolean useAgentInvoker = false;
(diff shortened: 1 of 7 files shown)


aeron-cluster/src/main/java/io/aeron/cluster/service/SnapshotDurationTracker.java
```diff
@@ -0,0 +1,101 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.aeron.cluster.service;
+
+import org.agrona.concurrent.status.AtomicCounter;
+
+// TODO: check against duty cycle tracker
+// TODO: worry about false sharing?
```
eliquinox, line 21: To be addressed

aeron-system-tests/src/test/java/io/aeron/cluster/ClusterTest.java
```diff
@@ -1922,6 +1914,57 @@ void shouldTakeASnapshotAfterReceivingAdminRequestOfTypeSnapshot()
         cluster.awaitNeutralControlToggle(leader);
     }
 
+    @Test
+    @InterruptAfter(20)
+    void shouldTrackSnapshotDuration()
+    {
+        cluster = aCluster()
+            .withServiceSupplier((i) -> new TestNode.TestService[]{
```
mjpt777, line 1922: Braces should be matched.

aeron-cluster/src/main/java/io/aeron/cluster/service/SnapshotDurationTracker.java
mjpt777: Snapshots are relatively infrequent so false sharing is not a concern.

mjpt777:
In general LGTM.

aeron-cluster/src/main/java/io/aeron/cluster/service/SnapshotDurationTracker.java
```diff
@@ -0,0 +1,98 @@
...
+    {
+        return snapshotDurationThresholdExceededCount;
+    }
+
+    /**
+     * Called when snapshotting has started.
+     *
+     * @param timeNanos snapshot start time in nanoseconds.
+     */
+    public void onSnapshotStart(final long timeNanos)
```
mjpt777, line 74: onSnapshotBegin

aeron-cluster/src/main/java/io/aeron/cluster/service/SnapshotDurationTracker.java
```diff
@@ -0,0 +1,98 @@
...
+    {
+        snapshotStartTimeNs = timeNanos;
+    }
+
+    /**
+     * Called when snapshot has been taken.
+     *
+     * @param timeNanos snapshot end time in nanoseconds.
+     */
+    public void onSnapshotTaken(final long timeNanos)
```
mjpt777, line 84: onSnapshotEnd

vyazelenko:
@eliquinox Why not use a plain DutyCycleStallTracker which we is used in all other places?

aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModule.java
```diff
@@ -3020,6 +3064,52 @@ public DutyCycleTracker dutyCycleTracker()
...
+        {
+            return snapshotDurationThresholdNs;
+        }
+
+        /**
+         * Set snapshot duration tracker used for monitoring consensus module snapshot duration.
+         * @param snapshotDurationTracker snapshot duration tracker.
+         * @return this for fluent API.
+         */
+        public ConsensusModule.Context snapshotDurationTracker(final SnapshotDurationTracker snapshotDurationTracker)
```
vyazelenko, line 3098: Use `Context` instead of `ConsensusModule.Context` like others methods do.

aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModule.java
```diff
@@ -1717,6 +1741,26 @@ public void conclude()
                     cycleThresholdNs);
             }
 
+            if (null == snapshotDurationTracker)
+            {
+                snapshotDurationTracker = new SnapshotDurationTracker(
+                    ClusterCounters.allocate(
+                        aeron,
+                        buffer,
+                        "Consensus module max snapshot duration in ns",
```
vyazelenko, line 1750: The naming is misleading as it does not measure the `ConsensusModule` snapshot time but instead measures the total snapshot time.

aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModule.java
```diff
@@ -1717,6 +1741,26 @@ public void conclude()
...
+                        AeronCounters.CLUSTER_CONSENSUS_MODULE_MAX_SNAPSHOT_DURATION_TYPE_ID,
+                        clusterId),
+                    ClusterCounters.allocate(
+                        aeron,
+                        buffer,
+                        "Consensus module max snapshot duration exceeded count: threshold=" +
```
vyazelenko, line 1756: Naming.

aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModuleAgent.java
```diff
@@ -1275,6 +1278,8 @@ void onServiceAck(
                     {
                         ClusterControl.ToggleState.reset(controlToggle);
                     }
+
+                    snapshotDurationTracker.onSnapshotEnd(clusterClock.timeNanos());
```
vyazelenko, line 1282: This will miss the case when snapshot is taken before the Cluster termination, i.e. in which case the snapshot time will not be recorded. Maybe move it into the `takeSnapshot` method.

vyazelenko (changes requested):

aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModule.java
vyazelenko: So both the names and static final constants should reflect that. Also the system property for the entire thing should just use `aeron.cluster.` prefix and the ones for CM would then use `aeron.cluster.consensus.module.` prefix.

vyazelenko (changes requested):

vyazelenko (approved):

PR #1534, opened 2023-11-16
Response Channels

mikeb01:
This adds the ability to create pairs of request/response channels.  Features include:
- New control-mode (response) to configure the publication and subscription of the response side of the channel pair.
- The ability to pass a response-correlation-id to indicate the association between a subscription and a publication for sending on the request channel and an image and a publication for sending on the response channel.
- Adding the response-correlation-id to the criteria used to match publication when creating concurrent publications to prevent accidental sharing of request/response pairs.
- Validation
- The ability for multiple publications to share an underlying SendChannelEndpoint.


aeron-client/src/main/java/io/aeron/protocol/ResponseSetupFlyweight.java
```diff
@@ -0,0 +1,144 @@
...
+     */
+    public ResponseSetupFlyweight streamId(final int streamId)
+    {
+        putInt(STREAM_ID_FIELD_OFFSET, streamId, LITTLE_ENDIAN);
+
+        return this;
+    }
+
+    /**
+     * Get stream id field.
```
ethanf, line 123: Copy/pasted comment here and below btw

aeron-driver/src/main/java/io/aeron/driver/PublicationImage.java
```diff
@@ -690,8 +695,14 @@ int sendPendingStatusMessage(final long nowNs)
     {
         int workCount = 0;
         final long changeNumber = endSmChange;
+        final boolean hasSmTimedOut = (timeOfLastSmNs + smTimeoutNs) - nowNs < 0;
 
-        if (changeNumber != lastSmChangeNumber || (timeOfLastSmNs + smTimeoutNs) - nowNs < 0)
+        if (null != responseSessionId && hasSmTimedOut)
```
vyazelenko, line 700: Double read of the volatile field `responseSessionId` (inside of the `if`). Also the order of the checks can be reversed to avoid reading that field id the `hasSmTimedOut == false`.

PR #1535, opened 2023-11-22
[C/C++] Add separate component versions (driver, client, archive client). Client appends its version info to the heartbeat timestamp counter upon initialization.

vyazelenko:
* Separate version for driver, client and archive client.
* Aeron client conductor appends client version info to the heartbeat timestamp counter label.


aeron-client/src/main/cpp/ClientConductor.h
```diff
@@ -32,6 +32,7 @@
 #include "LogBuffers.h"
 #include "HeartbeatTimestamp.h"
 #include "util/Export.h"
+#include "AeronVersion.h"
```
mikeb01, line 35: Do we need to implement this in the wrapper as well?

vyazelenko: @mikeb01 We can do for symmetry but the wrapper is delegating to the C client conductor which means that the latter is going to use the `aeron_version.h` anyhow.

PR #1547, opened 2024-01-23
[Java] Re-assemble header for fragmented messages.

vyazelenko:
Re-assemble header of the fragmented message so that it represents a complete message, i.e.:
* Compute `frameLength` for the entire message.
* Set the `BEGIN_FRAG_FLAG` on the `flags` to mark message as unfragmented.
* Preserve `termOffset`, `termId`, `sessionId`, `streamId` and `reserevedValue` of the first fragmented header.

### DIFF
--- a/aeron-client/src/main/java/io/aeron/logbuffer/Header.java
+++ b/aeron-client/src/main/java/io/aeron/logbuffer/Header.java
@@ -100,10 +100,12 @@ public final class Header
      * Set the number of times to left shift the term count to multiply by term length.
      *
      * @param positionBitsToShift number of times to left shift the term count to multiply by term length.
+     * @return this for a fluent API.
      */
-    public void positionBitsToShift(final int positionBitsToShift)
+    public Header positionBitsToShift(final int positionBitsToShift)
     {
         this.positionBitsToShift = positionBitsToShift;
+        return this;
     }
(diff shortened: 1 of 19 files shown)


aeron-client/src/main/java/io/aeron/logbuffer/Header.java
```diff
@@ -95,6 +96,16 @@ public int positionBitsToShift()
         return positionBitsToShift;
     }
 
+    /**
+     * Set the number of times to left shift the term count to multiply by term length.
+     *
+     * @param positionBitsToShift number of times to left shift the term count to multiply by term length.
+     */
+    public void positionBitsToShift(final int positionBitsToShift)
```
mjpt777, line 104: These methods could be a fluent API.

vyazelenko: Should the existing methods `offset(int)` and `buffer(DirectBuffer)` also be changed for fluent API?

mjpt777: Probably good for consistency.

vyazelenko: Done

PR #1548, opened 2024-01-29
Await close responses back from the client conductor in the wrapper.

mikeb01:
To prevent memory leaks if the subscription and client close in quick succession.


aeron-client/src/test/cpp_wrapper/SystemTest.cpp
```diff
@@ -192,3 +192,44 @@ TEST_F(SystemTest, shouldAddRemoveCloseHandler)
...
+
+        std::shared_ptr<Aeron> aeron = Aeron::connect(ctx);
+        int64_t i = aeron->addSubscription("aeron:ipc", 1000);
+        std::shared_ptr<Subscription> subscription;
+        do
+        {
+            subscription = aeron->findSubscription(i);
+        }
+        while (nullptr == subscription);
+    }
```
mjpt777, line 214: Why the extra braces?

PR #1549, opened 2024-02-05
Prevent sanitiser errors when shutting down subscription and client in quick succession

mikeb01:
Use the C client close handler to ensure that the client conductor thread drains before destructing the Aeron resources in the C++ wrapper.


aeron-client/src/test/cpp_wrapper/SystemTest.cpp
```diff
@@ -192,3 +192,44 @@ TEST_F(SystemTest, shouldAddRemoveCloseHandler)
     EXPECT_EQ(1, closeCount1);
     EXPECT_EQ(0, closeCount2);
 }
+
+//
+// These tests will fail with the sanitizer if not implemented correctly.
+//
+
+TEST_F(SystemTest, shouldFreeSubscritionDataCorrectly)
```
wojciech-adaptive, line 200: Typo in the name, same in the one below.

PR #1553, opened 2024-02-07
Add C driver logging for send_nak_message and resend

nbradac:
WIP

### DIFF
--- a/aeron-driver/src/main/c/agent/aeron_driver_agent.c
+++ b/aeron-driver/src/main/c/agent/aeron_driver_agent.c
@@ -131,6 +131,7 @@ static aeron_driver_agent_log_event_t log_events[] =
         { "ADD_DYNAMIC_DISSECTOR",                AERON_DRIVER_AGENT_EVENT_TYPE_OTHER,   false },
         { "DYNAMIC_DISSECTOR_EVENT",              AERON_DRIVER_AGENT_EVENT_TYPE_OTHER,   false },
         { "SEND_NAK_MESSAGE",                     AERON_DRIVER_AGENT_EVENT_TYPE_OTHER,   false },
+        { "RESEND",                               AERON_DRIVER_AGENT_EVENT_TYPE_OTHER,   false },
     };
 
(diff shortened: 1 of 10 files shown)


aeron-driver/src/main/c/agent/aeron_driver_agent.h
```diff
@@ -164,6 +166,19 @@ typedef struct aeron_driver_agent_flow_control_receiver_change_log_header_stct
...
 
+typedef struct aeron_driver_agent_send_nak_message_header_stct
+{
+    int64_t time_ns;
+    int32_t session_id;
+    int32_t stream_id;
+    int32_t term_id;
+    int32_t term_offset;
+    int32_t nak_length;
+    int32_t address_length;
```
mikeb01, line 177: To keep the structure simpler, I would just use a `struct sockaddr_storage` here instead of tracking the length to handle IPv4 and IPv6.

aeron-driver/src/main/c/agent/aeron_driver_agent.c
```diff
@@ -1949,6 +2005,50 @@ void aeron_driver_agent_log_dissector(int32_t msg_type_id, const void *message,
             break;
         }
 
+        case AERON_DRIVER_EVENT_SEND_NAK_MESSAGE:
+        {
+            aeron_driver_agent_send_nak_message_header_t *hdr = (aeron_driver_agent_send_nak_message_header_t *)message;
+
+            char *address_str = "unknown-address";
+            char address_buf[INET6_ADDRSTRLEN] = { 0 };
+            uint8_t *address_ptr = (uint8_t *)message + sizeof(aeron_driver_agent_send_nak_message_header_t);
```
mikeb01, line 2014: I think the code will be simpler following the approach in `AERON_DRIVER_EVENT_NAME_RESOLUTION_NEIGHBOR_REMOVED`.  If the log structure contains a `struct sockaddr_storage` then we can use `aeron_format_source_identity` to convert the address to a string.  This will handle the address type and will also render the port, which is relevant here.

Issue #1125, opened 2021-01-18
QoS configuration option


oktal:
Hi,

Currently, Aeron does not provide a way to configure the QoS of the underlying sockets created by the media driver.

This is something users might want to have control to, depending on the workload and the underlying setup (cross-DC communication, ...)

I think this can be implemented and supported:

* At the media driver level through a configuration parameter (1-byte integer `AERON_IP_TOS`)
* At the channel level as a param of the uri
* Both

What do you suggest ?

I can work on an implementation on my side and then submit a PR, I just need to know what criteria my PR should meet for you guys to merge it.

mjpt777:
We would be happy to consider a PR it it applies to both Java and C driver and comes with tests.

It should be configurable via system property, driver context, and URI param. We can discuss naming on a PR.

mrk-andreev:
Hi @mjpt777 ,

Is this feature still of interest to the community? Would it make sense to refine @pveentjer 's [results](https://github.com/pveentjer/aeron/commits/feature/tos/) and present them in a PR? 

I see that development of this feature is on hold but these changes look interesting to me as a developer.

PR #1555, opened 2024-02-11
[WIP] Add support for IP_TOS (Type Of Service)

pveentjer:
Adds support for IP_TOS.

Implements the following ticket:

https://github.com/real-logic/aeron/issues/1125

What is missing:
- probably some config logic. Still in the process of getting a proper understanding.
- implement TOS in the C driver.

There are some tests, but not sure if this is sufficient.

And perhaps rename 'ToS' to TypeOfService even though the socket property under Linux is called IP_TOS?


mikeb01:
This PR has been open an WIP for some time, do you intend to complete work on it?

PR #1559, opened 2024-02-16
[C] add response_client/server sample apps

nbradac:


aeron-samples/src/main/c/response/response_client.c
```diff
@@ -0,0 +1,371 @@
...
+        {
+            fprintf(stderr, "aeron_subscription_constants: %s\n", aeron_errmsg());
+            goto cleanup;
+        }
+
+        subscriber_registration_id = constants.registration_id;
+    }
+
+    {
+        char _channel_buf[256] = { 0 };
```
mikeb01, line 276: You can use `AERON_MAX_PATH` constant for URI related buffer sizes.

PR #1560, opened 2024-02-18
Use response channels in the Archive Client

mikeb01:
Allow response channels to be used for control, replay and replication channels when using the Archive client over a remote connection.
* Support the use of a randomly generated single use replay token to make replays over response channels work correctly.
* Using response channels with replication does not work with live destinations or tagged replication.


aeron-archive/src/main/java/io/aeron/archive/ArchiveConductor.java
```diff
@@ -132,6 +136,15 @@ abstract class ArchiveConductor
         dutyCycleTracker = ctx.conductorDutyCycleTracker();
         cachedEpochClock.update(epochClock.time());
 
+        try
+        {
+            secureRandom = SecureRandom.getInstanceStrong();
```
pcdv, line 141: Beware of using getInstanceStrong() : it can block for several seconds on some systems and deplete the entropy.

Not sure if there is a need for strong security here or if this is simply following the recommendation of some static analysis tool. I've been bitten in the past by doing this :)

mikeb01: It is being used to generate an authentication token, so I've picked the strongest generator as a conservative option.

mikeb01: I've moved the `getInstanceStrong()` usage to be there to construct a seed for the PRNG.

PR #1564, opened 2024-02-21
[Java] Async DNS resolution.

vyazelenko:
This PR implements asynchronous DNS resolution in the Java MediaDriver including:
* Client commands: `ADD_RCV_DESTINATION` (spy and network), `REMOVE_RCV_DESTINATION`, `ADD_SUBSCRIPTION` (spy and network), `ADD_PUBLICATION`, `ADD_DESTINATION` and `REMOVE_DESTINATION`.
* Address re-resolution (`onReResolveControl` and `onReResolveEndpoint`) which are triggered by the receiver and sender threads respectively.
* `ConsensusModule#connectIngress` - check for multicast address is now done without calling `NameResolver`.

What is not included:
* Initial host name resolution, i.e. `DriverConductor#onStart`.
* `DriverNameResolver` - bootstrap neighbor resolution and duty cycle.

### DIFF
--- a/aeron-driver/src/main/java/io/aeron/driver/DriverConductor.java
+++ b/aeron-driver/src/main/java/io/aeron/driver/DriverConductor.java
@@ -226,7 +226,7 @@ public final class DriverConductor implements Agent
         int workCount = 0;
         workCount += processTimers(nowNs);
         workCount += clientCommandAdapter.receive();
-        workCount += CommandProxy.drain(driverCmdQueue, Configuration.COMMAND_DRAIN_LIMIT, Runnable::run);
+        workCount += CommandProxy.drainQueue(driverCmdQueue, Configuration.COMMAND_DRAIN_LIMIT, Runnable::run);
         workCount += trackStreamPositions(workCount, nowNs);
         workCount += nameResolver.doWork(cachedEpochClock.time());
         workCount += freeEndOfLifeResources(ctx.resourceFreeLimit());
--- a/aeron-driver/src/main/java/io/aeron/driver/DriverConductor.java
+++ b/aeron-driver/src/main/java/io/aeron/driver/DriverConductor.java
@@ -87,6 +87,7 @@ public final class DriverConductor implements Agent
         SOCKET_SNDBUF_PARAM_NAME,
         RESPONSE_CORRELATION_ID_PARAM_NAME
     };
+
     private int nextSessionId = BitUtil.generateRandomisedId();
     private final long timerIntervalNs;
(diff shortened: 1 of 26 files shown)


aeron-driver/src/main/java/io/aeron/driver/DriverConductor.java
```diff
@@ -1373,6 +1447,67 @@ private ArrayList<SubscriberPosition> createSubscriberPositions(
         return subscriberPositions;
     }
 
+    private void executeAsyncClientTask(
```
mikeb01, line 1450: I think the naming of this function should reflect that is is performing the `UdpChannel::parse`.  I think this could be generalised as well, if the AsyncResult carried both a success result and a failure result.  Then the result will be forwarded back to the command queue to let the caller take the appropriate action based on the success or failure rather than routing it directly back to the client from here.  With this approach we could probably fold the functionality of this method with `executeAsyncTask`.

vyazelenko: @mikeb01 I went this route originally and had `UdpChannel` parsing be passed as `Supplier<UdpChannel>` parameter. I then realized that all client tasks always parse the channel and decided to avoid allocating a `Supplier` instance just for that.
_Btw, the `AsyncResult` already carries both ok and error result._

On your second point: the reason why the `executeAsyncClientTask` is not calling `executeAsyncTask` is, because we want to limit number of client actions to one in-flight but allow any number of other async tasks. To that end the `executeAsyncClientTask` is setting `asyncClientCommandInFlight = true` and is responsible for resetting this flag when the action completes.

aeron-driver/src/main/java/io/aeron/driver/Receiver.java
```diff
@@ -46,7 +47,7 @@ public final class Receiver implements Agent
     private final long reResolutionCheckIntervalNs;
     private long reResolutionDeadlineNs;
     private final DataTransportPoller dataTransportPoller;
-    private final OneToOneConcurrentArrayQueue<Runnable> commandQueue;
+    private final ManyToOneConcurrentLinkedQueue<Runnable> commandQueue;
```
mikeb01, line 50: I think that we should change the queues between the conductor, receiver, sender back to array queues.  If we are concerned about filling these up, we can check the available capacity before attempting to drain the driverCmdQueue in the conductor.  Keep these as linked queues makes the C implementation far more complex.

vyazelenko: Restored original queues for sender and receiver.

PR #1566, opened 2024-02-24
[C] async DNS res for the C media driver

nbradac:


aeron-client/src/main/c/concurrent/aeron_blocking_linked_queue.c
```diff
@@ -0,0 +1,111 @@
...
+ * limitations under the License.
+ */
+
+#include "concurrent/aeron_blocking_linked_queue.h"
+
+int aeron_blocking_linked_queue_init(aeron_blocking_linked_queue_t *queue)
+{
+    if (aeron_spsc_concurrent_linked_queue_init(&queue->spsc_queue) < 0)
+    {
+        return -1;
```
mikeb01, line 23: AERON_APPEND_ERR

aeron-client/src/main/c/concurrent/aeron_blocking_linked_queue.c
```diff
@@ -0,0 +1,111 @@
...
+int aeron_blocking_linked_queue_close(aeron_blocking_linked_queue_t *queue)
+{
+    aeron_mutex_lock(&queue->mutex);
+
+    if (0 != queue->size)
+    {
+        aeron_mutex_unlock(&queue->mutex);
+
+        // can't delete the queue until it's empty
+        return -1;
```
mikeb01, line 42: AERON_APPEND_ERR

aeron-client/src/main/c/concurrent/aeron_blocking_linked_queue.c
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    if (aeron_spsc_concurrent_linked_queue_close(&queue->spsc_queue) < 0)
+    {
+        return -1;
```
mikeb01, line 47: AERON_APPEND_ERR

aeron-client/src/main/c/concurrent/aeron_blocking_linked_queue.c
```diff
@@ -0,0 +1,111 @@
...
+{
+    int rc;
+
+    aeron_mutex_lock(&queue->mutex);
+
+    rc = aeron_spsc_concurrent_linked_queue_offer(&queue->spsc_queue, element);
+
+    if (rc == 0)
+    {
+        queue->size++;
```
mikeb01, line 68: I think this should be an atomic ordered increment to ensure that it will be visible when the poll side gets released from the condition variable.

aeron-client/src/main/c/concurrent/aeron_blocking_linked_queue.c
```diff
@@ -0,0 +1,111 @@
...
+    return rc;
+}
+
+void *aeron_blocking_linked_queue_poll(aeron_blocking_linked_queue_t *queue)
+{
+    void *element = NULL;
+
+    aeron_mutex_lock(&queue->mutex);
+
+    if (0 == queue->size)
```
mikeb01, line 83: This might need to be a while loop to cater for spurious wake ups.

aeron-client/src/main/c/concurrent/aeron_blocking_linked_queue.c
```diff
@@ -0,0 +1,111 @@
...
+        queue->size++;
+        aeron_cond_signal(&queue->cv);
+    }
+
+    aeron_mutex_unlock(&queue->mutex);
+
+    return rc;
+}
+
+void *aeron_blocking_linked_queue_poll(aeron_blocking_linked_queue_t *queue)
```
mikeb01, line 77: Because this method blocks, I think the naming convention here should be `_take` instead of `_poll`.

aeron-client/src/main/c/concurrent/aeron_blocking_linked_queue.c
```diff
@@ -0,0 +1,111 @@
...
+    aeron_mutex_lock(&queue->mutex);
+
+    if (0 == queue->size)
+    {
+        aeron_cond_wait(&queue->cv, &queue->mutex);
+    }
+
+    element = aeron_spsc_concurrent_linked_queue_poll(&queue->spsc_queue);
+
+    if (NULL != element)
```
mikeb01, line 90: If the element is null should we loop back and continue to block and wait?  This relates to the code above as well.  If we wake up (for some reason) and there isn't any data available, should we continue to block or put the responsibility onto the caller to make sure that it understands that sometimes the return value can be null and will need to handle that retry if appropriate.

aeron-client/src/main/c/concurrent/aeron_blocking_linked_queue.c
```diff
@@ -0,0 +1,111 @@
...
+    {
+        queue->size--;
```
mikeb01, line 92: I think an ordered release here as well just to be safe.

aeron-client/src/main/c/concurrent/aeron_spsc_concurrent_linked_queue.c
```diff
@@ -0,0 +1,115 @@
...
+    void *element;
+};
+
+int aeron_spsc_concurrent_linked_queue_node_create(void *element, aeron_spsc_concurrent_linked_queue_node_t **nodep)
+{
+    aeron_spsc_concurrent_linked_queue_node_t *node;
+
+    if (aeron_alloc((void **)&node, sizeof(aeron_spsc_concurrent_linked_queue_node_t)) < 0)
+    {
+        return -1;
```
mikeb01, line 31: AERON_APPEND_ERR

aeron-client/src/main/c/concurrent/aeron_spsc_concurrent_linked_queue.c
```diff
@@ -0,0 +1,115 @@
...
+    if (NULL == queue->tail->next)
+    {
+        // the queue is empty
+        aeron_spsc_concurrent_linked_queue_node_delete(queue->tail);
+
+        return 0;
+    }
+
+    // the queue is NOT empty
+    return -1;
```
mikeb01, line 75: AERON_SET_ERR

aeron-client/src/main/c/concurrent/aeron_spsc_concurrent_linked_queue.c
```diff
@@ -0,0 +1,115 @@
...
+    return 0;
+}
+
+int aeron_spsc_concurrent_linked_queue_init(aeron_spsc_concurrent_linked_queue_t *queue)
+{
+    aeron_spsc_concurrent_linked_queue_node_t *node;
+
+    if (aeron_spsc_concurrent_linked_queue_node_create(NULL, &node))
+    {
+        return -1;
```
mikeb01, line 55: AERON_APPEND_ERR

aeron-client/src/main/c/concurrent/aeron_spsc_concurrent_linked_queue.c
```diff
@@ -0,0 +1,115 @@
...
+    return -1;
+}
+
+int aeron_spsc_concurrent_linked_queue_offer(aeron_spsc_concurrent_linked_queue_t *queue, void *element)
+{
+    aeron_spsc_concurrent_linked_queue_node_t *node, *prev;
+
+    if (aeron_spsc_concurrent_linked_queue_node_create(element, &node))
+    {
+        return -1;
```
mikeb01, line 84: AERON_APPEND_ERR

aeron-client/src/main/c/concurrent/aeron_spsc_concurrent_linked_queue.h
```diff
@@ -0,0 +1,36 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#ifndef AERON_SPSC_CONCURRENT_LINKED_QUEUE_H
+#define AERON_SPSC_CONCURRENT_LINKED_QUEUE_H
+
+typedef struct aeron_spsc_concurrent_linked_queue_node_stct aeron_spsc_concurrent_linked_queue_node_t;
+
+typedef struct aeron_spsc_concurrent_linked_queue_stct
```
mikeb01, line 21: Noticing that a number of the operations on this queue aren't thread safe and rather than fix those, I think perhaps we just have a simple linked list and then rely on the locking in the blocking queue to ensure thread safety.

aeron-client/src/main/c/concurrent/aeron_spsc_concurrent_linked_queue.c
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    prev = queue->head;
```
mikeb01, line 87: These operations will need to be atomic otherwise they might not become visible in the right order.  See comment on the header for an alternative approach.

aeron-client/src/main/c/concurrent/aeron_thread.h
```diff
@@ -49,6 +49,12 @@ typedef pthread_t aeron_thread_t;
 #define aeron_thread_get_specific pthread_getspecific
 #define aeron_thread_set_specific pthread_setspecific
 
+typedef pthread_cond_t aeron_cond_t;
```
mikeb01, line 52: We need something similar to this for Windows as the missing types are causing a compilation failure on Windows.

aeron-client/src/main/c/concurrent/aeron_blocking_linked_queue.c
vyazelenko: @mikeb01 Both reading and writing threads are using the same lock so why would an ordered released be required here?

aeron-client/src/main/c/concurrent/aeron_blocking_linked_queue.c
vyazelenko: Yes, I agree it should be `while (0 == queue->size)` instead.

aeron-client/src/main/c/concurrent/aeron_blocking_linked_queue.c
vyazelenko: Given a shared lock between reader and write why would an ordered release be required?

aeron-client/src/main/c/concurrent/aeron_thread.h
vyazelenko: @mikeb01 C11 might a possible option (https://devblogs.microsoft.com/cppblog/c11-threads-in-visual-studio-2022-version-17-8-preview-2/) or https://learn.microsoft.com/en-us/windows/win32/Sync/condition-variables.

aeron-client/src/main/c/concurrent/aeron_blocking_linked_queue.c
mikeb01: I tend to be conservative with the C memory model (although I could be wrong), but I'm not 100% that C will have the same reordering restrictions for plain reads/writes that Java does.  Upon reflection it is probably fine.

aeron-client/src/main/c/concurrent/aeron_thread.h
mikeb01: Windows `CRITICAL_SECTION` and `CONDITION_VARIABLE` will workhere.

aeron-client/src/main/c/concurrent/aeron_executor.c
```diff
@@ -0,0 +1,224 @@
...
+        {
+            AERON_APPEND_ERR("%s", "");
+            return -1;
+        }
+
+        pthread_attr_t attr;
+        int result;
+        if ((result = aeron_thread_attr_init(&attr)) != 0)
+        {
+            AERON_SET_ERR(result, "%s", "aeron_thread_attr_init: %s", strerror(result));
```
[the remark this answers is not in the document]
mikeb01: Something like:
```
AERON_SET_ERR(result, "%s", "failed to create executor thread");
```
Will be sufficient.  The error handling library will call `strerror` on `result` when building the error message internally.

vyazelenko:
@nbradac You need to pull from master again.

aeron-client/src/main/c/concurrent/aeron_executor.c
```diff
@@ -0,0 +1,310 @@
...
+            }
+        }
+
+        if (aeron_blocking_linked_queue_init(&executor->queue) < 0)
+        {
+            AERON_APPEND_ERR("%s", "");
+            return -1;
+        }
+
+        pthread_attr_t attr;
```
mikeb01, line 148: Should be `aeron_thread_attr`

aeron-client/src/main/c/concurrent/aeron_executor.c
```diff
@@ -0,0 +1,310 @@
...
+
+struct aeron_executor_task_stct
+{
+    aeron_executor_t *executor;
+    aeron_executor_task_on_execute_func_t on_execute;
+    aeron_executor_task_on_complete_func_t on_complete;
+    void *clientd;
+    int result;
+    aeron_linked_queue_node_t *queue_node;
+    bool shutdown;
```
mikeb01, line 30: I think this should be marked as volatile.

aeron-driver/src/main/c/aeron_driver_conductor.c
```diff
@@ -2789,13 +2817,198 @@ void aeron_driver_conductor_on_unavailable_image(
...
+{
+    aeron_driver_async_client_command_t *async_client_command = task_clientd;
+    aeron_driver_conductor_t *conductor = executor_clientd;
+    int64_t correlation_id = async_client_command->correlated->correlation_id;
+
+    conductor->async_client_command_in_flight = false;
+
+    if (result < 0)
+    {
+        if (async_client_command->on_error == NULL)
```
mikeb01, line 2886: Flip order of operands in NULL check, e.g. `NULL == ....`

aeron-driver/src/main/c/aeron_driver_conductor.c
```diff
@@ -2789,13 +2817,198 @@ void aeron_driver_conductor_on_unavailable_image(
...
+        {
+            aeron_driver_conductor_on_error(conductor, errcode, errmsg, correlation_id);
+        }
+        else
+        {
+            // TODO set current errcode/errmsg to what was handed to this function???
```
mikeb01, line 2892: I think we do have to set those if we are going to use them below as they are thread local variables.  Can we put together a system test that will trigger this code path to see what would happen?

aeron-client/src/main/c/concurrent/aeron_executor.c
nbradac: I don't see an aeron_thread_attr defined anywhere.  Should I add a typedef in aeron_thread.h?  I think I was just copying what I saw in aeron_agent.c in aeron_agent_start().

aeron-driver/src/main/c/aeron_driver_conductor.c
nbradac: Will do.  Is this just a code style thing?  Or does this change result in different/better compiled code?

aeron-driver/src/main/c/aeron_driver_conductor.c
```diff
@@ -4183,83 +4579,114 @@ int aeron_driver_conductor_on_add_send_destination(
...
     }
 
-    AERON_SET_ERR(
-        -AERON_ERROR_CODE_UNKNOWN_PUBLICATION,
-        "unknown add destination client_id=%" PRId64 " registration_id=%" PRId64,
-        command->correlated.client_id,
-        command->registration_id);
+    async_client_command->async_command.async_resolve.uri_param_name = AERON_UDP_CHANNEL_ENDPOINT_KEY;
+    async_client_command->async_command.async_resolve.is_re_resolution = false;
+    memcpy(async_client_command->async_command.async_resolve.endpoint_name, uri->params.udp.endpoint, strlen(uri->params.udp.endpoint));
```
mikeb01, line 4648: I think we should use `strnlen(..., AERON_MAX_HOST_LEN)` and null the final byte just for belt and braces.

aeron-client/src/main/c/collections/aeron_linked_queue.c
```diff
@@ -0,0 +1,136 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <errno.h>
+#include "aeron_alloc.h"
+#include "collections/aeron_linked_queue.h"
+#include "util/aeron_error.h"
+
+struct aeron_linked_queue_node_stct {
```
mikeb01, line 22: Bracket on new line.

aeron-client/src/test/c/concurrent/aeron_executor_test.cpp
```diff
@@ -0,0 +1,259 @@
...
+
+#include <gtest/gtest.h>
+#include <queue>
+
+extern "C"
+{
+#include "concurrent/aeron_executor.h"
+}
+
+typedef struct {
```
mikeb01, line 25: Bracket on new line.

aeron-client/src/test/c/concurrent/aeron_executor_test.cpp
```diff
@@ -0,0 +1,259 @@
...
+    int some_value;
+    int some_other_value;
+} task_clientd_t;
```
mikeb01, line 28: Alias on new line.

aeron-driver/src/main/c/aeron_driver_conductor.c
nbradac: Should I define the field in the async_resolve struct to be AERON_MAX_HOST_LENGTH + 1?  Otherwise I can really only memcpy up to AERON_MAX_HOST_LEN - 1 so I can add the terminating null.

aeron-driver/src/main/c/aeron_driver_conductor.c
nbradac: Right now, the only time the `async_client_command->on_error` is used we don't check to see what the error was.  The original code just set `ss_family = AF_UNSPEC` and proceeded as normal.  The new code sets the on_error callback so that we have an opportunity to set AF_UNSPEC, but it then just calls the _complete() callback directly.  So I'm not sure we've got a good way to test this out at the moment - i.e. even if I set the errcode and errmsg thread local variables here, nobody ever reads them, so it won't matter.

PR #1567, opened 2024-02-26
[C] Implement header assembly for fragmented messages.

vyazelenko:
Align with the Java implementation.


aeron-client/src/main/c/aeron_fragment_assembler.h
```diff
@@ -87,4 +96,21 @@ inline int aeron_buffer_builder_append(
...
+    memcpy(buffer_builder->header.frame, header->frame, sizeof(aeron_data_header_t));
+}
+
+inline aeron_header_t* aeron_buffer_builder_complete_header(aeron_buffer_builder_t *buffer_builder, aeron_header_t *header)
+{
+    aeron_frame_header_t *frame_header = &buffer_builder->header.frame->frame_header;
+    int32_t max_payload_length = frame_header->frame_length - (int32_t)AERON_DATA_HEADER_LENGTH;
+    frame_header->frame_length =
+        (int32_t)aeron_logbuffer_compute_fragmented_length(buffer_builder->limit, max_payload_length);
+    frame_header->flags = (uint8_t)(header->frame->frame_header.flags | AERON_DATA_HEADER_BEGIN_FLAG);
```
mikeb01, line 112: Should this also be `|` with the flags from the first header as well?  We could end up losing flags from the first header, e.g. EOS flag and any potential flags that were added in the future.  If so we'd need to look at the Java implementation as well.

aeron-client/src/test/c/aeron_image_fragment_assembler_test.cpp
```diff
@@ -0,0 +1,348 @@
...
+    handle_fragment(handler, fragmentLength);
+    EXPECT_FALSE(isCalled);
+
+    termOffset += MTU_LENGTH;
+    m_header.initial_term_id = -1;
+    m_header.position_bits_to_shift = -7;
+    fillFrame(
+        AERON_FRAME_HEADER_VERSION,
+        AERON_DATA_HEADER_END_FLAG,
+        AERON_HDR_TYPE_EXT,
```
mikeb01, line 211: This will require changing the definition of `AERON_HDR_TYPE_EXT` to be `(-1)` or `((int16_t)0xFFFF)`, so that the type cast it explicit and there is no implicit overflow.  Fails with warnings as errors on GCC.

mikeb01:
Apart from the two comments, this all looks okay to me.

aeron-client/src/main/c/aeron_fragment_assembler.h
vyazelenko: Good point, I'll change both C and Java implementation to OR first and last frame flags.

aeron-client/src/test/c/aeron_image_fragment_assembler_test.cpp
vyazelenko: Done.

PR #1571, opened 2024-03-01
[Java] Use `archiveId` for looking up the RecordingPos counters.

vyazelenko:
* Add a command to expose `archiveId` to the Archive control protocol.
* Add `AeronArchive.archiveId()`.
* Use `archiveId` when searching for `RecordingPos` by `recordingId` and `sessionId`.

### DIFF
--- a/aeron-archive/src/main/java/io/aeron/archive/client/AeronArchive.java
+++ b/aeron-archive/src/main/java/io/aeron/archive/client/AeronArchive.java
@@ -31,7 +31,11 @@ import org.agrona.CloseHelper;
 import org.agrona.ErrorHandler;
 import org.agrona.LangUtil;
 import org.agrona.SemanticVersion;
-import org.agrona.concurrent.*;
+import org.agrona.concurrent.AgentInvoker;
+import org.agrona.concurrent.BackoffIdleStrategy;
+import org.agrona.concurrent.IdleStrategy;
+import org.agrona.concurrent.NanoClock;
+import org.agrona.concurrent.NoOpLock;
 
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
@@ -82,6 +86,7 @@ public final class AeronArchive implements AutoCloseable
     private boolean isClosed = false;
     private boolean isInCallback = false;
     private long lastCorrelationId = Aeron.NULL_VALUE;
+    private long archiveId = Aeron.NULL_VALUE;
     private final long controlSessionId;
     private final long messageTimeoutNs;
     private final Context context;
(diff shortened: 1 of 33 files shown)


aeron-archive/src/main/resources/archive/aeron-archive-codecs.xml
```diff
@@ -572,6 +572,14 @@
         <field name="recordingId"          id="3" type="int64"/>
     </sbe:message>
 
+    <sbe:message name="ArchiveIdRequest"
```
mikeb01, line 575: I was wondering if instead of (or as well as) having a separate request to query the archiveId, could we update the archive response message to always carry the archive id with it.  Then it would come back to the client during the initial connect stage and could be cached on the archive client.  It would also mean that it is always available in response to requests done via the proxy, e.g. startRecording so that it can be used immediately and not need an additional round trip.

aeron-archive/src/main/java/io/aeron/archive/client/AeronArchive.java
```diff
@@ -1576,6 +1581,38 @@ public long getRecordedLength(final long recordingId)
         }
     }
 
+    /**
+     * Get the id of the Archive.
+     *
+     * @return the id of the Archive.
+     */
+    public long archiveId()
```
mikeb01, line 1589: Should this be cached in `AeronArchive`?

vyazelenko: It is cached. See below.

aeron-archive/src/main/resources/archive/aeron-archive-codecs.xml
vyazelenko: That's an interesting point. This will have to be done in  backwards compatible way. Perhaps adding `ControlResponse2` could be an option.

mikeb01: Would we need a `ControlResponse2` to support older clients connecting to newer Archives?  Is that something we expect to support?

vyazelenko: @mikeb01 I think supporting old clients is important. I tried adding an optional field to the `ControlResponse` but that will not work for old clients connecting to the Archive.
```xml
<field name="archiveId" id="7" type="int64" presence="optional" sinceVersion="9" description="Id of the Archive sending the response."/>
```
Given the complexities of extending the response I suggest we keep a separate command instead.

aeron-archive/src/main/java/io/aeron/archive/client/AeronArchive.java
mikeb01: Missed that.  A style note, I personally like the early return approach for this sort of thing, e.g:
```java
    if (Aeron.NULL_VALUE != archiveId)
    {
        return achiveId;
    }
        
    ensureOpen();
    ensureNotReentrant();

    lastCorrelationId = aeron.nextCorrelationId();

    if (!archiveProxy.archiveId(lastCorrelationId, controlSessionId))
    {
        throw new ArchiveException("failed to send get recorded length request");
    }

    archiveId = pollForResponse(lastCorrelationId);
    return archiveId;
```
Not required change for merging though.

mikeb01:
Looks good to me.

PR #1579, opened 2024-04-02
Have cluster backup start from a specific position

mikeb01:
Where it will search for the most recent snapshot position less than or equal to the specified position. Only affects the first time the backup starts replicating from the cluster.

### DIFF
--- a/aeron-cluster/src/main/java/io/aeron/cluster/ClusterBackup.java
+++ b/aeron-cluster/src/main/java/io/aeron/cluster/ClusterBackup.java
@@ -494,6 +494,40 @@ public final class ClusterBackup implements AutoCloseable
         {
             return System.getProperty(CLUSTER_BACKUP_SOURCE_TYPE_PROP_NAME, CLUSTER_BACKUP_SOURCE_TYPE_DEFAULT);
         }
+
+        /**
+         * Sets the initial cluster replay start position so that it will start from the most recent snapshot
+         */
+        public static final long CLUSTER_REPLAY_LATEST_LOG_POSITION = Long.MAX_VALUE;
+
+        /**
+         * Sets the initial cluster replay start position to the earliest possible position, either the beginning of
+         * the log or the earliest snapshot.
+         */
+        public static final long CLUSTER_REPLAY_EARLIEST_LOG_POSITION = 0;
+
+        /**
+         * Default value for the initial cluster replay start position.
+         */
+        public static final long CLUSTER_REPLAY_INITIAL_LOG_POSITION_DEFAULT = CLUSTER_REPLAY_EARLIEST_LOG_POSITION;
+
+        /**
+         * Property name for setting the cluster replay log position.
+         */
+        public static final String CLUSTER_REPLAY_INITIAL_LOG_POSITION_PROP_NAME =
+            "cluster.backup.replay.initial.log.position";
+
+        /**
+         * Get the initial value for the cluster relay initial log position
+         *
+         * @return initial position to determine where to start replaying the log from.
+         * @see ClusterBackup.Context#clusterReplayInitialLogPosition(long)
+         */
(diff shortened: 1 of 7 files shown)


aeron-cluster/src/main/java/io/aeron/cluster/ClusterBackup.java
```diff
@@ -494,6 +494,40 @@ public static String clusterBackupSourceType()
...
+        /**
+         * Sets the initial cluster replay start position so that it will start from the most recent snapshot
+         */
+        public static final long CLUSTER_REPLAY_LATEST_LOG_POSITION = Long.MAX_VALUE;
+
+        /**
+         * Sets the initial cluster replay start position to the earliest possible position, either the beginning of
+         * the log or the earliest snapshot.
+         */
+        public static final long CLUSTER_REPLAY_EARLIEST_LOG_POSITION = 0;
```
vyazelenko, line 507: As per `CLUSTER_REPLAY_EARLIEST_LOG_POSITION` documentation using this constant should result in backup starting at the beginning of the log or the _earliest snapshot_. I think always starting from the beginning of the log is correct, because this retains the existing backup behavior.
Also we do not have multiple entries to choose the earliest snapshot to begin with and if we want to add this feature there should be a separate constant for it.

vyazelenko: Also should the value be `-1` instead of `0`?

vyazelenko (changes requested):

aeron-cluster/src/main/java/io/aeron/cluster/ClusterBackup.java
ZachBray: > Also we do not have multiple entries to choose the earliest snapshot to begin with and if we want to add this feature there should be a separate constant for it.

If I grokked the code properly, I think there is code in Cluster_Standby_ that assumes that the `BackupQueryResponse` contains more than just the most recent (local, non-standby) snapshot entries too. IIUC that is _not_ the case, i.e., `BackupQueryResponse` only gives information about the most recent snapshot entries.

(by "most recent snapshot entries" I mean snapshot entries all for the same log position but for different services) 

https://github.com/real-logic/premium-extensions/blob/9affd8d23781ba6291541190bc79b21d6878a776/aeron-cluster-standby/src/main/java/io/aeron/cluster/ClusterStandbySnapshotRetrieve.java#L134

Tangentally related question, if `BackupQueryResponse` is changed to include more than just the latest local snapshot, should `BackupQueryResponse` return standby snapshot references too? This might make it easier to start up new standby nodes in environments that only use standby snapshotting, but would it be messy? For example, a standby node restarted without its old recordings could see an entry of its own that is no longer available.

aeron-cluster/src/main/java/io/aeron/cluster/ClusterBackup.java
```diff
@@ -1785,6 +1820,38 @@ public long replicationProgressIntervalNs()
             return replicationProgressIntervalNs;
         }
 
+        /**
+         * Set the log position to start from for the first time that the backup connects to the cluster. This will
+         * only be used if there is no existing local log recording (that will resume from the recording stop position).
+         * The standby will search for the most recent snapshot less than the supplied value. Therefore, a value of 0
```
vyazelenko, line 1826: The documentation mentions `standby` but this is back config.

aeron-cluster/src/main/java/io/aeron/cluster/ClusterBackup.java
mikeb01: I think I'm going to change this from a position value to an enum with the various options (beginning of log, latest snapshot).  I don't think there is value in changing BackupQueryResponse.

PR #1584, opened 2024-04-11
[Java ]  Delegate decision on unknown schema to consensus agent

langera:


aeron-cluster/src/main/java/io/aeron/cluster/IngressAdapter.java
```diff
@@ -73,12 +72,12 @@ public Action onFragment(final DirectBuffer buffer, final int offset, final int
         messageHeaderDecoder.wrap(buffer, offset);
 
         final int schemaId = messageHeaderDecoder.schemaId();
+        final int templateId = messageHeaderDecoder.templateId();
         if (schemaId != MessageHeaderDecoder.SCHEMA_ID)
         {
-            throw new ClusterException("expected schemaId=" + MessageHeaderDecoder.SCHEMA_ID + ", actual=" + schemaId);
+            consensusModuleAgent.onUnknownMessage(schemaId, templateId, buffer, offset, length, header);
```
mjpt777, line 78: Should return after this so the same template id from another schema does not get dispatched.

aeron-cluster/src/test/java/io/aeron/cluster/IngressAdapterTest.java
```diff
@@ -0,0 +1,172 @@
...
+    {
+        buffer = new ExpandableArrayBuffer();
+        consensusModuleAgent = mock(ConsensusModuleAgent.class);
+        adapter = new IngressAdapter(0, consensusModuleAgent);
+    }
+
+    @SuppressWarnings("unused") // name used for test display name
+    @ParameterizedTest(name = "{index} {0}")
+    @MethodSource
+    public void shouldDelegateToConsensusModuleAgent(
```
mjpt777, line 64: Methods do not need to be public in a package private class.

langera: done

langera:
Consensus module can be configured now with `SchemaAdapter` instances to handle messages from schemas unknown to core Aeron.

aeron-cluster/src/test/java/io/aeron/cluster/ConsensusModuleAgentTest.java
```diff
@@ -171,7 +165,7 @@ public void shouldLimitActiveSessions()
         Tests.setField(agent, "appendPosition", mock(ReadableCounter.class));
         agent.onSessionConnect(correlationIdOne, 2, PROTOCOL_SEMANTIC_VERSION, RESPONSE_CHANNEL_ONE, new byte[0]);
 
-        clock.update(17, TimeUnit.MILLISECONDS);
+        clock.update(SCHEMA_ID, TimeUnit.MILLISECONDS);
```
mikeb01, line 168: Is there are reason why we are setting the clock time to the schema id?

langera: accidental overuse of the number 17 :)

aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModule.java
```diff
@@ -3971,6 +3971,27 @@ public TimerServiceSupplier timerServiceSupplier()
...
+        /**
+         * Registers a ConsensusModuleExtension to extend beahviour of
+         * consensus module instead of using ClusteredServices
+         *
+         * @param extensionSupplier supplier for schema adapter
+         * @return this for a fluent API.
+         */
+        public Context consensusModuleExtension(final Supplier<ConsensusModuleExtension> extensionSupplier)
+        {
+            consensusModuleExtension = extensionSupplier.get();
```
mjpt777, line 3983: New instance should be setup in context conclude. Supplier can be set as a property or loaded dynamically.

aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModuleAgent.java
```diff
@@ -3598,4 +3611,31 @@ public String toString()
...
+
+    private static final class UnknownConsensusModuleExtension implements ConsensusModuleExtension
+    {
+
+        private static final UnknownConsensusModuleExtension INSTANCE = new UnknownConsensusModuleExtension();
+        private UnknownConsensusModuleExtension()
+        {
+        }
+
+        @Override
```
mjpt777, line 3623: We don't use @Override in our code style.

aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModuleAgent.java
```diff
@@ -406,6 +408,17 @@ public void onLoadBeginSnapshot(
         }
     }
 
+    public ControlledFragmentHandler.Action onUnknownMessageSchema(
```
mjpt777, line 411: This should be renamed to `onExtensionMessage`.

aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModuleAgent.java
```diff
@@ -406,6 +408,17 @@ public void onLoadBeginSnapshot(
...
+        final int schemaId,
+        final int templateId,
+        final DirectBuffer buffer,
+        final int offset,
+        final int length,
+        final Header header)
+    {
+        return consensusModuleExtension.onFragment(schemaId, templateId, buffer, offset, length, header);
```
mjpt777, line 419: Fragments would be assembled at this stage so it would be `onMessage`.

aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModuleAgent.java
```diff
@@ -3598,4 +3611,31 @@ public String toString()
...
+        @Override
+        public ControlledFragmentHandler.Action onFragment(
+            final int schemaId,
+            final int templateId,
+            final DirectBuffer buffer,
+            final int offset,
+            final int length,
+            final Header header)
+        {
+            throw new ClusterException("expected schemaId=" + MessageHeaderDecoder.SCHEMA_ID + ", actual=" + schemaId);
```
mjpt777, line 3638: Perform null check in Agent rather than have another implementation that could defeat monomorphic optimisation.

langera: removed unknown schema impl

aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModuleAgent.java
langera: renamed

aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModuleAgent.java
langera: renamed

aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModuleAgent.java
langera: removed

aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModuleExtension.java
```diff
@@ -0,0 +1,62 @@
...
+     */
+    ControlledFragmentHandler.Action onMessage(
+        int schemaId,
+        int templateId,
+        DirectBuffer buffer,
+        int offset,
+        int length,
+        Header header);
+
+    @Override
```
mjpt777, line 58: Needs removed.

langera: supplier called in conclude

PR #1591, opened 2024-04-22
Correctly close a subscription and publication

sergiu128:
Previous to this commit, a `close()` on a publication would not release the publication through the underlying conductor. Analog for subscriptions.


mikeb01:
This change results in a large number segfaults in the test code.

PR #1597, opened 2024-05-08
Expose a "retransmitted bytes" counter

ZachBray:
Exposes a counter for the number of bytes retransmitted due to resending after receiving NAKs.

Caveats (shown in 44e89dfd7539ea2fa1f6a82655c119ef2681ab97):
- Only counts bytes sent to multiple MDC destinations once
- Retransmitted bytes are not included in total bytes

I looked into addressing those caveats, but I couldn't decide on the best approach.

Would it be best to revert the disabled tests? Whilst it is useful documentation, it will probably rot.


mjpt777:
LGTM

PR #1602, opened 2024-05-20
[C] add and use aeron_mkdir_recursive

nbradac:
aeron_mkdir_recursive() will recursively create parent directories until it finds one that's already created.

Also, it calls itself... recursively.

(Windows tests are failing... why is Windows different!?!?!)

### DIFF
--- a/aeron-client/src/main/c/util/aeron_fileutil.c
+++ b/aeron-client/src/main/c/util/aeron_fileutil.c
@@ -453,6 +453,51 @@ int aeron_open_file_rw(const char *path)
 }
 #endif
 
+int aeron_mkdir_recursive(const char *pathname, mode_t mode)
+{
+    if (aeron_mkdir(pathname, mode) == 0)
+    {
+        return 0;
+    }
+
+    if (errno != ENOENT)
+    {
+        return -1;
+    }
+
+    char *_pathname = strdup(pathname);
+    char *p;
+    int rc = -1;
+
+    for (p = _pathname + strlen(_pathname) - 1; p != _pathname; p--)
+    {
+        if (*p == AERON_FILE_SEP)
+        {
+            *p = '\0';
+            // _pathname is now the parent directory of the original pathname
+            rc = aeron_mkdir_recursive(_pathname, mode);
+            break;
+        }
+    }
+
+    if (p == _pathname)
+    {
+        // after iterating through _pathname, we didn't find a separator
+        rc = -1;
+    }
+
+    free(_pathname);
+
+    if (rc == 0)
+    {
+        // if rc is 0, then we were able to create the parent directory
+        // so retry the original pathname
+        return aeron_mkdir(pathname, mode);
+    }
+
+    return rc;
+}
+
 #include <inttypes.h>
 
 #define AERON_BLOCK_SIZE (4 * 1024)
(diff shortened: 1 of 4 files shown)


aeron-client/src/main/c/util/aeron_fileutil.c
```diff
@@ -453,6 +453,51 @@ int aeron_open_file_rw(const char *path)
...
+int aeron_mkdir_recursive(const char *pathname, int permission)
+{
+    if (aeron_mkdir(pathname, permission) == 0)
+    {
+        return 0;
+    }
+
+    if (errno != ENOENT)
+    {
+        return -1;
```
mikeb01, line 465: Need to use AERON_SET_ERR here.

aeron-client/src/main/c/util/aeron_fileutil.c
```diff
@@ -453,6 +453,51 @@ int aeron_open_file_rw(const char *path)
...
+            *p = '\0';
+            // _pathname is now the parent directory of the original pathname
+            rc = aeron_mkdir_recursive(_pathname, permission);
+            break;
+        }
+    }
+
+    if (p == _pathname)
+    {
+        // after iterating through _pathname, we didn't find a separator
```
mikeb01, line 485: Need AERON_SET_ERR

aeron-client/src/main/c/util/aeron_fileutil.c
```diff
@@ -453,6 +453,51 @@ int aeron_open_file_rw(const char *path)
...
+    char *p;
+    int rc = -1;
+
+    for (p = _pathname + strlen(_pathname) - 1; p != _pathname; p--)
+    {
+        if (*p == AERON_FILE_SEP)
+        {
+            *p = '\0';
+            // _pathname is now the parent directory of the original pathname
+            rc = aeron_mkdir_recursive(_pathname, permission);
```
mikeb01, line 478: We need to be a little bit smart about the error here.  If this returns -1, then we should be using AERON_APPEND_ERR to build the error stack.

aeron-client/src/main/c/util/aeron_fileutil.c
```diff
@@ -462,40 +462,43 @@ int aeron_mkdir_recursive(const char *pathname, int permission)
...
+
+            if (0 == rc)
+            {
+                // if rc is 0, then we were able to create the parent directory
+                // so retry the original pathname
+                return aeron_mkdir(pathname, permission);
+            }
+            else
+            {
+                AERON_APPEND_ERR("%s", "");
```
mikeb01, line 491: We could also change this to be: `AERON_APPEND_ERR("pathname=%s", pathname);`, which would show each of the component parts of the path as it was being processed.

aeron-client/src/test/c/util/aeron_fileutil_test.cpp
```diff
@@ -25,6 +25,18 @@ extern "C"
 #include "util/aeron_error.h"
 }
 
+#if defined(AERON_COMPILER_GCC)
+#define removeDir remove
+#elif defined(AERON_COMPILER_MSVC)
+#define removeDir RemoveDirectoryA
```
mikeb01, line 31: Could `aeron_delete_directory` work here?

PR #1603, opened 2024-05-22
Add unicast semantics to RetransmitHandler

nbradac:
The RetransmitHandler now supports unicast semantics where it only keeps track of (at most) one outstanding NAK, and it immediately replaces old NAKs with new NAKs, on the assumption that a new NAK means the receiver has moved past the previous gap (i.e. the previous NAK 'worked' and the RX was received).

To test this, a new 'MultiGapLossGenerator' was created that will introduce loss across a series of regularly spaced intervals to force the condition where more than the max outstanding retransmit actions are sent in a short period of time.  The previous (i.e. 'multicast') retransmit handler logic would just drop subsequent NAKs, requiring some NAKs be resent.


aeron-driver/src/main/java/io/aeron/driver/DriverConductor.java
```diff
@@ -1598,7 +1598,8 @@ private NetworkPublication newNetworkPublication(
                 ctx.senderCachedNanoClock(),
                 ctx.systemCounters().get(INVALID_PACKETS),
                 ctx.retransmitUnicastDelayGenerator(),
-                ctx.retransmitUnicastLingerGenerator());
+                ctx.retransmitUnicastLingerGenerator(),
+                udpChannel.isMulticast());
```
mikeb01, line 1602: We use to have a variable or function called `hasGroupSemantics` which I think got dropped during the previous iteration through the RetransmitHandler.  I think we need to reintroduce that as we should only be using the unicast semantics if it actually unicast.  Specifically we need to treat MDC like it is multicast for retransmits and we should also take into account the `group=true` URI parameter.

aeron-driver/src/main/c/aeron_retransmit_handler.c
```diff
@@ -31,13 +31,22 @@ int aeron_retransmit_handler_init(
...
     int64_t *invalid_packets_counter,
     uint64_t delay_timeout_ns,
-    uint64_t linger_timeout_ns)
+    uint64_t linger_timeout_ns,
+    bool is_multicast)
 {
     handler->invalid_packets_counter = invalid_packets_counter;
     handler->delay_timeout_ns = delay_timeout_ns;
     handler->linger_timeout_ns = linger_timeout_ns;
+    handler->is_multicast = is_multicast;
```
mikeb01, line 40: Need to take into account MDC & `group=true` (see comment on Java implementation).

mikeb01:
LGTM

PR #1605, opened 2024-05-24
[Java] Version interface to allow testing usage of it

langera:
This should allow to verify logic of validating versions in a test. 
constants are still there for the cases where access to version info is on the hot path (+ regardless JIT might optimize it anyway?)


mikeb01:
LGTM.

PR #1607, opened 2024-06-03
[C] Declare a volatile pointer to `aeron_image_list_stct` instead of a pointer to a volatile struct.

vyazelenko:


aeron-client/src/main/c/aeron_subscription.c
```diff
@@ -119,7 +119,7 @@ int aeron_subscription_close(
     return 0;
 }
 
-int aeron_subscription_alloc_image_list(volatile aeron_image_list_t **image_list, size_t length)
+int aeron_subscription_alloc_image_list(aeron_image_list_t * volatile * image_list, size_t length)
```
mikeb01, line 122: With our coding style should it be:
```
int aeron_subscription_alloc_image_list(aeron_image_list_t *volatile *image_list, size_t length)
```

mikeb01:
With in a number of methods in the `aeron_subscription.c` file (e.g. `aeron_subscription_controlled_poll`) we are declaring the variable:
```
volatile aeron_image_list_t *image_list;
```
Should that be
```
aeron_image_list_t *volatile image_list;
```

vyazelenko:
@mikeb01 I've addressed all of the issues and replace all volatile struct occurrences.

mikeb01:
I'll have a look again today.

PR #1608, opened 2024-06-04
[java] optimize the logic of LogAdapter to determine whether the stream is closed.

WorkingChen:
The purpose of this optimization is to quickly detect when the stream is closed, and then enter the election phase. In the scenario of leader.gracefulClose(), it shortens the time for cluster election.


WorkingChen:
> Unfortunately closing based on EOS in the consensus module could result in data not being completely written to the archive. By waiting until the Image has closed, means that the other subscription reading that image (i.e. the Archive) has also received the EOS and closed the recording. We feel that it is cleaner to allow the archive to complete the writing of the data it has received instead of closing early and relying on replication/catch up during the election process to back fill that data.

Can we determine the status of the counter (appendPosition)? Because after the archive completes its work, it will also release the Recording-pos counter, which directly modifies the memory state. We don't need to wait for onTimer (invoke image.unavailableHandler()) or slowTickWork (invoke pollArchiveEvents()) to know that the data has been written and completed, and we should re-enter the election state.

PR #1615, opened 2024-06-27
Intercept replay and recording session state changes

nbradac:


aeron-archive/src/main/java/io/aeron/archive/ReplicationSession.java
```diff
@@ -539,7 +554,7 @@ private int replayToken()
         if (NULL_VALUE != replayToken || !ChannelUri.parse(replicationChannel).hasControlModeResponse())
         {
             workCount++;
-            state(State.GET_ARCHIVE_PROXY);
+            state(State.GET_ARCHIVE_PROXY, "replay token"); // TODO
```
mikeb01, line 557: Outstanding TODO.

aeron-archive/src/main/java/io/aeron/archive/ReplicationSession.java
```diff
@@ -563,7 +578,7 @@ private int replayToken()
         if (hasResponse(poller))
         {
             replayToken = poller.relevantId();
-            state(State.GET_ARCHIVE_PROXY);
+            state(State.GET_ARCHIVE_PROXY, "hasResponse"); // TODO
```
mikeb01, line 581: Outstanding TODO.

aeron-agent/src/main/java/io/aeron/agent/ArchiveEventEncoder.java
```diff
@@ -29,6 +29,198 @@ private ArchiveEventEncoder()
...
+        int encodedLength = encodeLogHeader(encodingBuffer, offset, captureLength, length);
+
+        encodingBuffer.putLong(offset + encodedLength, id, LITTLE_ENDIAN);
+        encodedLength += SIZE_OF_LONG;
+        encodingBuffer.putLong(offset + encodedLength, recordingId, LITTLE_ENDIAN);
+        encodedLength += SIZE_OF_LONG;
+        encodingBuffer.putLong(offset + encodedLength, position, LITTLE_ENDIAN);
+        encodedLength += SIZE_OF_LONG;
+
+        encodedLength += encodingBuffer.putStringAscii(offset + encodedLength, reason);
```
mikeb01, line 53: Because we have two variable length strings to be encoded, we should consider the total length of the strings to determine if the data needs to be truncated.  It would probably be safer to put the state change before the reason text as the state description is likely to be shorter.

aeron-agent/src/main/java/io/aeron/agent/ArchiveEventEncoder.java
```diff
@@ -29,6 +29,198 @@ private ArchiveEventEncoder()
...
+        encodingBuffer.putLong(offset + encodedLength, correlationId, LITTLE_ENDIAN);
+        encodedLength += SIZE_OF_LONG;
+        encodingBuffer.putLong(offset + encodedLength, streamId, LITTLE_ENDIAN);
+        encodedLength += SIZE_OF_LONG;
+        encodingBuffer.putLong(offset + encodedLength, recordingId, LITTLE_ENDIAN);
+        encodedLength += SIZE_OF_LONG;
+        encodingBuffer.putLong(offset + encodedLength, startPosition, LITTLE_ENDIAN);
+        encodedLength += SIZE_OF_LONG;
+
+        encodedLength += encodingBuffer.putStringAscii(offset + encodedLength, publicationChannel);
```
mikeb01, line 91: This should be using something like `CommonEncoder.encodeTrailingString` to truncate the message if required.

aeron-archive/src/main/java/io/aeron/archive/ReplaySession.java
```diff
@@ -437,7 +440,7 @@ private boolean hasPublicationAdvanced(final long position, final int alignedLen
 
             if (replayPosition >= replayLimit)
             {
-                state(State.INACTIVE);
+                state(State.INACTIVE, "position past limit");
```
vyazelenko, line 443: Add `replayPosition` and `replayLimit` to the message.

aeron-archive/src/main/java/io/aeron/archive/ReplaySession.java
```diff
@@ -517,7 +520,7 @@ else if (replayLimit >= oldStopPosition)
 
         if (replayPosition >= replayLimit)
         {
-            state(State.INACTIVE);
+            state(State.INACTIVE, "position past limit (notExtended)");
```
vyazelenko, line 523: Add `replayPosition` and `replayLimit` to the message.

aeron-archive/src/main/java/io/aeron/archive/ReplicationSession.java
```diff
@@ -258,7 +260,7 @@ public int doWork()
         }
         catch (final Exception ex)
         {
-            state(State.DONE);
+            state(State.DONE, "exception during doWork: " + ex.getMessage());
```
vyazelenko, line 263: Drop the prefix and simply use exception message in all cases.

aeron-archive/src/main/java/io/aeron/archive/ReplicationSession.java
```diff
@@ -300,8 +302,9 @@ public void onRecordingDescriptor(
 
         if (Aeron.NULL_VALUE != fileIoMaxLength && fileIoMaxLength < mtuLength)
         {
-            state(State.DONE);
-            error("Replication fileIoMaxLength is less than than the recording mtuLength", ArchiveException.GENERIC);
+            final String errorMsg = "Replication fileIoMaxLength is less than than the recording mtuLength";
```
vyazelenko, line 305: Add values for both `fileIoMaxLength` and `mtuLength` to the error.

aeron-archive/src/main/java/io/aeron/archive/ReplicationSession.java
```diff
@@ -346,27 +349,38 @@ else if (isDestinationRecordingEmpty)
...
             signal(stopPosition, SYNC);
             nextState = State.DONE;
+            if (startPosition == stopPosition)
+            {
+                reason = "start/stop positions are the same";
+            }
+            else
+            {
+                reason = "stop position equals catalog stop position";
+            }
```
vyazelenko, line 380: Since we issue a `SYNC` signal here we can simply use "sync" or "in sync" (or maybe "complete") as a reason. I do not see a value in printing how the sync was reached.

aeron-archive/src/main/java/io/aeron/archive/ReplicationSession.java
```diff
@@ -396,13 +410,13 @@ private int connect()
...
                     asyncConnect = null;
-                    state(State.REPLICATE_DESCRIPTOR);
+                    state(State.REPLICATE_DESCRIPTOR, "");
                     workCount += 1;
                 }
             }
             catch (final AeronException ex)
             {
-                state(State.DONE);
+                state(State.DONE, "exception during connect: " + ex.getMessage());
```
vyazelenko, line 419: Drop prefix.

aeron-archive/src/main/java/io/aeron/archive/ReplicationSession.java
```diff
@@ -481,7 +496,7 @@ else if (epochClock.time() >= (timeOfLastActionMs + actionTimeoutMs))
                     throw new ArchiveException("cannot live merge without active source recording");
                 }
 
-                state(State.EXTEND);
+                state(State.EXTEND, "srcRecordingPosition");
```
vyazelenko, line 499: We should either include value of the `srcRecordingPosition` into the message or maybe use empty string, i.e. as the state transition `SRC_RECORDING_POSITION -> EXTEND` is self explanatory.

aeron-archive/src/main/java/io/aeron/archive/ReplicationSession.java
```diff
@@ -516,7 +531,7 @@ private int extend()
 
         if (null == recordingSubscription)
         {
-            state(State.DONE);
+            state(State.DONE, "null recordingSubscription");
```
vyazelenko, line 534: The `recordingSubscription` will be `null` only if `conductor.extendRecording` call fails with an error. The error message is directly sent via the ControlSession back to the AeronArchive client. The response will be logged (if enabled) via `CMD_OUT_RESPONSE` event. However it would be great if we could get the error message and capture it in a `reason` field here. That would require changing `conductor.extendRecording` method (e.g. return `Object` instead of `Subscription` and return error String if error and actual subscription if success).

aeron-archive/src/main/java/io/aeron/archive/ReplicationSession.java
```diff
@@ -526,7 +541,7 @@ private int extend()
                 recordingSubscription.asyncAddDestination(replayDestination);
             }
 
-            state(State.REPLAY_TOKEN);
+            state(State.REPLAY_TOKEN, "");
```
vyazelenko, line 544: Should be "replay destination added" probably including `replayDestination` value.

aeron-archive/src/main/java/io/aeron/archive/ReplicationSession.java
vyazelenko: Use an empty string here. This is a technical state transition that is only happening when response channels feature is used during replication.

aeron-archive/src/main/java/io/aeron/archive/ReplicationSession.java
vyazelenko: Same as above.

aeron-archive/src/main/java/io/aeron/archive/ReplicationSession.java
```diff
@@ -576,7 +591,7 @@ private int getArchiveProxy()
         if (NULL_VALUE == replayToken)
         {
             ++workCount;
-            state(State.REPLAY);
+            state(State.REPLAY, "no replayToken");
```
vyazelenko, line 594: Empty string will be less confusing.

aeron-archive/src/main/java/io/aeron/archive/ReplicationSession.java
```diff
@@ -624,7 +639,7 @@ private int getArchiveProxy()
         }
 
         responseArchiveProxy = new ArchiveProxy(responsePublication);
-        state(State.REPLAY);
+        state(State.REPLAY, "getArchiveProxy");
```
vyazelenko, line 642: "ArchiveProxy created"

aeron-archive/src/main/java/io/aeron/archive/ReplicationSession.java
```diff
@@ -703,7 +718,7 @@ else if (epochClock.time() >= (timeOfLastActionMs + actionTimeoutMs))
             if (hasResponse(poller))
             {
                 srcReplaySessionId = poller.relevantId();
-                state(State.AWAIT_IMAGE);
+                state(State.AWAIT_IMAGE, "");
```
vyazelenko, line 721: `"srcReplaySessionId=" + srcReplaySessionId`

aeron-archive/src/main/java/io/aeron/archive/ReplicationSession.java
```diff
@@ -722,7 +737,7 @@ private int awaitImage()
         if (null != image)
         {
             this.image = image;
-            state(null == liveDestination ? State.REPLICATE : State.CATCHUP);
+            state(null == liveDestination ? State.REPLICATE : State.CATCHUP, "awaitImage");
```
vyazelenko, line 740: `"image.correlationId=" + image.correlationId() + ", image.sessionId=" + image.sessionId() + ", image.jointPosition=" + image.joinPosition()`

aeron-archive/src/main/java/io/aeron/archive/ReplicationSession.java
```diff
@@ -767,7 +782,7 @@ private int replicate()
             }
 
             srcReplaySessionId = NULL_VALUE;
-            state(State.DONE);
+            state(State.DONE, "replicate");
```
vyazelenko, line 785: Use `isSync ? "sync" : "done"` instead of `"replicate"`. There is already a dedicated log event `logReplicationSessionDone` which will print full information.

aeron-archive/src/main/java/io/aeron/archive/ReplicationSession.java
```diff
@@ -836,7 +851,7 @@ else if (shouldStopReplay(position))
                     replayDestination = null;
                     recordingSubscription = null;
                     signal(position, MERGE);
-                    state(State.DONE);
+                    state(State.DONE, "attemptLiveJoin");
```
vyazelenko, line 854: "merge"

aeron-archive/src/main/java/io/aeron/archive/ReplicationSession.java
```diff
@@ -780,7 +795,7 @@ private int catchup()
 
         if (image.position() >= srcRecordingPosition)
         {
-            state(State.ATTEMPT_LIVE_JOIN);
+            state(State.ATTEMPT_LIVE_JOIN, "image position >= srcRecordingPosition");
```
vyazelenko, line 798: Include values.

aeron-archive/src/main/java/io/aeron/archive/ReplicationSession.java
```diff
@@ -396,13 +410,13 @@ private int connect()
                 {
                     srcArchive = archive;
                     asyncConnect = null;
-                    state(State.REPLICATE_DESCRIPTOR);
+                    state(State.REPLICATE_DESCRIPTOR, "");
```
vyazelenko, line 413: "connected to the Archive"

aeron-archive/src/main/java/io/aeron/archive/ReplicationSession.java
```diff
@@ -346,27 +349,38 @@ else if (isDestinationRecordingEmpty)
...
-                state(State.DONE);
-                error("cannot live merge without active source recording", ArchiveException.GENERIC);
+                final String errorMsg = "cannot live merge without active source recording";
+                state(State.DONE, errorMsg);
+                error(errorMsg, ArchiveException.GENERIC);
                 return;
             }
 
             nextState = State.SRC_RECORDING_POSITION;
+            reason = "has liveDestination";
```
vyazelenko, line 365: Use `"liveDestination=" + liveDestination` instead.

aeron-archive/src/main/java/io/aeron/archive/ReplicationSession.java
```diff
@@ -346,27 +349,38 @@ else if (isDestinationRecordingEmpty)
         }
 
         State nextState = State.EXTEND;
+        String reason = "onRecordingDescriptor";
```
vyazelenko, line 352: Use empty string by default.

aeron-archive/src/main/java/io/aeron/archive/ReplicationSession.java
```diff
@@ -202,7 +204,7 @@ public int doWork()
         {
             if (null != recordingSubscription && recordingSubscription.isClosed())
             {
-                state(State.DONE);
+                state(State.DONE, "recording subscription is closed");
```
vyazelenko, line 207: "recording subscription closed"

aeron-archive/src/main/java/io/aeron/archive/ReplicationSession.java
```diff
@@ -146,6 +146,8 @@ enum State
         {
             isDestinationRecordingEmpty = false;
         }
+
+        logStarted(replicationId, controlSession.sessionId(), srcRecordingId, dstRecordingId, replicationChannel);
```
vyazelenko, line 150: I do not think we need an additional `logStarted` method. The start of the replication is logged via `CMD_IN_REPLICATE` or `CMD_IN_REPLICATE2` events.

aeron-archive/src/main/java/io/aeron/archive/ReplaySession.java
```diff
@@ -586,12 +589,36 @@ static boolean notHeaderAligned(
...
+        final long sessionId,
+        final long controlSessionId,
+        final long correlationId,
+        final long streamId,
+        final long recordingId,
+        final long startPosition,
+        final String publicationChannel)
+    {
+    }
+
```
vyazelenko, line 621: Not needed. Covered via existing `CMD_IN_REPLAY` and `CMD_IN_START_BOUNDED_REPLAY` events.

aeron-archive/src/main/java/io/aeron/archive/ReplaySession.java
```diff
@@ -206,7 +209,7 @@ public int doWork()
         if (State.INACTIVE == state)
         {
             closeRecordingSegment();
-            state(State.DONE);
+            state(State.DONE, "state is inactive");
```
vyazelenko, line 212: Use empty string.

aeron-archive/src/main/java/io/aeron/archive/RecordingSession.java
```diff
@@ -260,21 +261,41 @@ else if (image.isEndOfStream() || image.isClosed())
...
+    }
+
+    @SuppressWarnings("unused")
+    private void logStarted(
+        final long recordingId,
+        final long controlSessionId,
+        final long correlationId,
+        final String subscriptionChannel)
+    {
+    }
```
vyazelenko, line 300: Drop this method, i.e. this is covered via existing events `CMD_IN_START_RECORDING` and `CMD_IN_START_RECORDING2`.

aeron-archive/src/main/java/io/aeron/archive/RecordingSession.java
```diff
@@ -238,7 +239,7 @@ private int record()
             }
             else if (image.isEndOfStream() || image.isClosed())
             {
-                state(State.INACTIVE);
+                state(State.INACTIVE, "EOS: " + image.isEndOfStream() + ", Closed: " + image.isClosed());
```
vyazelenko, line 242: `"image.isEndOfStream=" + image.isEndOfStream() + ", image.isClosed=" + image.isClosed()`

aeron-archive/src/main/java/io/aeron/archive/RecordingSession.java
```diff
@@ -136,7 +138,7 @@ public int doWork()
 
         if (State.INACTIVE == state)
         {
-            state(State.STOPPED);
+            state(State.STOPPED, "state is stopped");
```
vyazelenko, line 141: Use empty string.

aeron-agent/src/test/java/io/aeron/agent/ArchiveEventLoggerTest.java
```diff
@@ -66,7 +66,8 @@ void after()
         names = {
             "CMD_OUT_RESPONSE", "REPLICATION_SESSION_STATE_CHANGE",
             "CONTROL_SESSION_STATE_CHANGE", "REPLAY_SESSION_ERROR", "CATALOG_RESIZE",
-            "REPLICATION_SESSION_DONE"
+            "REPLICATION_SESSION_DONE", "REPLAY_SESSION_STATE_CHANGE", "RECORDING_SESSION_STATE_CHANGE",
+            "REPLICATION_SESSION_STARTED", "REPLAY_SESSION_STARTED", "RECORDING_SESSION_STARTED"
```
vyazelenko, line 70: Not needed. Covered by `CMD_IN_REPLICATE`, `CMD_IN_REPLICATE2` and `CMD_IN_TAGGED_REPLICATE`.
Please check those events that they log enough details.

aeron-agent/src/main/java/io/aeron/agent/ArchiveEventLogger.java
```diff
@@ -45,6 +45,11 @@ public final class ArchiveEventLogger
 
     static final EnumSet<ArchiveEventCode> CONTROL_REQUEST_EVENTS = complementOf(of(
         CMD_OUT_RESPONSE,
+        REPLICATION_SESSION_STARTED,
+        REPLAY_SESSION_STARTED,
+        RECORDING_SESSION_STARTED,
+        REPLAY_SESSION_STATE_CHANGE,
+        RECORDING_SESSION_STATE_CHANGE,
```
vyazelenko, line 52: `*_SESSION_STARTED` are not needed since they are covered by existing events. See below.

vyazelenko (changes requested):

aeron-agent/src/main/java/io/aeron/agent/ArchiveEventCode.java
```diff
@@ -100,8 +100,22 @@ public enum ArchiveEventCode implements EventCode
...
-        42, ReplayTokenRequestDecoder.TEMPLATE_ID, ArchiveEventDissector::dissectControlRequest);
+        42, ReplayTokenRequestDecoder.TEMPLATE_ID, ArchiveEventDissector::dissectControlRequest),
 
+    REPLAY_SESSION_STATE_CHANGE(43, -1,
+        (event, buffer, offset, builder) -> dissectReplaySessionStateChange(buffer, offset, builder)),
+
+    RECORDING_SESSION_STATE_CHANGE(44, -1,
+        (event, buffer, offset, builder) -> dissectRecordingSessionStateChange(buffer, offset, builder)),
+
+    RECORDING_SESSION_STARTED(45, -1,
```
vyazelenko, line 111: Covered by `CMD_IN_START_RECORDING2` and `CMD_IN_START_RECORDING`.

aeron-agent/src/main/java/io/aeron/agent/ArchiveEventCode.java
```diff
@@ -100,8 +100,22 @@ public enum ArchiveEventCode implements EventCode
...
+        (event, buffer, offset, builder) -> dissectRecordingSessionStarted(buffer, offset, builder)),
+
+    REPLAY_SESSION_STARTED(46, -1,
```
vyazelenko, line 114: Covered by `CMD_IN_REPLAY` and `CMD_IN_START_BOUNDED_REPLAY`.

aeron-agent/src/main/java/io/aeron/agent/ArchiveEventCode.java
```diff
@@ -100,8 +100,22 @@ public enum ArchiveEventCode implements EventCode
...
+        (event, buffer, offset, builder) -> dissectReplaySessionStarted(buffer, offset, builder)),
+
+    REPLICATION_SESSION_STARTED(47, -1,
```
vyazelenko, line 117: Covered by `CMD_IN_REPLICATE`, `CMD_IN_REPLICATE2` and `CMD_IN_TAGGED_REPLICATE`.

aeron-agent/src/main/java/io/aeron/agent/ArchiveEventDissector.java
```diff
@@ -515,6 +515,106 @@ static void dissectReplicationSessionDone(
...
+        absoluteOffset += dissectLogHeader(CONTEXT, REPLAY_SESSION_STATE_CHANGE, buffer, absoluteOffset, builder);
+
+        final long sessionId = buffer.getLong(absoluteOffset, LITTLE_ENDIAN);
+        absoluteOffset += SIZE_OF_LONG;
+        final long recordingId = buffer.getLong(absoluteOffset, LITTLE_ENDIAN);
+        absoluteOffset += SIZE_OF_LONG;
+        final long position = buffer.getLong(absoluteOffset, LITTLE_ENDIAN);
+        absoluteOffset += SIZE_OF_LONG;
+
+        builder.append(": sessionId=").append(sessionId);
```
vyazelenko, line 531: This should be `": replaySessionId="`.
What would be also valuable is to print the `sessionId` of the replay Publication which is [encoded in the lower 32 bits](https://github.com/real-logic/aeron/blob/master/aeron-archive/src/main/java/io/aeron/archive/ArchiveConductor.java#L782) of the `replaySessionId`. 


For example:
```
replaySessionId=16231259302 sessionId=-948609882
```
or
```
replaySessionId=16231259302 replayId=3 sessionId=-948609882
```

aeron-agent/src/main/java/io/aeron/agent/ArchiveEventDissector.java
```diff
@@ -528,7 +528,9 @@ static void dissectReplaySessionStateChange(
         final long position = buffer.getLong(absoluteOffset, LITTLE_ENDIAN);
         absoluteOffset += SIZE_OF_LONG;
 
-        builder.append(": sessionId=").append(sessionId);
+        builder.append(": replaySessionId=").append(sessionId);
+        builder.append(" replayId=").append(sessionId >> 32);
+        builder.append(" sessionId=").append(sessionId & 0xFFFF_FFFFL);
```
vyazelenko, line 533: You can just cast to `int`.

nbradac: Is it 'better' to cast it to an int?  I don't have strong feelings either way, but I kind of like the explicit nature of doing a bitwise AND.

nbradac: Casting to an int yields a different value compared with '& 0xFFFF_FFFFL'.  The sign gets switched.

nbradac: Oh I get it... if I use the &, the returned value is still a long, and the last bit is 0, so the number is positive.  If I cast to an int, it's an int... in which case (in the case of 20000000000), the last bit is a 1.  So the number is negative.  The 32 bits in question are the same, but how it looks when it gets printed is quite different.

Which is 'correct'?  Seems like in most places the session id is passed around as a long, in which case casting to an int and printing that is probably the wrong approach.

vyazelenko: `Publication#sessionId` is a 32 bit value (`int` in Java) so casting to an `int` is correct way to extract this value out of `long` value.
Running this test:
```java
@ParameterizedTest
    @ValueSource(ints = { Integer.MIN_VALUE, -777, -1, 0, 1, 42, Integer.MAX_VALUE })
    void extractSessionId(final int sessionId)
    {
        final int replayId = 0xFF;
        final long replicationSessionId = ((long)replayId << 32) | sessionId;
        System.out.println("replicationSessionId=" + replicationSessionId);
        System.out.println("cast: " + (int)replicationSessionId);
        System.out.println("mask: " + (replicationSessionId & 0xFFFF_FFFFL));
    }
```
I get the following output:
```
replicationSessionId=-2147483648
cast: -2147483648
mask: 2147483648
replicationSessionId=-777
cast: -777
mask: 4294966519
replicationSessionId=-1
cast: -1
mask: 4294967295
replicationSessionId=1095216660480
cast: 0
mask: 0
replicationSessionId=1095216660481
cast: 1
mask: 1
replicationSessionId=1095216660522
cast: 42
mask: 42
replicationSessionId=1097364144127
cast: 2147483647
mask: 2147483647
```

vyazelenko (approved):

PR #1618, opened 2024-07-04
[Java] add attemptLiveJoinBackoff to ReplayMerge

nbradac:


aeron-archive/src/main/java/io/aeron/archive/client/ReplayMerge.java
```diff
@@ -524,6 +537,17 @@ else if (shouldStopAndRemoveReplay(position))
         return workCount;
     }
 
+    private long acquireNextAttemptLiveJoinBackoff()
+    {
+        if (attemptLiveJoinBackoff >= ATTEMPT_LIVE_JOIN_BACKOFF_MAX_MS)
+        {
+            return ATTEMPT_LIVE_JOIN_BACKOFF_MAX_MS;
+        }
+        attemptLiveJoinBackoff *= 2;
```
mikeb01, line 546: This could probably be inlined into the line below.

aeron-archive/src/main/java/io/aeron/archive/client/ReplayMerge.java
```diff
@@ -71,6 +72,8 @@ enum State
     private long nextTargetPosition = Aeron.NULL_VALUE;
     private long positionOfLastProgress = Aeron.NULL_VALUE;
     private long timeOfLastProgressMs;
+    private long timeOfNextAttemptLiveJoin;
+    private long attemptLiveJoinBackoff = 8;
```
mikeb01, line 76: Suffix time based variables with the units, e.g. `timeOfNextAttemptLiveJoinMs`.

mikeb01:
Couple of small style amendments, otherwise looks good.

aeron-archive/src/main/java/io/aeron/archive/client/ReplayMerge.java
```diff
@@ -148,6 +152,7 @@ public ReplayMerge(
 
         subscription.asyncAddDestination(replayDestination);
         timeOfLastProgressMs = epochClock.time();
+        timeOfNextAttemptLiveJoinMs = epochClock.time();
```
vyazelenko, line 155: Should call `epochClock.time()` only once and assign to both fields.

aeron-archive/src/main/java/io/aeron/archive/client/ReplayMerge.java
```diff
@@ -524,6 +538,16 @@ else if (shouldStopAndRemoveReplay(position))
         return workCount;
     }
 
+    private long acquireNextAttemptLiveJoinBackoffMs()
+    {
+        if (attemptLiveJoinBackoffMs >= ATTEMPT_LIVE_JOIN_BACKOFF_MAX_MS)
+        {
+            return ATTEMPT_LIVE_JOIN_BACKOFF_MAX_MS;
+        }
```
vyazelenko, line 546: This check seems to redundant as the next line uses `min` function, i.e. the value `attemptLiveJoinBackoffMs` will not increase past `ATTEMPT_LIVE_JOIN_BACKOFF_MAX_MS`.
Was it added for efficiency reasons, i.e. to avoid a multiple and field assignment?

vyazelenko (changes requested):

mjpt777:
I think it is better to throttle the attempts to call get recording position. This would also work for the earlier state in the workflow.

vyazelenko:
@mjpt777 @nbradac I've [pushed a change to ReplayMerge](https://github.com/real-logic/aeron/commit/47d57f3d05483fa915de57abf605b8f6bd479851) where I replaced `getRecordingPosition` call with a call to `getMaxRecordedPosition`. This is to avoid a fallback call to `getStopPosition`. I've also removed a second call to resolve position from the `attemptToJoinLive`.

aeron-archive/src/main/java/io/aeron/archive/client/ReplayMerge.java
```diff
@@ -524,6 +533,21 @@ else if (shouldStopAndRemoveReplay(position))
...
+            getRecordingPositionBackoffMs = Long.min(
+                getRecordingPositionBackoffMs * 2, GET_RECORDING_POSITION_BACKOFF_MAX_MS);
+            timeOfNextGetRecordingPositionMs = nowMs + getRecordingPositionBackoffMs;
+
+            return true;
+        }
+
+        return false;
+    }
+
```
vyazelenko, line 550: This method should get a `correlationId` directly if it going to call the `getMaxRecordedPosition` method instead of it being supplied at the call site. Otherwise lots of ids will be wasted during the back-off periods.

Another thing to consider if the back-off should be always applied and not only upon successful calls.

nbradac: Good idea.  I didn't think about the wasted correlationIds.

vyazelenko: It is more about not doing redundant work than actual waste on the 64-bit value.

vyazelenko (approved):

aeron-archive/src/main/java/io/aeron/archive/client/ReplayMerge.java
```diff
@@ -506,6 +510,31 @@ else if (shouldStopAndRemoveReplay(position))
...
+        if (result)
+        {
+            activeCorrelationId = correlationId;
+        }
+
+        // increase backoff regardless of result
+        getMaxRecordedPositionBackoffMs = Long.min(
+            getMaxRecordedPositionBackoffMs * 2, GET_MAX_RECORDED_POSITION_BACKOFF_MAX_MS);
+        timeOfNextGetMaxRecordedPositionMs = nowMs + getMaxRecordedPositionBackoffMs;
```
vyazelenko, line 533: This back-off reset can happen already when transitioning to `REPLAY` state.

There might be state flickering back and force between `CATCHUP` and `ATTEMPT_LIVE_JOIN`. The latter performs those excessive calls to resolve position which we want to throttle.

nbradac: I'm not sure what you're getting at here.  I reset the backoff when moving to `CATCHUP` the first time.  But after that, each time the state flickers back and forth, I let the backoff increase.  It was that flickering that was previously causing the vast numbers of calls to `getRecordingPosition`.

vyazelenko: Just wanted to mention that we can reset the back-off counter one state earlier. But this won't change the outcome so I'm fine with either approach.

PR #1621, opened 2024-07-09
Tooling Part 1

marc-adaptive:
Move aeron tooling from aeron-samples into aeron-client package. Support existing java samples APIs.


aeron-client/src/main/java/io/aeron/reports/LossReport.java
```diff
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package io.aeron.driver.reports;
+package io.aeron.reports;
```
vyazelenko, line 16: Should be a sub-package of `io.aeron.tooling`. The latter probably should be named `io.aeron.tools`.

nickelaway:
I see we have the archive-specific SegmentInspector tool here. What about ArchiveTool? Maybe ClusterTool and ClusterControl, too?

PR #1623, opened 2024-07-10
modify the generation method of replay token

WorkingChen:
Using random numbers as replay tokens may lead to token conflicts, introducing instability. I think it could be replaced with the aeron.nextCorrelationId() method. Alternatively, changing to the following code would also be safer.

image


vyazelenko:
@WorkingChen Thanks for the report. This should be fixed in https://github.com/real-logic/aeron/commit/d548968156d6aada215c1f867fbdf8b07e1a6b23.

PR #1636, opened 2024-07-19
Archive client [C and C++ wrapper]

nbradac:
c-archive-client
c-archive-client-run
run-archive-client-run!!!


aeron-archive/src/main/cpp_wrapper/client/archive/ArchiveContext.h
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    template<typename IdleStrategy = aeron::concurrent::YieldingIdleStrategy>
+    void idleStrategy()
+    {
+        // TODO I have no idea what is happening here...
+
+        IdleStrategy idleStrategy;
+
+        m_idleFunc = [&idleStrategy](int work_count){
```
nbradac, line 92: I'm trying to compensate for the fact that the various implemented idle strategies don't all inherit from a single class, so I can't set an idle strategy as a parameter.  I can do it via a template, however.  I think this works, but it sure does feel ugly.  I'm creating a lambda that wraps around the idle strategy that shows up because of the template.  I have no idea what sorts of naughty behaviours might result from this.  But... maybe it's ok?

aeron-client/src/main/cpp_wrapper/Context.h
```diff
@@ -564,6 +564,12 @@ class Context
         return { path, 0, length };
     }
 
+    // TODO find a better way to expose this to ONLY the archive cpp_wrapper?
+    inline aeron_context_t *context() const
+    {
+        return m_context;
+    }
```
nbradac, line 571: Is it ok to just provide a public API that returns the backing aeron_context_t?

aeron-client/src/main/cpp_wrapper/util/Exceptions.h
```diff
@@ -138,7 +138,7 @@ AERON_DECLARE_SOURCED_EXCEPTION(UnknownSubscriptionException, ExceptionCategory:
 AERON_DECLARE_SOURCED_EXCEPTION(ReentrantException, ExceptionCategory::EXCEPTION_CATEGORY_ERROR);
 AERON_DECLARE_SOURCED_EXCEPTION(UnsupportedOperationException, ExceptionCategory::EXCEPTION_CATEGORY_ERROR);
 
-#define AERON_MAP_TO_SOURCED_EXCEPTION_AND_THROW(code, message)                     \
+#define AERON_MAP_TO_SOURCED_EXCEPTION_AND_THROW_WITH_DEFAULT(code, message, defaultException)  \
```
nbradac, line 141: This was reworked a bit so that I can throw an ArchiveException as the default when these macros are leveraged from the archive cpp_wrapper project.

aeron-archive/src/main/cpp_wrapper/client/archive/ArchiveContext.h
```diff
@@ -0,0 +1,147 @@
...
+            idleStrategy.idle(work_count);
+        };
+    }
+
+    aeron_archive_context_t *m_ctx = nullptr; // TODO this should be private
+
+private:
+    CredentialsSupplier m_credentialsSupplier;
+
+    aeron_archive_encoded_credentials_t m_lastEncodedCredentials = {};
```
nbradac, line 102: I need a credentials_t that I can fill out that can be used by whatever underlying archive client code needs credentials.  The upcall triggers a call to the C++ code, then I grab the values out of the 'pair' that's returned, fill out this struct, then pass the address back down.  I *believe* we can only get one of these upcalls at a time, so I think this is safe.  🤞

aeron-archive/src/main/cpp_wrapper/client/archive/ArchiveContext.h
```diff
@@ -0,0 +1,147 @@
...
+ * limitations under the License.
+ */
+#ifndef AERON_ARCHIVE_WRAPPER_CONTEXT_H
+#define AERON_ARCHIVE_WRAPPER_CONTEXT_H
+
+#include "AeronArchive.h"
+#include "CredentialsSupplier.h"
+#include "concurrent/YieldingIdleStrategy.h"
+
+#include "Context.h"
```
nbradac, line 23: I'd love for the aeron::archive::client::Context to be defined in a file called 'Context.h'.  But if I do that, then I have no way of including the Context.h from the base client cpp_wrapper project.  I'm not sure how to fix that... or even if it matters.

aeron-archive/src/main/cpp_wrapper/client/archive/ArchiveContext.h
nbradac: I think I figured this out.  The latest version of this function is simpler and makes more sense.

PR #1637, opened 2024-07-22
[C] Allocate a buffer for client name.

amoskvin:
Setting the client name with the C library causes a crash when closing the context because it tries to free a pointer that hasn't been heap-allocated. I fixed it by copying the name to a new buffer.

The issue can be reproduced with sample binaries:
```
$ AERON_CLIENT_NAME=oops ./binaries/cpong 
Subscribing Ping at channel aeron:udp?endpoint=localhost:20123 on Stream ID 1002
Publishing Pong at channel aeron:udp?endpoint=localhost:20124 on Stream ID 1003
Subscription channel status 1
Publication channel status 1
^Cmunmap_chunk(): invalid pointer
Aborted (core dumped)
```

Another option would be to just not free, but I think holding it in a new buffer is more robust and consistent with `aeron_context_set_dir()`.

### DIFF
--- a/aeron-client/src/main/c/aeron_context.c
+++ b/aeron-client/src/main/c/aeron_context.c
@@ -110,7 +110,13 @@ int aeron_context_init(aeron_context_t **context)
 
     if ((value = getenv(AERON_CLIENT_NAME_ENV_VAR)))
     {
-        _context->client_name = value;
+        size_t name_length = strlen(value);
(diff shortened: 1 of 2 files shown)


mikeb01:
This should be fix in [9b3e7cb](https://github.com/real-logic/aeron/commit/9b3e7cb17cc07b5a61e49ef7f09166747c7bc75c).

PR #1638, opened 2024-07-22
[Java] Reduce archive listRecordingsForUri idling

grddev:
With the recent reduction to MAX_SCANS_PER_WORK_CYCLE in https://github.com/real-logic/aeron/commit/1c558899c4d8e62b351adac9addd177d9d108f51, we noted that some list operations took significantly longer time, but when profiling it looked like archive was just sleeping.

As it turns out, list recordings only return a positive work count if it actually found something, but this means that a listRecordingsForUri API call that matches only a small subset of recordings (or even none at all) will idle every 64 recordings, which for a large catalog caused the list operation to take multiple seconds of sleeping.

This instead changes work count to be based on the number of recordings processed, which for listRecordings should be exactly the same, but for listRecordingsForUri should only idle if there are no more recordings to check at the moment.

### DIFF
--- a/aeron-archive/src/main/java/io/aeron/archive/AbstractListRecordingsSession.java
+++ b/aeron-archive/src/main/java/io/aeron/archive/AbstractListRecordingsSession.java
@@ -101,7 +101,7 @@ abstract class AbstractListRecordingsSession implements Session
             }
         }
 
-        final int alreadySent = sent;
+        final int batchStartPosition = position;
         for (int recordsScanned = 0; sent < count && recordsScanned < MAX_SCANS_PER_WORK_CYCLE; recordsScanned++)
         {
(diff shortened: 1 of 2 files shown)


mikeb01:
This idea looks good, one question though, why is the result divided by 2?
```
return (position - batchStartPosition) / 2;
```
Is this to return 0 work if we only scan a single record?

grddev:
> This idea looks good, one question though, why is the result divided by 2?

@mikeb01 My understanding of the code is that the index uses two positions per recording, and thus the result is divided by two to represent the number of recordings considered.

mikeb01:
> thus the result is divided by two to represent the number of recordings considered.

Understood, we typically don't worry too much about about specific values from doWork calls.

PR #1639, opened 2024-07-22
C++ wrapper improvements

ZachBray:
Josh raised a couple of issues through premium support tickets. More notes in the commit messages. Based on the README, I think it is okay to assume `>= C++11`, but please correct me if I am wrong.


mjpt777:
Yes we can assume C++ 11 as minimum.

PR #1640, opened 2024-07-22
[Java] add config option and uri param for setting max retransmits

nbradac:
Could use some feedback on uri param name and system property name.

Also, right now I'm grabbing the default right before passing the value to the RetransmitHandler.  I *could* change that so the PublicationParams object always ends up with a valid maxRetransmits value (set either via the uri, system property, or the default).  Not sure if matters much either way, but perhaps there are reasons for/against such a refactor.


aeron-driver/src/main/java/io/aeron/driver/PublicationParams.java
```diff
@@ -408,6 +411,34 @@ private void getUntetheredRestingTimeout(final ChannelUri channelUri, final Medi
...
+            maxRetransmits = Integer.parseInt(maxRetransmtsString);
+        }
+        catch (final NumberFormatException ex)
+        {
+            throw new IllegalArgumentException("invalid " + MAX_RETRANSMITS_PARAM_NAME + ", must be a number", ex);
+        }
+
+        if (maxRetransmits < 1)
+        {
+            throw new IllegalArgumentException("invalid " + MAX_RETRANSMITS_PARAM_NAME + ", must be > 0");
```
mikeb01, line 436: I think we should have an upper bound on this as well because it will affect the amount of space allocated.

aeron-driver/src/test/java/io/aeron/driver/RetransmitHandlerTest.java
```diff
@@ -77,8 +77,14 @@ class RetransmitHandlerTest
...
 
-    private RetransmitHandler handler = new RetransmitHandler(
-        () -> currentTime, invalidPackets, DELAY_GENERATOR, LINGER_GENERATOR, true, retransmitOverflow);
+    private RetransmitHandler handler = new RetransmitHandler(() ->
+        currentTime,
+        invalidPackets,
+        DELAY_GENERATOR,
+        LINGER_GENERATOR,
+        true,
+        Configuration.maxRetransmits(),
```
mikeb01, line 86: I'd prefer this to be and explicit value in the test as we could end up with non-reproducible tests if system properties started creeping in from somewhere.

aeron-client/src/main/java/io/aeron/ChannelUriStringBuilder.java
```diff
@@ -2154,6 +2157,56 @@ public Long untetheredRestingTimeoutNs()
...
+        else
+        {
+            try
+            {
+                return maxRetransmits(Integer.parseInt(valueStr));
+            }
+            catch (final NumberFormatException ex)
+            {
+                throw new IllegalArgumentException(
+                    MAX_RETRANSMITS_PARAM_NAME + " must be a number");
```
vyazelenko, line 2195: Add original exception as a cause like other methods do (e.g. `termOffset`).

aeron-driver/src/main/c/aeronmd.h
```diff
@@ -637,6 +637,14 @@ uint64_t aeron_driver_context_get_nak_unicast_delay_ns(aeron_driver_context_t *c
 int aeron_driver_context_set_nak_unicast_retry_delay_ratio(aeron_driver_context_t *context, uint64_t value);
 uint64_t aeron_driver_context_get_nak_unicast_retry_delay_ratio(aeron_driver_context_t *context);
 
+/**
+ * Max number of active retransmissions tracked for udp streams with group semantics.
+ */
+#define AERON_MAX_RETRANSMITS_ENV_VAR "AERON_RETRANSMITS_ACTIVE_MAX"
```
vyazelenko, line 643: The name of the variable does not match the value, why? For all other parameters we use the `VALUE_ENV_VAR="VALUE"`.

aeron-client/src/main/c/uri/aeron_uri.h
```diff
@@ -75,6 +75,7 @@ aeron_uri_params_t;
 #define AERON_URI_NAK_DELAY_KEY "nak-delay"
 #define AERON_URI_UNTETHERED_WINDOW_LIMIT_TIMEOUT_KEY "untethered-window-limit-timeout"
 #define AERON_URI_UNTETHERED_RESTING_TIMEOUT_KEY "untethered-resting-timeout"
+#define AERON_URI_MAX_RETRANSMITS_KEY "retransmits-active-max"
```
vyazelenko, line 78: Variable name does not match the value. Other keys use `AERON_URI_VALUE_KEY "value"`.

aeron-client/src/main/cpp/ChannelUri.h
```diff
@@ -71,6 +71,7 @@ static constexpr const char RESPONSE_CORRELATION_ID_PARAM_NAME[] = "response-cor
 static constexpr const char NAK_DELAY_PARAM_NAME[] = "nak-delay";
 static constexpr const char UNTETHERED_WINDOW_LIMIT_TIMEOUT_PARAM_NAME[] = "untethered-window-limit-timeout";
 static constexpr const char UNTETHERED_RESTING_TIMEOUT_PARAM_NAME[] = "untethered-resting-timeout";
+static constexpr const char MAX_RETRANSMITS_PARAM_NAME[] = "retransmits-active-max";
```
vyazelenko, line 74: Name and value do not match unlike other parameters.

aeron-client/src/main/cpp_wrapper/ChannelUri.h
```diff
@@ -71,6 +71,7 @@ static constexpr const char RESPONSE_CORRELATION_ID_PARAM_NAME[] = "response-cor
 static constexpr const char NAK_DELAY_PARAM_NAME[] = "nak-delay";
 static constexpr const char UNTETHERED_WINDOW_LIMIT_TIMEOUT_PARAM_NAME[] = "untethered-window-limit-timeout";
 static constexpr const char UNTETHERED_RESTING_TIMEOUT_PARAM_NAME[] = "untethered-resting-timeout";
+static constexpr const char MAX_RETRANSMITS_PARAM_NAME[] = "retransmits-active-max";
```
vyazelenko, line 74: Naming.

aeron-client/src/main/java/io/aeron/CommonContext.java
```diff
@@ -419,6 +419,13 @@ public static InferableBoolean parse(final String value)
      */
     public static final String UNTETHERED_RESTING_TIMEOUT_PARAM_NAME = "untethered-resting-timeout";
 
+    /**
+     * Parameter name to set the max number of active retransmits for a publication
+     *
+     * @since 1.45.0
+     */
+    public static final String MAX_RETRANSMITS_PARAM_NAME = "retransmits-active-max";
```
vyazelenko, line 427: Naming. Why the value is `retransmits-active-max` and not simply `max-retransmits` or `retransmits-max`?
In any case the variable name should reflect the value like we have for other parameters.

aeron-driver/src/main/c/uri/aeron_driver_uri.c
```diff
@@ -48,6 +48,37 @@ int aeron_uri_get_term_length_param(aeron_uri_params_t *uri_params, aeron_driver
...
+
+        if (-1 == aeron_parse_size64(value_str, &value))
+        {
+            AERON_SET_ERR(EINVAL, "could not parse %s=%s in URI", AERON_URI_MAX_RETRANSMITS_KEY, value_str);
+            return -1;
+        }
+
+        if (value < 1 || value > AERON_RETRANSMIT_HANDLER_MAX_RETRANSMITS_MAX)
+        {
+            AERON_SET_ERR(
```
vyazelenko, line 67: Would be good to include an actual value that fails this validation.

aeron-driver/src/main/java/io/aeron/driver/PublicationParams.java
```diff
@@ -408,6 +411,37 @@ private void getUntetheredRestingTimeout(final ChannelUri channelUri, final Medi
...
+            throw new IllegalArgumentException(
+                "invalid " + RETRANSMITS_ACTIVE_MAX_PARAM_NAME + ", must be a number", ex);
+        }
+
+        if (retransmitsActiveMax < 1 || retransmitsActiveMax > Configuration.RETRANSMITS_ACTIVE_MAX_MAX)
+        {
+            throw new IllegalArgumentException(
+                "invalid " + RETRANSMITS_ACTIVE_MAX_PARAM_NAME + ", must be > 0 and <= " +
+                    Configuration.RETRANSMITS_ACTIVE_MAX_MAX);
+        }
```
vyazelenko, line 440: please add actual value to the exception message.

aeron-driver/src/main/c/uri/aeron_driver_uri.c
```diff
@@ -48,6 +48,38 @@ int aeron_uri_get_term_length_param(aeron_uri_params_t *uri_params, aeron_driver
...
+        {
+            AERON_SET_ERR(EINVAL, "could not parse %s=%s in URI", AERON_URI_RETRANSMITS_ACTIVE_MAX_KEY, value_str);
+            return -1;
+        }
+
+        if (value < 1 || value > AERON_RETRANSMIT_HANDLER_RETRANSMITS_ACTIVE_MAX_MAX)
+        {
+            AERON_SET_ERR(
+                EINVAL,
+                "invalid %s=%llu, must be > 0 and <= %i",
```
mikeb01, line 69: We should use the [standardised fixed width integer type](https://en.cppreference.com/w/c/types/integer) macros for printing, e.g. in this case PRIu64.

aeron-driver/src/main/java/io/aeron/driver/DriverConductor.java
```diff
@@ -1640,6 +1640,7 @@ private NetworkPublication newNetworkPublication(
                 ctx.retransmitUnicastDelayGenerator(),
                 ctx.retransmitUnicastLingerGenerator(),
                 udpChannel.hasGroupSemantics(),
+                params.hasRetransmitsActiveMax ? params.retransmitsActiveMax : Configuration.retransmitsActiveMax(),
```
mikeb01, line 1643: The fallback value should be derived from the driver context, the `Configuration.retransmitsActiveMax()` should be used to set initial value on the context.

nbradac: Good catch!

aeron-driver/src/main/c/uri/aeron_driver_uri.c
```diff
@@ -48,6 +48,38 @@
...
+        }
+
+        if (value < 1 || value > AERON_RETRANSMIT_HANDLER_MAX_RESEND_MAX)
+        {
+            AERON_SET_ERR(
+                EINVAL,
+                "invalid %s=" PRIu64 ", must be > 0 and <= %i",
+                AERON_URI_MAX_RESEND_KEY,
+                value,
+                AERON_RETRANSMIT_HANDLER_MAX_RESEND_MAX);
```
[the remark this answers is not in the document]
nbradac: D'oh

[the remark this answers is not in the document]
mjpt777: Should it be resend hander and not retransmit handler?

mikeb01:
LGTM

aeron-driver/src/main/c/aeron_retransmit_handler.c
```diff
@@ -34,13 +34,14 @@ int aeron_retransmit_handler_init(
     uint64_t delay_timeout_ns,
     uint64_t linger_timeout_ns,
     bool has_group_semantics,
+    uint32_t max_retransmits,
```
vyazelenko, line 37: Shouldn't this be called `max_resend` now?

aeron-driver/src/main/c/aeron_retransmit_handler.h
```diff
@@ -66,6 +67,7 @@ int aeron_retransmit_handler_init(
     uint64_t delay_timeout_ns,
     uint64_t linger_timeout_ns,
     bool has_group_semantics,
+    uint32_t max_retransmits,
```
vyazelenko, line 70: `max_resend`?

vyazelenko:
LGTM

PR #1643, opened 2024-07-23
[Java] modify onStart callback for ConsensusModuleExtension.

WorkingChen:
Ensure the onStart method is executed at program startup, regardless of whether a snapshot exists or not.


mjpt777:
This is not an end user API. It is experimental and will be marked so in the next release. It's purpose is to support Aeron product extensions we have have as separate offerings.

PR #1647, opened 2024-08-09
Feature/recording log version

mikeb01:
* Adds support for version the RecordingLog file.

Because the RecordingLog file doesn't contain a header, we need to introduce a marker value (magic number) so that we can distinguish between the files with versioning and those without.  This will check for that value and if not present will migrate the recording to the new structure (which includes a header that has a version field).


aeron-cluster/src/test/java/io/aeron/cluster/RecordingLogVersioningTest.java
```diff
@@ -0,0 +1,230 @@
...
+            final int entryType,
+            final long recordingId,
+            final long leadershipTermId,
+            final long termBaseLogPosition,
+            final long logPosition,
+            final long timestamp,
+            final int serviceId,
+            final String endpoint)
+        {
+
```
mjpt777, line 77: Blank line

mjpt777 (approved):
Some minor comments left. 

One thing that occurs to me is an entry type of NULL_VALUE could be used with one of the other fields used to store the version.

pveentjer:
@mikeb01 do you want this feature to be merged as part of 1.47.0?

mikeb01:
> do you want this feature to be merged as part of 1.47.0?

No, does not need to be.

PR #1654, opened 2024-09-04
[Java] load snapshot with no services fixes

langera:


mikeb01:
Is it possible to produce a system test to show that without these fixes Cluster breaks?

mjpt777:
A better approach is taken here https://github.com/real-logic/aeron/commit/7606549b525ec6bb6939a1a744095cf0b574e990.

PR #1658, opened 2024-09-12
[Java] Allow aeron archive client to reside behind a proxy

frantahaas:
Hey, I'm currently using following workaround to have an Aeron archive client running behind a loadbalancer in k8s. Essentially, the AeronArchive does not allow me to specify a different channel to receive (subscribe) data (locally) and another channel (loadbalancer) that I want the Archive to actually send data to.

Would you allow something similar to this get integrated into the mainline? Or suggest a proper solution for my use-case? Thanks a lot

### DIFF
--- a/aeron-archive/src/main/java/io/aeron/archive/client/AeronArchive.java
+++ b/aeron-archive/src/main/java/io/aeron/archive/client/AeronArchive.java
@@ -1258,6 +1258,60 @@ public final class AeronArchive implements AutoCloseable
      * Replay a length in bytes of a recording from a position and for convenience create a {@link Subscription}
      * to receive the replay. If the position is {@link #NULL_POSITION} then the stream will be replayed from the start.
      *
+     * Use in case when the replay client resides behind a proxy.
+     *
+     * @param recordingId    to be replayed.
+     * @param position       from which the replay should begin or {@link #NULL_POSITION} if from the start.
+     * @param length         of the stream to be replayed or {@link Long#MAX_VALUE} to follow a live recording.
+     * @param replayChannel  which the replay should be subscribed locally.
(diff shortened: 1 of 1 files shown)


mikeb01:
While the feature is still experimental, you can look at using [ResponseChannels](https://github.com/real-logic/aeron/blob/master/aeron-system-tests/src/test/java/io/aeron/archive/ArchiveResponseClientTest.java) to connect to the Archive.

frantahaas:
Thanks for the reply. So, I've looked at those tests and tried something out which works locally but didn't manage to get it working in k8s.

Our topology is following:

Client <--- Proxy <--- Archive
Client --------------> Archive

On the proxy, I've two ports open, one for the client control response 2022 and one for replay channel 2033.

How exactly am I supposed to configure the control on the Archive? In other terms, is the Archive supposed to have direct access to that uri? If so, does that actually help me in my case?

```
        final AeronArchive.Context aeronArchiveCtx = new AeronArchive.Context()
            .controlRequestChannel(archive.context().controlChannel())
            .controlResponseChannel("aeron:udp?control-mode=response|control=localhost/ClientIP/ProxyIP/ArchiveIP/:10002");
```

In my hack, I'm actually opening subscriptions on the Client's IP while sending Archive the Proxy's IP to response and replay to which gets fwded to Client.

vyazelenko:
@frantahaas To use response channels feature you need configure `control` on the Archive to point to the Archive's IP.
```java
final AeronArchive.Context aeronArchiveCtx = new AeronArchive.Context()
      .controlRequestChannel("aeron:udp?endpoint=ArchiveIP:10001")
      .controlResponseChannel("aeron:udp?control-mode=response|control=ArchiveIP:10002");
```

PR #1659, opened 2024-09-13
[Java] refactor cluster tool to allow extension

langera:
Refactoring of ClusterTool to allow extension by premium tools + reuse of command operations.

The operations are separated to their own class `ClusterToolOperator` which can be used on its own + extended.

There is an interface `ClusterTool.Action` to define one action + a class `ClusterTool.ClusterToolCommand` which defines an action + its description.

`ClusterTool` main simply executes any `command` that exists in a predefined Map (by name == the arg for the command).

This preserves backward compatibility and allows to compose other mains of new/modified/reused commands.  
Except `main`, all the other public static methods in `ClusterTool` also kept for backward compatibility and delegate to the `ClusterToolOperator`

### DIFF
--- a/aeron-cluster/src/main/java/io/aeron/cluster/ClusterTool.java
+++ b/aeron-cluster/src/main/java/io/aeron/cluster/ClusterTool.java
@@ -15,57 +15,25 @@
  */
 package io.aeron.cluster;
 
-import io.aeron.*;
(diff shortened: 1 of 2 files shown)


langera:
Closing. Not ready yet on the premium side

PR #1672, opened 2024-10-23
Tethering join position

marc-adaptive:
For untethered subscriptions transitioning from resting to active, avoid using their own position in minimum join position


callmerockett:
Hi @marc-adaptive Did you notice some misbehavior due to this rejoining at an old position? I'm investigating 
a back pressure for indefinite time in the scenario with untethered subscriptions on older version of aeron driver and client and thought that could be related.

marc-adaptive:
@callmerockett Yes an untethered subscription could rejoin at an invalid position. Could you please try again using the latest driver from 1.44.6, 1.45.1, 1.46.7, or master?

PR #1674, opened 2024-10-25
Use provided credentials on standby snapshot

TAregger:
Currently the cluster is not able to receive standby snapshot notifications when using an authenticator that checks the credentials.
This is because the provided credentials on standby snapshot notifications are not read.


vyazelenko:
@TAregger good catch.

Issue #1690, opened 2024-11-16
AeronCluster client never reconnects to the cluster after all nodes have been stopped


the-thing:
Aeron version: 1.44.1
Java 17

Currently when running AeronCluster client against a single node cluster it is not possible to recover connection after stopping the node and starting it back again.

Client is constantly running and AeronCluster client agent is continuously polling for egress messages and sending keep alive messages, but it seems that it never receives the leadership event `io.aeron.cluster.client.AeronCluster#onNewLeader` for a single node cluster which might be the problem here.

Keep alive results are taken from debugger logging output at https://github.com/real-logic/aeron/blob/55535ce818cc82aedb08df10af3843d81d9cedfd/aeron-cluster/src/main/java/io/aeron/cluster/client/AeronCluster.java#L447

```
Keep alive result: 896
Keep alive result: 960
Keep alive result: 1024
Keep alive result: 1088
Keep alive result: 1152
[single node cluster stopped here]
Keep alive result: -4
Keep alive result: -4
Keep alive result: -4
2024-11-16 16:01:39.699 [test-client-7263566828269871104] WARN BufferedConnection - Failed to send keep alive message [lastHeartbeatSentTimeNs=23356506689600,nowNs=23357507749100,aeronClusterClosed=false,ingressPublicationClosed=true,egressSubscriptionClosed=false]
Keep alive result: -4
```

I understand that running a single node cluster doesn't make much sense, except when running some local end to end tests, but in this case in my connection recovery code I have to check for `io.aeron.cluster.client.AeronCluster#sendKeepAlive` returning false (which can also be due to back pressure) and also checking for `io.aeron.cluster.client.AeronCluster#ingressPublication` close state to recreate the cluster client via `io.aeron.cluster.client.AeronCluster#asyncConnect(io.aeron.cluster.client.AeronCluster.Context)`. Creating the new AeronCluster client seems to be the only way to reconnect to cluster after node restart.

This doesn't seem to be a problem for a 3 node cluster. Stopping 2 out of 3 nodes and later starting one back again seems to recover automatically.

```
Keep alive result: 1088
Keep alive result: 1152
[two of the cluster nodes stopped]
Keep alive result: -1
Keep alive result: -1
Keep alive result: -1
[...]

PR #1691, opened 2024-11-21
[Java] AeronCluster client does not close itself after whole cluster shutdown

the-thing:
Failing tests scenario discussed in https://github.com/real-logic/aeron/issues/1690 (slow test)


the-thing:
Thanks. I will check it out. Was this the fix d11fb360ae77439bdade105e5cde062d5b5cf743?

wojciech-adaptive:
> @wojciech-adaptive Thanks. I will check it out. Was this the fix [d11fb36](https://github.com/aeron-io/aeron/commit/d11fb360ae77439bdade105e5cde062d5b5cf743)?

Yes.

PR #1694, opened 2024-12-06
Add snd/rcv buffer length to LogBuffer

pveentjer:


aeron-client/src/main/java/io/aeron/logbuffer/LogBufferDescriptor.java
```diff
@@ -135,6 +135,17 @@ public class LogBufferDescriptor
      */
     public static final int LOG_DEFAULT_FRAME_HEADER_OFFSET;
 
+    /**
+     * Offset within the log metadata where the receive buffer size (rcvBuf) is stored.
+     */
+    public static final int LOG_RCVBUF_LENGTH_OFFSET;
```
vyazelenko, line 141: Naming, i.e. should be `LOG_SOCKET_RCVBUF_LENGTH_OFFSET`.

aeron-client/src/main/java/io/aeron/logbuffer/LogBufferDescriptor.java
```diff
@@ -135,6 +135,17 @@ public class LogBufferDescriptor
...
+
+    /**
+     * Offset within the log metadata where the send buffer size (sndBuf) is stored.
+     */
+    public static final int LOG_SNDBUF_LENGTH_OFFSET;
```
vyazelenko, line 146: Naming, i.e. should be `LOG_SOCKET_SNDBUF_LENGTH_OFFSET`.

aeron-client/src/main/java/io/aeron/logbuffer/LogBufferDescriptor.java
```diff
@@ -908,4 +925,48 @@ public static int computeAssembledFrameLength(final int length, final int maxPay
...
+        metadataBuffer.putInt(LOG_RCVBUF_LENGTH_OFFSET, rcvBuf);
+    }
+
+    /**
+     * Get the value of the send buffer length for this log.
+     *
+     * @param metadataBuffer containing the meta data.
+     * @return the value of the send buffer length.
+     */
+    public static int sndbufLength(final UnsafeBuffer metadataBuffer)
```
vyazelenko, line 957: Should be named the same as in other places, i.e. `socketSndbufLength`.

aeron-client/src/main/java/io/aeron/logbuffer/LogBufferDescriptor.java
```diff
@@ -908,4 +925,48 @@ public static int computeAssembledFrameLength(final int length, final int maxPay
...
+        return metadataBuffer.getInt(LOG_RCVBUF_LENGTH_OFFSET);
+    }
+
+    /**
+     * Set the value of the socket receive buffer length for this log.
+     *
+     * @param metadataBuffer containing the meta data.
+     * @param rcvBuf         value to be set.
+     */
+    public static void rcvbufLength(final UnsafeBuffer metadataBuffer, final int rcvBuf)
```
vyazelenko, line 946: Should be named the same as in other places, i.e. `socketRcvbufLength`.

aeron-driver/src/main/java/io/aeron/driver/DriverConductor.java
```diff
@@ -1764,15 +1764,44 @@ private NetworkPublication newNetworkPublication(
...
+        final int sndbufLength = channelEndpoint.socketSndbufLength();
+        if (sndbufLength == 0)
+        {
+            return ctx.osDefaultSocketSndbufLength();
+        }
+        else
+        {
+            return sndbufLength;
+        }
+    }
```
vyazelenko, line 1778: There is already a method `io.aeron.driver.media.UdpChannel#socketSndbufLengthOrDefault`. However, getting the correct value is more nuanced, i.e.:
* For network publication it should be taken from `SendChannelEndpoint#socketSndbufLength` instead.
* For IPC publication the socket buffers make no sense.
* For publication image the values should be taken from the `ReceiveChannelEndpoint`.

aeron-driver/src/main/java/io/aeron/driver/DriverConductor.java
```diff
@@ -1764,15 +1764,44 @@ private NetworkPublication newNetworkPublication(
...
+
+    private int getRcvbufLength(final UdpChannelTransport channelEndpoint)
```
vyazelenko, line 1780: See the above comment on `getSndbufLength`.

vyazelenko (changes requested):

aeron-client/src/main/java/io/aeron/logbuffer/LogBufferDescriptor.java
pveentjer: Done

pveentjer: Done

aeron-driver/src/main/java/io/aeron/driver/DriverConductor.java
pveentjer: Done

aeron-client/src/main/java/io/aeron/logbuffer/LogBufferDescriptor.java
pveentjer: Done

aeron-client/src/main/java/io/aeron/logbuffer/LogBufferDescriptor.java
pveentjer: Done

aeron-driver/src/main/java/io/aeron/driver/DriverConductor.java
pveentjer: I just figured out that Aeron will detect setting a socket buffer length that is larger than the maximum.

```
WARNING: Could not set desired SO_SNDBUF, adjust OS to allow aeron.socket.so_sndbuf attempted=67108864, actual=212992
WARNING: Could not set desired SO_RCVBUF, adjust OS to allow aeron.socket.so_rcvbuf attempted=67108864, actual=212992
```

pveentjer: The problem with the channelEndpoint.socketRcvbufLength() and channelEndpoint.socketSndbufLength() is that these methods can't be relied upon in the case that the configured value is larger than the OS maximum. This is the value that is configured; but not the value that is actually used.

The code to get the actual values:
```
           System.out.println("actual.sndBufLength:" + sendDatagramChannel.socket().getSendBufferSize());
            System.out.println("actual.rcvBufLength:" + receiveDatagramChannel.socket().getReceiveBufferSize());
```

And the corresponding output:

```
channelEndpoint.socketRcvbufLength:67108864
channelEndpoint.socketSndbufLength:67108864
actual.sndBufLength:212992
actual.rcvBufLength:212992
```

So the actual values are the value that need to be published in the log buffer. WDYT?

vyazelenko: @pveentjer [UdpChannelTransport#openDatagramChannel](https://github.com/real-logic/aeron/blob/2f8e497a97b850e2786f6a65397ead859ccdb680/aeron-driver/src/main/java/io/aeron/driver/media/UdpChannelTransport.java#L236-L244) is where we assign `SO_SNDBUF` and `SO_RCVBUF`.
I don't know if it is possible to get the actual values used after the configuration completes (at least from the Java API).

vyazelenko: I think we should go with the current approach.

vyazelenko (changes requested):
@pveentjer The same changes need to be implemented in the C media driver.

aeron-driver/src/main/java/io/aeron/driver/DriverConductor.java
pveentjer: To get the actual values, use:

```
    System.out.println("actual.sndBufLength:" + sendDatagramChannel.socket().getSendBufferSize());
    System.out.println("actual.rcvBufLength:" + receiveDatagramChannel.socket().getReceiveBufferSize());
```

pveentjer:
Closing this one since we are going for a different approach (include it in the channel-uri and store the channel uri in the log buffer)

PR #1705, opened 2025-01-08
[Cluster] rename the cluster-markfile sbe NULL definition to compatible with C++

WorkingChen:


vyazelenko:
The value is now called `UNKNOWN`. See https://github.com/real-logic/aeron/commit/2f9520935f404db922f6fedb1cae8eb4cd64566f.

Issue #1708, opened 2025-01-10
Aeron C archive client segfaults if the archive async connect times out


matteo-gsr:
![image](https://github.com/user-attachments/assets/aa173783-cea8-499a-b621-b13427510171)

issue is that the exclusive publication conductor is null and when it blows up when it gets into the aeron client `aeron_client_conductor_offer_remove_command`.

Edit: there's a double free issue when we try to free the exclusive_publication twice, first from the async instance and then from the archive_proxy one.

vyazelenko:
@matteo-gsr Should be fixed with https://github.com/real-logic/aeron/commit/754104b42b398c5a49d01b00209f191504f476bc.

PR #1709, opened 2025-01-10
[C] archive client double-free issue on async connect timeout

matteo-gsr:
see https://github.com/real-logic/aeron/issues/1708.


vyazelenko:
Already addressed by https://github.com/real-logic/aeron/commit/754104b42b398c5a49d01b00209f191504f476bc.

PR #1713, opened 2025-01-14
Replacement of ThreadHints.onSpinWait by Thread.

pveentjer:
The ThreadHints class in Agrona has been deprecated and all call to ThreadHints.onSpinWait have been replaced by Thread.onSpinWait.

For more info see:

https://github.com/real-logic/agrona/pull/312

### DIFF
--- a/aeron-samples/src/main/java/io/aeron/samples/raw/BurstSendReceiveUdpPing.java
+++ b/aeron-samples/src/main/java/io/aeron/samples/raw/BurstSendReceiveUdpPing.java
@@ -21,7 +21,6 @@ import org.agrona.BitUtil;
 import org.agrona.SystemUtil;
 import org.agrona.concurrent.HighResolutionTimer;
 import org.agrona.concurrent.SigInt;
-import org.agrona.hints.ThreadHints;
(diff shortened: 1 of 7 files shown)


aeron-driver/src/main/java/io/aeron/driver/DriverConductor.java
```diff
@@ -1916,6 +1916,11 @@ private void initLogMetadata(
     {
         final UnsafeBuffer logMetaData = rawLog.metaData();
 
+        System.out.println("SessionId: " + sessionId +
+            " streamId:" + streamId +
+            " initialTermId:" + initialTermId +
+            " mtuLength:" + mtuLength);
+
```
vyazelenko, line 1923: What is this?

vyazelenko (changes requested):

aeron-driver/src/main/java/io/aeron/driver/DriverConductor.java
pveentjer: Debug code for Aeron insights. I already removed it.

PR #1718, opened 2025-01-15
[WIP] Adding missing fields to the logbuffer metadata

pveentjer:
The following fields are missing:

    int64_t entity_tag;
    int64_t response_correlation_id;
    uint8_t group;
    uint8_t is_response;


nbradac:
These changes were taken care of via a different PR.

Issue #1732, opened 2025-01-26
Executor thread - cpu affinity not working


matteo-gsr:
Hi,

I am using the embedded C media driver in dedicated threading mode and I am setting a CPU affinity on each of the conductor, sender and receiver threads.

It seems there's an aeron_executor thread which doesn't get the affinity. It looks like it's spawn by the conductor agent. However, it's not getting the same affinity.

It is problematic given it then ends up on the same isolated core of the critical thread, which I really don't want to.

vyazelenko:
Fixed with https://github.com/aeron-io/aeron/commit/ae665d9e53eba4427b3f64b6549caa985fa3cf11.

There is now a separate configuration option to set CPU affinity for the async executor thread, i.e. via `aeron.driver.async.executor.cpu.affinity` property or `AERON_DRIVER_ASYNC_EXECUTOR_CPU_AFFINITY` env variable.

matteo-gsr:
@vyazelenko - sorry to revive this thread. Is it possible to add the setter methods to set the affinity? 

e.g.

https://github.com/aeron-io/aeron/blob/master/aeron-driver/src/main/c/aeron_driver_context.c#L3085

Ideally, one would have the setters for these members:

* set_sender_cpu_affinity_no
* set_receiver_cpu_affinity_no
* set_conductor_cpu_affinity_no
* set_async_executor_cpu_affinity_no

Thank you.

vyazelenko:
@matteo-gsr I've remove the feature to assign explicit thread affinity for the async executor thread in the upcoming 1.48.0 release as we want to treat it as any other auxiliary thread, i.e. it should be scheduled freely by the OS scheduler on non-isolated CPU cores.

vyazelenko:
@matteo-gsr You've mentioned in a comment to https://github.com/aeron-io/aeron/commit/b655c1122b7d85963e1a577e71fa25fa7acfce11 that you needed to set the affinity for the `aeron_executor` thread, because it was otherwise running one of the cores where a critical thread was running. 

[...]

PR #1737, opened 2025-02-04
[C] Adding setter methods for CPU affinity to media driver

matteo-gsr:
see #1732. 🙏


matteo-gsr:
> @matteo-gsr For each setter method there must be a getter method defined and both methods should be declared in [aeronmd.h](https://github.com/aeron-io/aeron/blob/master/aeron-driver/src/main/c/aeronmd.h) file.

thanks @vyazelenko - I amended the MR. Let me know if anything else is missing.

PR #1741, opened 2025-02-11
[C++ Wrapper] add DistinctErrorLog

WorkingChen:


WorkingChen:
> A C++ Wrapper implementation should wrap a C implementation of the corresponding component, i.e. [aeron_distinct_error_log.h](https://github.com/aeron-io/aeron/blob/master/aeron-client/src/main/c/concurrent/aeron_distinct_error_log.h) in this case.

The modification has been made. Please review it again. 😄

PR #1759, opened 2025-03-05
fix: check socket error codes to avoid overcounting short sends (fixes #1756)

### LINKED ISSUE #1756: Aeron socket/transport return codes lose context, resulting in erroneous short send reports
Methods such as https://github.com/aeron-io/aeron/blob/8d8096d12b50363939bad2f0234ade5af7347dcb/aeron-client/src/main/c/aeron_socket.c#L137 return 0 when the underlying socket returns `EAGAIN, EWOULDBLOCK, ECONNREFUSED, EINTR`. This results in methods that check for short sends with logic equivalent to `0 <= result && bytes_sent < (int64_t)iov.iov_len` erroneously reporting a short send. This is notably problematic for `aeron_network_publication_setup_message_check` and `aeron_network_publication_heartbeat_message_check`, as `ECONNREFUSED` is a likely status in both cases when using multicast.

https://github.com/aeron-io/aeron/blob/bf3af7ce34b9dcfa1a23cf04382205c87738364f/aeron-driver/src/main/c/aeron_network_publication.c#L471-L477
https://github.com/aeron-io/aeron/blob/bf3af7ce34b9dcfa1a23cf04382205c87738364f/aeron-driver/src/main/c/aeron_network_publication.c#L526-L533

I'm happy to submit a patch for this, but I'd appreciate guidance on the preferred approach to fixing it. Threading `errno` through the relevant functions, e.g.

```
aeron_network_publication_setup_message_check
aeron_network_publication_do_send
aeron_send_channel_send
aeron_udp_channel_transport_send
aeron_udp_channel_transport_send_connected
aeron_send
```

will break many function signatures, and changing the semantics of the send and receive return codes is similarly evasive. Checking `errno` at the caller as part of the guard clauses around calls to `aeron_counter_get(*->short_sends_counter)` would work, as I don't see anything in the call stack clobbering `errno`, but it's a little fragile. Reclassifying short sends as `0 <= result && bytes_sent > 0 && bytes_sent < (int64_t)iov.iov_len` would work, but trade false negatives for false positives.

klittlepage:
Connected Aeron sockets can produce zero-sized reads and writes in non-exceptional situations, e.g., when performing a network publication setup check against a host that hasn't started listening. This commit adds a check for non-exceptional socket errors to prevent incrementing the short send counter in such situations.

Fixes https://github.com/aeron-io/aeron/issues/1756

### DIFF
--- a/aeron-client/src/main/c/aeron_socket.c
+++ b/aeron-client/src/main/c/aeron_socket.c
@@ -120,7 +120,7 @@ ssize_t aeron_sendmsg(aeron_socket_t fd, struct msghdr *msghdr, int flags)
 
     if (result < 0)
     {
-        if (EAGAIN == errno || EWOULDBLOCK == errno || ECONNREFUSED == errno || EINTR == errno)
+        if (aeron_is_acceptable_socket_error())
         {
             return 0;
         }
@@ -140,7 +140,7 @@ ssize_t aeron_send(aeron_socket_t fd, const void *buf, size_t len, int flags)
 
     if (result < 0)
     {
-        if (EAGAIN == errno || EWOULDBLOCK == errno || ECONNREFUSED == errno || EINTR == errno)
+        if (aeron_is_acceptable_socket_error())
         {
             return 0;
         }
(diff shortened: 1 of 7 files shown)


vyazelenko:
See https://github.com/aeron-io/aeron/issues/1756#issuecomment-3804968516.

PR #1767, opened 2025-03-23
[C++ Wrapper] fix uri_buffer length in Subscription.tryResolveChannelEndpointPort()

WorkingChen:


WorkingChen:
This method returns channel, so it is wrong to request a buffer length of #define AERON_CLIENT_MAX_LOCAL_ADDRESS_STR_LEN (64)

image

PR #1770, opened 2025-03-24
MDC short send fix

nbradac:
The Java media driver sets the initial result ('bytes sent' effectively) to the number of bytes to be sent in the two implementations of `MultiSndDestination#send()`.  The first commit here does the same for the equivalent function in the C media driver.  The second commit is a new system test that fails when run against the C media driver before the fix.

### DIFF
--- a/aeron-driver/src/main/c/media/aeron_udp_destination_tracker.c
+++ b/aeron-driver/src/main/c/media/aeron_udp_destination_tracker.c
@@ -109,7 +109,7 @@ int aeron_udp_destination_tracker_send(
     int result = (int)iov_length;
     int to_be_removed = 0;
 
-    *bytes_sent = 0;
+    *bytes_sent = iov->iov_len;
 
     int starting_index = tracker->round_robin_index++;
     if (starting_index >= length)
--- a/aeron-system-tests/src/test/java/io/aeron/SpySimulatedConnectionTest.java
+++ b/aeron-system-tests/src/test/java/io/aeron/SpySimulatedConnectionTest.java
@@ -288,8 +288,22 @@ class SpySimulatedConnectionTest
         spy = client.addSubscription(spyForChannel(channel), STREAM_ID);
         publication = client.addExclusivePublication(channel + "|ssc=true", STREAM_ID);
 
-        Tests.awaitConnected(spy);
-        Tests.awaitConnected(publication);
+        while (!spy.isConnected() || !publication.isConnected())
+        {
+            Tests.yield();
+        }
+
+        while (publication.offer(buffer, 0, buffer.capacity()) < 0L)
+        {
+            Tests.yield();
+        }
+
+        final FragmentHandler mockFragmentHandler = mock(FragmentHandler.class);
+
+        while (spy.poll(mockFragmentHandler, 1) == 0)
+        {
+            Tests.yield();
+        }
 
         assertEquals(0, driver.counters().getCounterValue(SystemCounterDescriptor.SHORT_SENDS.id()));
     }
--- a/aeron-system-tests/src/test/java/io/aeron/SpySimulatedConnectionTest.java
+++ b/aeron-system-tests/src/test/java/io/aeron/SpySimulatedConnectionTest.java
@@ -38,7 +38,6 @@ import org.junit.jupiter.params.provider.MethodSource;
 
 import java.util.List;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicInteger;
 
 import static java.util.Arrays.asList;
 import static org.junit.jupiter.api.Assertions.*;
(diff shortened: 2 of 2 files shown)


aeron-system-tests/src/test/java/io/aeron/SpySimulatedConnectionTest.java
```diff
@@ -276,6 +278,35 @@ else if (!isSubscriptionClosed)
...
+        Thread.sleep(500);
+
+        AtomicInteger shortSendsCounterId = new AtomicInteger(0);
+        driver.counters().forEach((counterId, typeId, keyBuffer, label) -> {
+            if (AeronCounters.DRIVER_SYSTEM_COUNTER_TYPE_ID == typeId &&
+                SystemCounterDescriptor.SHORT_SENDS.id() == keyBuffer.getInt(0))
+            {
+                shortSendsCounterId.set(counterId);
+            }
+        });
```
vyazelenko: `Short send` is a system counter and thus it has a fixed id, i.e. just use `SystemCounterDescriptor.SHORT_SENDS.id()` directly.

aeron-system-tests/src/test/java/io/aeron/SpySimulatedConnectionTest.java
```diff
@@ -276,6 +278,36 @@ else if (!isSubscriptionClosed)
...
+
+        final String channel = "aeron:udp?control=localhost:40456|control-mode=dynamic";
+
+        spy = client.addSubscription(spyForChannel(channel), STREAM_ID);
+        publication = client.addExclusivePublication(channel + "|ssc=true", STREAM_ID);
+
+        Tests.awaitConnected(spy);
+        Tests.awaitConnected(publication);
+
+        Thread.sleep(500);
```
vyazelenko, line 295: Why sleep? Why not exchange a single message instead? This would not be clock dependent.

Also, if you need to wait for something then wait for `Bytes sent` to be incremented in addition to observing the data by the spy.

aeron-driver/src/main/c/media/aeron_udp_destination_tracker.c
```diff
@@ -109,7 +109,7 @@ int aeron_udp_destination_tracker_send(
     int result = (int)iov_length;
     int to_be_removed = 0;
 
-    *bytes_sent = 0;
+    *bytes_sent = iov->iov_len;
```
vyazelenko, line 112: This might not work correctly, because `send_func` might reset the value back to zero in some cases. See [aeron_udp_channel_transport_send](https://github.com/aeron-io/aeron/blob/512fb2464b2de3ba03b71dd1734d29b457e67371/aeron-driver/src/main/c/media/aeron_udp_channel_transport.c#L630) and the functions it calls.

vyazelenko (changes requested):

aeron-system-tests/src/test/java/io/aeron/SpySimulatedConnectionTest.java
nbradac: I need `aeron_network_publication_setup_message_check` to be called, so I need the sender thread to loop through at least once.  FWIW, if I remove the sleep, the test still seems to 'work' - i.e. without the fix to the C media driver, the subsequent assert fails every time, indicating we got at least one short send.  Seems like it's possible that by the time both the spy and the publication are 'connected', the sender thread must have already already run through a few times.  But I wasn't 100% sure about that, looking through the code.

aeron-driver/src/main/c/media/aeron_udp_destination_tracker.c
nbradac: I think that's ok... because I think that would indicate an actual short send.

vyazelenko: This will over-count `bytes sent` as `bytes_sent` is also incremented in the `send_func`.
Thinking about it a bit more why is the fix necessary at all given the `send_func` logic? Could it be that the bug is actually in the `send_func` implementations?

aeron-system-tests/src/test/java/io/aeron/SpySimulatedConnectionTest.java
```diff
@@ -276,6 +277,23 @@ else if (!isSubscriptionClosed)
...
+        launch();
+
+        final String channel = "aeron:udp?control=localhost:40456|control-mode=dynamic";
+
+        spy = client.addSubscription(spyForChannel(channel), STREAM_ID);
+        publication = client.addExclusivePublication(channel + "|ssc=true", STREAM_ID);
+
+        Tests.awaitConnected(spy);
+        Tests.awaitConnected(publication);
+
```
vyazelenko, line 293: Should call `waitUntilFullConnectivity()` which will also publish some data. Otherwise, how we are to check for short sends if nothing was sent yet?

vyazelenko (changes requested):

PR #1773, opened 2025-03-27
[Java] add low level send/receive short send system counters

nbradac:


aeron-driver/src/main/java/io/aeron/driver/media/SendChannelEndpoint.java
```diff
@@ -734,13 +755,20 @@ else if (datagramChannel.isOpen())
...
+
+                if (bytesSent < bytesToSend)
+                {
+                    transportShortSends.incrementRelease();
+                }
             }
         }
         catch (final PortUnreachableException ignore)
         {
+            transportShortSends.incrementRelease();
```
mikeb01, line 767: I'm wondering if the transport short sends should exclude these and have each as separate counters.  I think the transport short sends should be just the ones seen on line 761, which should be analogous to EAGAIN/EWOULDBLOCK.  I think this should be a separate counter which would be analogous to ECONNREFUSED on the C side.

aeron-driver/src/main/java/io/aeron/driver/media/SendChannelEndpoint.java
```diff
@@ -734,13 +755,20 @@ else if (datagramChannel.isOpen())
...
         }
         catch (final IOException ex)
         {
+            transportShortSends.incrementRelease();
```
mikeb01, line 771: I'm not sure if we should track this one in the same counter at all.  I think errors are not short sends in this case.  However, we'd need to double check the exact errors we get here.  I think these end up being configuration type errors.

vyazelenko:
We will address this with endpoint/destination refactoring.

PR #1774, opened 2025-04-02
synchronize session ids across cluster nodes

ryanoftalltrees:


ryanoftalltrees:
I left the option to set the session id outside of the constructor, but if the approach of reusing rejected session ids works, then we may not need it.  I also left the option to create a KeepAlive object that requires a client and one that doesn't.

aeron-cluster/src/main/java/io/aeron/cluster/ClusterSession.java
```diff
@@ -73,6 +73,11 @@ enum Action
         state(State.INIT);
     }
 
+    ClusterSession(final int responseStreamId, final String responseChannel)
+    {
+        this(Aeron.NULL_VALUE, responseStreamId, responseChannel);
+    }
+
```
vyazelenko, line 80: Why new constructor? Who uses it?

aeron-cluster/src/main/java/io/aeron/cluster/ClusterSession.java
```diff
@@ -88,6 +93,14 @@ public void close(final Aeron aeron, final ErrorHandler errorHandler)
...
     }
 
+    public void id(final long sessionId)
+    {
+        if (Aeron.NULL_VALUE == id)
+        {
+            id = sessionId;
+        }
+    }
+
```
vyazelenko, line 103: This should simply assign value, i.e. no check is needed here.

aeron-cluster/src/main/java/io/aeron/cluster/ClusterSession.java
```diff
@@ -290,11 +303,14 @@ boolean appendSessionToLogAndSendOpen(
     {
         if (responsePublication.availableWindow() > 0)
         {
+            // need session id here...
```
vyazelenko, line 306: leftover comment

aeron-cluster/src/main/java/io/aeron/cluster/ClusterSession.java
```diff
@@ -290,11 +303,14 @@ boolean appendSessionToLogAndSendOpen(
...
             final long resultingPosition = logPublisher.appendSessionOpen(this, leadershipTermId, clusterTimestamp);
             if (resultingPosition > 0)
             {
                 open(resultingPosition);
                 timeOfLastActivityNs(nowNs);
+
+                // might increment counter here...
```
vyazelenko, line 313: leftover comment

aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModuleAgent.java
```diff
@@ -42,11 +42,7 @@
 import org.agrona.concurrent.*;
 import org.agrona.concurrent.status.CountersReader;
 
-import java.util.ArrayDeque;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.List;
-import java.util.Objects;
+import java.util.*;
```
vyazelenko, line 45: Keep non-star imports please.

aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModuleAgent.java
```diff
@@ -657,9 +653,18 @@ public void onSessionConnect(
...
+        // use null value always here.  create session id later
+        // after authentication
+        //final long clusterSessionId = Cluster.Role.LEADER == role ? nextSessionId++ : NULL_VALUE;
+        /**
+         * let's try a new strategy here.  set the session id here and increment the counter
+         * if the session is rejected, make this session id available again.  is it possible
+         * for several sessions to be in process of authentication at a time?  if so, then this
+         * gets more complicated.
+         */
+        final long sessionId = getNewSessionId();
```
vyazelenko, line 665: This approach is much more complicated and error prone than the original one with not assigning an id until session is authenticated. Why was this approaches chosen?

aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModuleAgent.java
```diff
@@ -691,6 +696,29 @@ else if (pendingUserSessions.size() + sessions.size() >= ctx.maxConcurrentSessio
...
+            final LongHashSet.LongIterator it = rejectedSesssionIds.iterator(); it.hasNext();
+            while (it.hasNext())
+            {
+                final long value = it.nextValue();
+                if (value < min)
+                {
+                    min = value;
+                }
+            }
+            rejectedSesssionIds.remove(min);
```
vyazelenko, line 717: This is very inefficient.

aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModuleAgent.java
```diff
@@ -2604,8 +2629,15 @@ private int processPendingSessions(
                 {
                     case CLIENT:
                     {
+                        // next line tells everyone this session is open
+                        // TODO check if any session ID dependencies b/t original onSessionConnect() and here
```
vyazelenko, line 2633: leftover comments

aeron-system-tests/src/test/java/io/aeron/cluster/ClusterTest.java
```diff
@@ -2675,6 +2681,112 @@ public void onNewLeadershipTermEvent(
...
+            recordingId = snapshot.recordingId;
+        }
+
+        final AeronArchive.Context archiveCtx = new AeronArchive.Context()
+            .controlRequestChannel(node.archive().context().localControlChannel())
+            .controlResponseChannel(node.archive().context().localControlChannel())
+            .controlRequestStreamId(node.archive().context().localControlStreamId())
+            .aeronDirectoryName(node.mediaDriver().aeronDirectoryName());
+
+        try (
```
vyazelenko, line 2751: This method should use [ConsensusModuleSnapshotAdapter](https://github.com/aeron-io/aeron/blob/master/aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModuleSnapshotAdapter.java) with a `ConsensusModuleSnapshotListener` configured for proper asserting the state of the snapshot.

aeron-test-support/src/main/java/io/aeron/test/cluster/TestCluster.java
```diff
@@ -2324,12 +2346,53 @@ public byte[] onChallenge(final byte[] encodedChallenge)
...
+    {
+        private long keepAliveDeadlineMs;
+
+        public void init()
+        {
+            if (client == null)
+            {
+                return; // no op for non-existent client
+            }
+            final EpochClock epochClock = client.context().aeron().context().epochClock();
```
vyazelenko, line 2366: Why was this modified?

vyazelenko (changes requested):

aeron-cluster/src/main/java/io/aeron/cluster/ClusterSession.java
ryanoftalltrees: I left this here in case we need to revert back to it but yes, agreed it should go if we always set the id out of the gate.

vyazelenko:
That new test is failing in CI.
https://github.com/aeron-io/aeron/actions/runs/14226204604

aeron-system-tests/src/test/java/io/aeron/cluster/ClusterTest.java
ryanoftalltrees: I still need to do this.  It wasn't straightforward.  will tackle that next

aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModuleAgent.java
```diff
@@ -2604,8 +2607,13 @@ private int processPendingSessions(
                 {
                     case CLIENT:
                     {
+                        if (Cluster.Role.LEADER == role)
+                        {
+                            nextCommittedSessionId = Math.max(session.id() + 1, nextCommittedSessionId);
+                        }
```
vyazelenko, line 2613: `processPendingSessions` method is only called when `role == LEADER` so no need to check for the role.
However, there is another issue, i.e. update of `nextCommittedSessionId` before session is appended to the log. Instead the logic should be moved inside of the if statement body for `session.appendSessionToLogAndSendOpen` and it should be simplified to:
```java
nextCommittedSessionId = session.id() + 1;
```

aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModuleAgent.java
```diff
@@ -3399,15 +3407,17 @@ private void snapshotState(
 
         snapshotTaker.markBegin(SNAPSHOT_TYPE_ID, logPosition, leadershipTermId, 0, clusterTimeUnit, ctx.appVersion());
 
+        final long actualNextSessionId = role == Cluster.Role.LEADER ? nextCommittedSessionId : nextSessionId;
         if (pendingServiceMessageTrackers.length > 0)
         {
             final PendingServiceMessageTracker trackerOne = pendingServiceMessageTrackers[0];
             snapshotTaker.snapshotConsensusModuleState(
-                nextSessionId, trackerOne.nextServiceSessionId(), trackerOne.logServiceSessionId(), trackerOne.size());
+                actualNextSessionId, trackerOne.nextServiceSessionId(),
```
vyazelenko, line 3415: Simply use `nextCommittedSessionId` which is guaranteed to be correct for any node role.

aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModuleAgent.java
```diff
@@ -3399,15 +3407,17 @@ private void snapshotState(
...
+                trackerOne.logServiceSessionId(), trackerOne.size());
         }
         else
         {
-            snapshotTaker.snapshotConsensusModuleState(nextSessionId, 0, 0, 0);
+            snapshotTaker.snapshotConsensusModuleState(actualNextSessionId, 0, 0, 0);
```
vyazelenko, line 3420: Simply use `nextCommittedSessionId` which is guaranteed to be correct for any node role.

aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModuleAgent.java
```diff
@@ -3399,15 +3407,17 @@ private void snapshotState(
 
         snapshotTaker.markBegin(SNAPSHOT_TYPE_ID, logPosition, leadershipTermId, 0, clusterTimeUnit, ctx.appVersion());
 
+        final long actualNextSessionId = role == Cluster.Role.LEADER ? nextCommittedSessionId : nextSessionId;
```
vyazelenko, line 3410: No need for this (see below).

aeron-test-support/src/main/java/io/aeron/test/cluster/TestCluster.java
```diff
@@ -1368,7 +1370,7 @@ public void awaitServiceMessagePredicate(final TestNode node, final IntPredicate
         awaitServiceMessagePredicate(node, service, countPredicate);
     }
 
-    private final KeepAlive clientKeepAlive = new KeepAlive();
+    private final KeepAlive clientKeepAlive;
```
vyazelenko, line 1373: Why this change?

vyazelenko (changes requested):

aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModuleAgent.java
```diff
@@ -2604,8 +2607,10 @@ private int processPendingSessions(
                 {
                     case CLIENT:
                     {
+                        nextCommittedSessionId = session.id() + 1;
```
vyazelenko, line 2610: This line needs to be moved inside of the if statement that follows it, i.e. after a session was appended to the log.
The `nextCommittedSessionId` needs to be protected against going backwards, i.e. pending sessions are processed from newest to oldest so the subsequent session will decrease the value.
Instead we should also check that the value is larger, i.e. add `if (session.id() >= nextCommittedSessionId)`.

vyazelenko (changes requested):

aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModuleAgent.java
```diff
@@ -2605,8 +2608,10 @@ private int processPendingSessions(
                     case CLIENT:
                     {
                         if (session.appendSessionToLogAndSendOpen(
-                            logPublisher, egressPublisher, leadershipTermId, memberId, nowNs, clusterClock.time()))
+                            logPublisher, egressPublisher, leadershipTermId,
+                            memberId, nowNs, clusterClock.time()))
                         {
+                            nextCommittedSessionId = session.id() + 1;
```
vyazelenko, line 2614: Protection against regression is missing, i.e. we must prevent `nextCommittedSessionId` from going backwards. Therefore a check is required to enforce `session.id() >= nextCommittedSessionId`.

vyazelenko (changes requested):

aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModuleAgent.java
ryanoftalltrees: actually, going with this:  nextCommittedSessionId = Math.max(session.id(),nextCommittedSessionId)+1;

aeron-system-tests/src/test/java/io/aeron/cluster/ClusterTest.java
```diff
@@ -2675,6 +2685,200 @@ public void onNewLeadershipTermEvent(
...
+
+        // wait for cluster to hold election
+        final TestNode leader = cluster.awaitLeader();
+
+        // the first client is successful
+        cluster.connectClient();
+
+        // the second client fails
+        assertThrowsExactly(AuthenticationException.class, () -> cluster.connectClient());
+        assertThrowsExactly(AuthenticationException.class, () -> cluster.connectClient());
```
vyazelenko, line 2743: Instead of calling `TestClutser#connectClient` we should create a standalone `AeronCluster` instance. Something like this:
```java
// Original client - should always be allowed
final AeronCluster client = connectClient();

final AeronCluster.Context clientContext = new AeronCluster.Context()
    .aeronDirectoryName(client.context().aeronDirectoryName())
    .aeron(client.context().aeron())
    .ingressChannel(client.context().ingressChannel())
    .ingressStreamId(555)
    .isIngressExclusive(true)
    .egressChannel(client.context().egressChannel())
    .egressStreamId(777)
    .ingressEndpoints(client.context().ingressEndpoints());

// Another session -> also OK
final AeronCluster client2 = AeronCluster.connect(clientContext.clone());

// Any further connections are rejected
assertThrowsExactly(AuthenticationException.class, () -> AeronCluster.connect(clientContext.clone()));
```

vyazelenko (changes requested):

aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModuleAgent.java
```diff
@@ -2605,8 +2608,10 @@ private int processPendingSessions(
                     case CLIENT:
                     {
                         if (session.appendSessionToLogAndSendOpen(
-                            logPublisher, egressPublisher, leadershipTermId, memberId, nowNs, clusterClock.time()))
+                            logPublisher, egressPublisher, leadershipTermId,
+                            memberId, nowNs, clusterClock.time()))
                         {
+                            nextCommittedSessionId = Math.max(session.id(), nextCommittedSessionId) + 1;
```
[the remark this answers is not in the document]
ryanoftalltrees: done

aeron-system-tests/src/test/java/io/aeron/cluster/ClusterTest.java
ryanoftalltrees: done

PR #1775, opened 2025-04-04
[Java] New term buffer cleanup logic

vyazelenko:
* Cleanup term buffers in fixed 4KB chunks.
* Prevent `pub-lmt` advancing too much ahead to prevent it intersecting with a dirty term buffer:
  * `pub-lmt` can never end up in a dirty term.
  * `pub-lmt` is not allowed to be in a term adjacent to a dirty term if the clean position points to the start of a dirty term.
* Do not skip a term when cleaning a `PublicationImage`.
* Only update `pub-lmt` on every trip gain, i.e. every 1/8th of a term or one full publication window (whichever happens first).

### DIFF
--- a/aeron-driver/src/main/java/io/aeron/driver/NetworkPublication.java
+++ b/aeron-driver/src/main/java/io/aeron/driver/NetworkPublication.java
@@ -38,6 +38,7 @@ import org.agrona.concurrent.status.AtomicCounter;
 import org.agrona.concurrent.status.Position;
 import org.agrona.concurrent.status.ReadablePosition;
 
+import java.lang.invoke.VarHandle;
 import java.net.InetSocketAddress;
 import java.nio.ByteBuffer;
 import java.util.ArrayList;
@@ -50,7 +51,6 @@ import static io.aeron.logbuffer.TermScanner.*;
 import static io.aeron.protocol.DataHeaderFlyweight.BEGIN_AND_END_FLAGS;
 import static io.aeron.protocol.DataHeaderFlyweight.BEGIN_END_AND_EOS_FLAGS;
 import static io.aeron.protocol.StatusMessageFlyweight.END_OF_STREAM_FLAG;
-import static org.agrona.BitUtil.SIZE_OF_LONG;
 
 class NetworkPublicationPadding1
 {
@@ -618,7 +618,7 @@ public final class NetworkPublication
     {
         final long senderPosition = this.senderPosition.get();
         final int activeTermId = computeTermIdFromPosition(senderPosition, positionBitsToShift, initialTermId);
-        final int termOffset = (int)senderPosition & termLengthMask;
+        final int termOffset = (int)(senderPosition & termLengthMask);
 
         if (!hasInitialConnection || isSetupElicited)
         {
@@ -905,12 +905,11 @@ public final class NetworkPublication
         if (position > cleanPosition)
         {
             final UnsafeBuffer dirtyTermBuffer = termBuffers[indexByPosition(cleanPosition, positionBitsToShift)];
-            final int bytesForCleaning = (int)(position - cleanPosition);
-            final int termOffset = (int)cleanPosition & termLengthMask;
-            final int length = Math.min(bytesForCleaning, termBufferLength - termOffset);
+            final int termOffset = (int)(cleanPosition & termLengthMask);
+            final int length = Math.min((int)(position - cleanPosition), termBufferLength - termOffset);
 
-            dirtyTermBuffer.setMemory(termOffset + SIZE_OF_LONG, length - SIZE_OF_LONG, (byte)0);
-            dirtyTermBuffer.putLongRelease(termOffset, 0);
+            dirtyTermBuffer.setMemory(termOffset, length, (byte)0);
+            VarHandle.storeStoreFence();
             this.cleanPosition = cleanPosition + length;
         }
     }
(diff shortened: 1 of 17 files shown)


aeron-driver/src/main/c/aeron_driver_context.c
```diff
@@ -1420,6 +1420,45 @@ int aeron_driver_context_validate_mtu_length(uint64_t mtu_length)
...
 }
 
+int aeron_driver_context_validate_receiver_window_length(const char* param, size_t window_length, size_t mtu_length)
+{
+    if (window_length < (mtu_length << 1))
+    {
+        AERON_SET_ERR(
+            EINVAL,
+            "%s=%" PRIu64 " must be at least two times larger than the %s=%" PRIu64,
+            param,
```
mikeb01, line 1430: You should cast the `size_t` to a `uint64_t` if you are using the `PRIu64` format options.  The other option is `%z` which is explicitly for `size_t`, but we have generally avoided that.

aeron-driver/src/main/c/aeron_driver_context.c
```diff
@@ -1420,6 +1420,45 @@ int aeron_driver_context_validate_mtu_length(uint64_t mtu_length)
...
+    {
+        AERON_APPEND_ERR("%s", "");
+        return -1;
+    }
+
+    if (window_length > (term_length >> 1))
+    {
+        AERON_SET_ERR(
+            EINVAL,
+            "%s=%" PRIu64 " must not exceed half the %s=%" PRIu64,
```
mikeb01, line 1452: Cast to `uint64_t`.

aeron-driver/src/main/c/aeron_term_cleaner.h
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+#ifndef AERON_TERM_CLEANER_H
+#define AERON_TERM_CLEANER_H
+
+#include <stddef.h>
+#include <stdint.h>
+#include "util/aeron_bitutil.h"
+
+#define AERON_TERM_CLEANER_TERM_CLEANUP_BLOCK_LENGTH (4096u)
```
mikeb01, line 24: Use:
```
#define AERON_TERM_CLEANER_TERM_CLEANUP_BLOCK_LENGTH UINT32_C(4096)
```
To declare this as unsigned and use the appropriate system dependent suffix.

aeron-driver/src/main/c/aeron_ipc_publication.c
```diff
@@ -286,50 +288,58 @@ int aeron_ipc_publication_update_pub_pos_and_lmt(aeron_ipc_publication_t *public
...
+            int64_t proposed_limit = min_sub_pos + publication->term_window_length;
+            if (proposed_limit >= publication->conductor_fields.trip_limit)
+            {
+                const int64_t max_allowed_gap = (int64_t)publication->mapped_raw_log.term_length << 1;
+                const int64_t term_length_mask = (int64_t)publication->mapped_raw_log.term_length - 1;
+                const int64_t new_term_based_limit_position = proposed_limit - (proposed_limit & term_length_mask);
+                const int64_t clean_position = publication->conductor_fields.clean_position;
+                const int64_t wrap_around_gap = new_term_based_limit_position - clean_position;
+                if (wrap_around_gap <= max_allowed_gap &&
+                    (wrap_around_gap < max_allowed_gap || 0 != (clean_position & term_length_mask)))
```
mikeb01, line 307: I'm a little unsure of this logic.  By my reading, if `wrap_around_gap < max_allowed_gap` is true then the whole evaluation of this logic should be true, because if `x < y` then `x <= y`.  Therefore we could rewrite this as:
```
if (wrap_around_gap < max_allowed_gap ||
    (wrap_around_gap == max_allowed_gap && 0 != (clean_position & term_length_mask)))
```
However, for `wrap_around_gap == max_allowed_gap` then `clean_position` must be `new_term_based_limit_position - term_length / 2` in which case `0 != (clean_position & term_length_mask)` will always be true.  This suggests that the logic only needs to be `wrap_around_gap <= max_allowed_gap`.

I'm not sure if that is what you are going for here or if there is some additional case that you are trying to capture.

aeron-driver/src/main/c/aeron_driver_context.c
vyazelenko: Done.

aeron-driver/src/main/c/aeron_ipc_publication.c
vyazelenko: I've re-written this logic. It now compares both values at term-boundary and handles offset separately.

aeron-driver/src/main/java/io/aeron/driver/NetworkPublication.java
```diff
@@ -741,13 +742,21 @@ int updatePublisherPositionAndLimit()
                     minConsumerPosition = Math.min(minConsumerPosition, spyPosition.getVolatile());
                 }
 
-                final long proposedPublisherLimit = minConsumerPosition + termWindowLength;
-                final long publisherLimit = this.publisherLimit.get();
-                if (proposedPublisherLimit > publisherLimit)
+                workCount += cleanBufferTo(minConsumerPosition - termBufferLength);
+                final long newLimitPosition = minConsumerPosition + termWindowLength;
```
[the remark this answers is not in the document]
vyazelenko: Done.

PR #1781, opened 2025-04-17
Publication revoke

nbradac:


aeron-client/src/main/java/io/aeron/protocol/DataHeaderFlyweight.java
```diff
@@ -156,7 +166,18 @@ public static boolean isHeartbeat(final UnsafeBuffer packet, final int length)
      */
     public static boolean isEndOfStream(final UnsafeBuffer packet)
     {
-        return BEGIN_END_AND_EOS_FLAGS == (packet.getByte(FLAGS_FIELD_OFFSET) & 0xFF);
+        return (packet.getByte(FLAGS_FIELD_OFFSET) & EOS_FLAG) != 0;
```
nbradac, line 169: It's worth pointing out that this change means that old java drivers won't detect EOS or REVOKED when they see a REVOKE/EOS from a newer driver.

aeron-driver/src/main/c/aeron_ipc_publication.c
```diff
@@ -449,6 +455,32 @@ void aeron_ipc_publication_on_time_event(
...
+                        AERON_IPC_CHANNEL,
+                        AERON_IPC_CHANNEL_LEN);
+                }
+            }
+
+            publication->conductor_fields.state = AERON_IPC_PUBLICATION_STATE_DRAINING;
+            break;
+        }
+
         case AERON_IPC_PUBLICATION_STATE_DRAINING:
```
nbradac, line 484: In the revoked case, I'm not sure I'm skipping through here quick enough - i.e. for IPC we're still going through the DRAINING state.

aeron-driver/src/main/c/aeron_network_publication.c
```diff
@@ -1224,6 +1273,16 @@ void aeron_network_publication_on_time_event(
             break;
         }
 
+        case AERON_NETWORK_PUBLICATION_STATE_REVOKED:
+        {
+            aeron_driver_conductor_publication_transition_to_revoked(conductor, &publication->conductor_fields.managed_resource);
+            aeron_driver_conductor_cleanup_spies(conductor, publication);
+
+            publication->conductor_fields.time_of_last_activity_ns = now_ns;
+            publication->conductor_fields.state = AERON_NETWORK_PUBLICATION_STATE_LINGER;
```
nbradac, line 1282: I don't like that the state transitions are different for IPC vs. Network publications.

aeron-driver/src/main/c/aeron_driver_common.h
```diff
@@ -35,6 +35,9 @@ typedef struct aeron_driver_managed_resource_stct
     void *clientd;
     void (*decref)(void *);
     void (*incref)(void *);
+    // TODO rework this - maybe add a 'managed resource type' enum and use that to determine what function to call when revoking publication?
```
nbradac, line 38: https://github.com/nbradac/aeron/pull/1

aeron-client/src/main/java/io/aeron/protocol/DataHeaderFlyweight.java
nbradac: This change is being pushed to the 1.47.x branch for backwards compatibility issues, should they arise.  Otherwise we're just going to document the fact that revoke() shouldn't be used against < 1.48 _Java_ drivers.

aeron-client/src/main/c/aeron_exclusive_publication.h
```diff
@@ -52,6 +52,7 @@ typedef struct aeron_exclusive_publication_stct
     void *on_close_complete_clientd;
 
     volatile bool is_closed;
+    volatile bool revoke_on_close;
```
nbradac, line 55: Does this need to be volatile?

aeron-driver/src/main/c/aeron_driver_conductor.c
```diff
@@ -3325,16 +3318,35 @@ aeron_rb_read_action_t aeron_driver_conductor_on_command(
...
+            aeron_remove_publication_command_t *command = (aeron_remove_publication_command_t *)message;
 
-            if (length < sizeof(aeron_remove_command_t))
+            aeron_remove_publication_command_t complete_command;
+            aeron_remove_publication_command_t *complete_command_p;
+
+            if (length < sizeof(aeron_remove_publication_command_t))
             {
-                goto malformed_command;
+                if (length == 24) // old aeron_remove_command_t has no flags field
```
mikeb01, line 3328: This shouldn't be a magic number, consider using a constant or `offsetof(aeron_remove_publication_stct, flags)`.

aeron-system-tests/src/test/java/io/aeron/PublicationRevokeTest.java
```diff
@@ -0,0 +1,329 @@
...
+import static io.aeron.driver.status.SystemCounterDescriptor.PUBLICATIONS_REVOKED;
+import static io.aeron.driver.status.SystemCounterDescriptor.PUBLICATION_IMAGES_REVOKED;
+import static java.util.concurrent.TimeUnit.MILLISECONDS;
+import static java.util.concurrent.TimeUnit.SECONDS;
+import static org.agrona.BitUtil.SIZE_OF_INT;
+import static org.junit.jupiter.api.Assertions.*;
+import static org.mockito.Mockito.*;
+
+@ExtendWith(InterruptingTestCallback.class)
+class PublicationRevokeTest
```
mikeb01, line 47: I'd like to see a test with a larger number of messages where we revoke part way through the subscription reading those messages.

aeron-client/src/main/c/aeron_exclusive_publication.h
mikeb01: I don't think that it does.  It looks like this field should only be modified by the client thread.

aeron-client/src/test/cpp_wrapper/PublicationRevokeTest.cpp
```diff
@@ -0,0 +1,288 @@
...
+    POLL_FOR(1 == imageUnavailable, invoker);
+
+    ASSERT_EQ(0, sub->imageCount());
+
+    ASSERT_EQ(1, aeron->countersReader().getCounterValue(AERON_SYSTEM_COUNTER_PUBLICATIONS_REVOKED));
+    ASSERT_EQ(expectedPublicationImagesRevoked, aeron->countersReader().getCounterValue(AERON_SYSTEM_COUNTER_PUBLICATION_IMAGES_REVOKED));
+}
+
+/*
+TEST_P(PublicationRevokeTest, revokeTestConcurrent)
```
mikeb01, line 134: Delete commented code.

aeron-driver/src/main/c/aeron_network_publication.c
```diff
@@ -492,34 +496,47 @@ int aeron_network_publication_heartbeat_message_check(
...
 
     if (publication->has_initial_connection &&
         now_ns > (publication->time_of_last_data_or_heartbeat_ns + AERON_NETWORK_PUBLICATION_HEARTBEAT_TIMEOUT_NS))
     {
+        uint8_t flags;
+
+        uint8_t is_publication_revoked;
+        AERON_GET_ACQUIRE(is_publication_revoked, publication->log_meta_data->is_publication_revoked);
+
+        if (is_publication_revoked)
```
mikeb01, line 512: Potentially use this simplification:
```
        uint8_t flags = AERON_DATA_HEADER_BEGIN_FLAG | AERON_DATA_HEADER_END_FLAG;
        if (is_publication_revoked)
        {
            flags |= AERON_DATA_HEADER_REVOKED_FLAG;
            flags |= AERON_DATA_HEADER_EOS_FLAG;
        }
        else if (publication->signal_eos & is_end_of_stream)
        {
            flags |= AERON_DATA_HEADER_EOS_FLAG;
        }
```

aeron-driver/src/main/c/aeron_network_publication.c
```diff
@@ -1271,8 +1342,9 @@ void aeron_network_publication_on_time_event(
             bool has_received_unicast_eos = false;
             AERON_GET_ACQUIRE(has_received_unicast_eos, publication->has_received_unicast_eos);
 
-            if (has_received_unicast_eos ||
-                now_ns > (publication->conductor_fields.time_of_last_activity_ns + publication->linger_timeout_ns))
+            if (publication->conductor_fields.refcnt == 0 &&
```
mikeb01, line 1345: As a belt and braces I think we should use:
```
publication->conductor_fields.refcnt <= 0
```

aeron-driver/src/main/c/aeron_ipc_publication.c
```diff
@@ -488,7 +542,10 @@ void aeron_ipc_publication_on_time_event(
 
         case AERON_IPC_PUBLICATION_STATE_LINGER:
         {
-            publication->conductor_fields.has_reached_end_of_life = true;
+            if (publication->conductor_fields.refcnt == 0)
```
mikeb01, line 545: As a belt and braces I think we should use:
```
publication->conductor_fields.refcnt <= 0
```

mikeb01:
This is looking good.  However, I would like to see some tests for multicast and MDC with multiple subscriptions to ensure that the events are propagated correctly out to all clients.  It would be good to run these tests with multiple drivers for the subscription clients to poke that behaviour.

PR #1785, opened 2025-04-25
Image.reject()

nbradac:


aeron-driver/src/main/c/aeron_ipc_publication.c
```diff
@@ -257,6 +264,71 @@ bool aeron_ipc_publication_free(aeron_ipc_publication_t *publication)
...
+    {
+        uint8_t buffer[AERON_COMMAND_PUBLICATION_ERROR_MAX_LENGTH];
+        aeron_command_publication_error_t *error = (aeron_command_publication_error_t *)buffer;
+        int32_t error_length = reason_length <= AERON_ERROR_MAX_TEXT_LENGTH ? reason_length : AERON_ERROR_MAX_TEXT_LENGTH;
+
+        struct sockaddr_storage src_addr_storage;
+        memset(&src_addr_storage, 0, sizeof(struct sockaddr_storage));
+
+        struct sockaddr_in *src_addr_in = (struct sockaddr_in *)&src_addr_storage;
+        src_addr_in->sin_family = AF_INET;
```
mikeb01, line 284: I'm wondering if we should use `AF_LOCAL` or `AF_UNSPEC` here because this is IPC rather than UDP?

nbradac: Possibly.  I'm not sure how that would work in Java... and the Java publication error frame callback expects an InetSocketAddress.

FWIW, using AF_LOCAL causes the C driver to encode the address type as just 0 in the aeron_publication_error_t, and then the Java client throws an exception because it expects either an IPV4 or IPV6 value.

mikeb01: Probably best just to leave it as is and set it all to zero.

mikeb01: Specifically, I'm wondering if we should set the address:
```
src_addr_in->sin_addr.s_addr = 0;
```
Rather than using the loopback address.  But this is fairly minor.

nbradac: I'm not sure how to do that in Java.

aeron-driver/src/main/java/io/aeron/driver/IpcPublication.java
```diff
@@ -119,7 +130,11 @@ enum State
         this.unblockTimeoutNs = ctx.publicationUnblockTimeoutNs();
         this.untetheredWindowLimitTimeoutNs = params.untetheredWindowLimitTimeoutNs;
         this.untetheredRestingTimeoutNs = params.untetheredRestingTimeoutNs;
-        this.unblockedPublications = ctx.systemCounters().get(UNBLOCKED_PUBLICATIONS);
+        this.imageLivenessTimeoutNs = ctx.imageLivenessTimeoutNs();
+        final SystemCounters systemCounters = ctx.systemCounters();
+        this.unblockedPublications = systemCounters.get(UNBLOCKED_PUBLICATIONS);
+        this.errorMessagesReceived = systemCounters.get(ERROR_FRAMES_RECEIVED);
```
mikeb01, line 136: I think we should have a separate counter for IPC rejections.  There are no actual error frames involved here, but we would still like some observability.

aeron-driver/src/main/c/aeron_ipc_publication.c
```diff
@@ -203,14 +203,21 @@ int aeron_ipc_publication_create(
...
+    _pub->in_cool_down = false;
+    _pub->cool_down_expire_time_ns = 0;
 
     _pub->conductor_fields.consumer_position = aeron_ipc_publication_producer_position(_pub);
     _pub->conductor_fields.last_consumer_position = _pub->conductor_fields.consumer_position;
     _pub->conductor_fields.clean_position = _pub->conductor_fields.consumer_position;
 
     _pub->unblocked_publications_counter = aeron_system_counter_addr(
         system_counters, AERON_SYSTEM_COUNTER_UNBLOCKED_PUBLICATIONS);
+    _pub->error_messages_received_counter = aeron_system_counter_addr(
```
mikeb01, line 217: I think we should have a separate counter for IPC rejections.  There are no actual error frames involved here, but we would still like some observability.

mikeb01:
I'm happy with this as a whole.  I think if we could make the counter changes, then I would be happy for it to be merged.

aeron-driver/src/main/java/io/aeron/driver/status/SystemCounterDescriptor.java
```diff
@@ -231,7 +231,12 @@ public enum SystemCounterDescriptor
     /**
      * A count of the number of error frames sent by this driver.
      */
-    ERROR_FRAMES_SENT(39, "Error Frames sent");
+    ERROR_FRAMES_SENT(39, "Error Frames sent"),
+
+    /**
+     * A count of the number of IPC rejections encountered by this driver.
+     */
+    IPC_REJECTIONS(40, "IPC rejections");
```
vyazelenko, line 239: I think this should be a more generic counter, i.e. `Images rejected` that should be incremented in both UDP and IPC cases.

aeron-driver/src/main/c/aeron_system_counters.c
```diff
@@ -62,7 +62,8 @@ static aeron_system_counter_t system_counters[] =
         { "Retransmitted bytes", AERON_SYSTEM_COUNTER_RETRANSMITTED_BYTES },
         { "Retransmit Pool Overflow count", AERON_SYSTEM_COUNTER_RETRANSMIT_OVERFLOW },
         { "Error Frames received", AERON_SYSTEM_COUNTER_ERROR_FRAMES_RECEIVED },
-        { "Error Frames sent", AERON_SYSTEM_COUNTER_ERROR_FRAMES_SENT }
+        { "Error Frames sent", AERON_SYSTEM_COUNTER_ERROR_FRAMES_SENT },
+        { "IPC rejections", AERON_SYSTEM_COUNTER_IPC_REJECTIONS }
```
vyazelenko, line 66: See comment in Java impl.

vyazelenko (changes requested):

aeron-driver/src/main/java/io/aeron/driver/status/SystemCounterDescriptor.java
nbradac: I like that.  I'll make that change.

vyazelenko:
@nbradac The new counter should be also incremented when rejecting a UDP image, i.e. either inside of [DriverConductor#onRejectImage](https://github.com/aeron-io/aeron/blob/78eb344ece12db8f7fb2c141e329b47c93651290/aeron-driver/src/main/java/io/aeron/driver/DriverConductor.java#L1503) or [PublicationImage#reject](https://github.com/aeron-io/aeron/blob/78eb344ece12db8f7fb2c141e329b47c93651290/aeron-driver/src/main/java/io/aeron/driver/PublicationImage.java#L943).

PR #1787, opened 2025-04-29
Provided inline versions for concurrency primitives.

pveentjer:


vyazelenko (changes requested):
@pveentjer We have the following macro definition that need to be replaced with inline functions:
* `AERON_GET_AND_ADD_INT32` - one to one translation.
* `AERON_GET_AND_ADD_INT64` - one to one translation.
* `AERON_GET_ACQUIRE` - requires multiple typed functions.
* `AERON_SET_RELEASE` - requires multiple typed functions.

The last two macro definitions are used on the following types:
* `bool` (e.g. [aeron_agent_is_running](https://github.com/aeron-io/aeron/blob/8a6f9a047148edfbd3b4c69d084b759dc86b39cd/aeron-client/src/main/c/aeron_agent.h#L108))
* `enum` (e.g. [aeron_client_registration_status_t](https://github.com/aeron-io/aeron/blob/8a6f9a047148edfbd3b4c69d084b759dc86b39cd/aeron-client/src/main/c/aeron_client.c#L354))
* `int32_t` (e.g. [aeron_counter_heartbeat_timestamp_find_counter_id_by_registration_id](https://github.com/aeron-io/aeron/blob/8a6f9a047148edfbd3b4c69d084b759dc86b39cd/aeron-client/src/main/c/aeron_client_conductor.h#L428))
* `int64_t` (e.g. [aeron_exclusive_publication_channel_status](https://github.com/aeron-io/aeron/blob/8a6f9a047148edfbd3b4c69d084b759dc86b39cd/aeron-client/src/main/c/aeron_exclusive_publication.c#L930))
* `struct` (e.g. [aeron_image_list_stct](https://github.com/aeron-io/aeron/blob/8a6f9a047148edfbd3b4c69d084b759dc86b39cd/aeron-client/src/main/c/aeron_subscription.c#L245))
* `uint64_t` (e.g. [aeron_mpsc_concurrent_array_queue_offer](https://github.com/aeron-io/aeron/blob/8a6f9a047148edfbd3b4c69d084b759dc86b39cd/aeron-client/src/main/c/concurrent/aeron_mpsc_concurrent_array_queue.h#L61))
* `void*` (e.g. [aeron_mpsc_concurrent_array_queue_drain](https://github.com/aeron-io/aeron/blob/8a6f9a047148edfbd3b4c69d084b759dc86b39cd/aeron-client/src/main/c/concurrent/aeron_mpsc_concurrent_array_queue.h#L100))
* `int` (e.g. [is_running](https://github.com/aeron-io/aeron/blob/8a6f9a047148edfbd3b4c69d084b759dc86b39cd/aeron-driver/src/main/c/aeronmd.c#L52))

Adding function definition is not enough. This also needs to replace all macro usage with those functions.

We must prove that switching to the inline functions will not regress performance compared to the macro version. This needs to be verified using multiple compilers (Clang, GCC, MSVC) and platforms. First, we should verify that the generated assembly for each function vs macro is the same using [Compiler Explorer](https://godbolt.org/). Then to do some perf testing (IPC and echo) on several platforms.

pveentjer:
I wanted to keep the initial work a small as possible so it will be easy to change. I'll add the different type permutations as you suggested above; thanks for the listing!

vyazelenko:
> I wanted to keep the initial work a small as possible.

We won't be able to verify absence of performance regressions without integrating everything, i.e. there should be a single PR that replace macro with inline functions.

PR #1788, opened 2025-04-29
Prevent message gaps in cluster sessions

wojciech-adaptive:
It's possible that after some transient network issue where an image goes unavailable and then available again, a leader will process ingress messages with a gap, and similarly it's possible that a client will process egress messages with a gap. This silent message loss could lead to issues in application protocols which do not expect gaps, so we want to prevent that. This should be very unlikely with the default settings though.

The commits:

- fix the issue on the cluster side for ingress,
- add an optimization to close a session when image goes unavailable without waiting for the session timeout,
- fix the issue on the client side for egress.

A breaking change if someone would run an old client and a new one on the same driver with the same channel, because different rejoin settings are conflicting.

### DIFF
--- a/aeron-cluster/src/main/java/io/aeron/cluster/ClusterSession.java
+++ b/aeron-cluster/src/main/java/io/aeron/cluster/ClusterSession.java
@@ -24,6 +24,7 @@ import io.aeron.cluster.codecs.EventCode;
 import io.aeron.exceptions.AeronException;
 import io.aeron.exceptions.RegistrationException;
 import io.aeron.logbuffer.BufferClaim;
+import io.aeron.logbuffer.Header;
 import org.agrona.*;
 import org.agrona.collections.ArrayUtil;
 import org.agrona.concurrent.errors.DistinctErrorLog;
@@ -55,6 +56,7 @@ final class ClusterSession implements ClusterClientSession
     private long openedLogPosition = AeronArchive.NULL_POSITION;
     private long closedLogPosition = AeronArchive.NULL_POSITION;
     private transient long timeOfLastActivityNs;
+    private transient long ingressImageCorrelationId = Aeron.NULL_VALUE;
     private long responsePublicationId = Aeron.NULL_VALUE;
     private final int responseStreamId;
     private final String responseChannel;
@@ -417,6 +419,24 @@ final class ClusterSession implements ClusterClientSession
         return requestInput;
     }
 
+    void linkIngressImage(final Header header)
+    {
+        if (Aeron.NULL_VALUE == ingressImageCorrelationId)
+        {
+            ingressImageCorrelationId = ((Image)header.context()).correlationId();
+        }
+    }
+
+    void unlinkIngressImage()
+    {
+        ingressImageCorrelationId = Aeron.NULL_VALUE;
+    }
+
+    long ingressImageCorrelationId()
+    {
+        return ingressImageCorrelationId;
+    }
+
     static void checkEncodedPrincipalLength(final byte[] encodedPrincipal)
     {
         if (null != encodedPrincipal && encodedPrincipal.length > MAX_ENCODED_PRINCIPAL_LENGTH)
@@ -435,6 +455,7 @@ final class ClusterSession implements ClusterClientSession
             ", openedLogPosition=" + openedLogPosition +
             ", closedLogPosition=" + closedLogPosition +
             ", timeOfLastActivityNs=" + timeOfLastActivityNs +
+            ", ingressImageCorrelationId=" + ingressImageCorrelationId +
             ", responseStreamId=" + responseStreamId +
             ", responseChannel='" + responseChannel + '\'' +
             ", responsePublicationId=" + responsePublicationId +
--- a/aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModuleAgent.java
+++ b/aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModuleAgent.java
@@ -3598,10 +3598,7 @@ final class ConsensusModuleAgent
             return; // don't subscribe to ingress if follower and multicast ingress
         }
 
-        if (!ingressUri.containsKey(REJOIN_PARAM_NAME))
-        {
-            ingressUri.put(REJOIN_PARAM_NAME, "false");
-        }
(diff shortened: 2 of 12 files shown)


aeron-cluster/src/main/java/io/aeron/cluster/ClusterSession.java
```diff
@@ -55,6 +56,7 @@ enum Action
     private long openedLogPosition = AeronArchive.NULL_POSITION;
     private long closedLogPosition = AeronArchive.NULL_POSITION;
     private transient long timeOfLastActivityNs;
+    private transient long ingressImageCorrelationId = Aeron.NULL_VALUE;
```
vyazelenko, line 59: Drop `transient`.

aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModuleAgent.java
```diff
@@ -3567,6 +3585,11 @@ private void connectIngress()
             return; // don't subscribe to ingress if follower and multicast ingress
         }
 
+        if (!ingressUri.containsKey(REJOIN_PARAM_NAME))
+        {
+            ingressUri.put(REJOIN_PARAM_NAME, "false");
+        }
+
```
vyazelenko, line 3592: Why is this conditional, i.e. why not always set `rejoin=false`?

aeron-cluster/src/main/java/io/aeron/cluster/client/AeronCluster.java
```diff
@@ -1341,6 +1342,13 @@ public void conclude()
             {
                 throw new ConfigurationException("egressChannel must be specified");
             }
+
+            final ChannelUri egressChannelUri = ChannelUri.parse(egressChannel);
+            if (egressChannelUri.isUdp() && !egressChannelUri.containsKey(REJOIN_PARAM_NAME))
+            {
+                egressChannelUri.put(REJOIN_PARAM_NAME, "false");
+                egressChannel = egressChannelUri.toString();
+            }
```
vyazelenko, line 1351: Why is this conditional, i.e. why not always set `rejoin=false`?

vyazelenko (changes requested):

aeron-cluster/src/main/java/io/aeron/cluster/client/AeronCluster.java
wojciech-adaptive: Primarily because it's breaking if someone deploys new version on the same MD with the same channel, so this gives an escape hatch. Also in case someone doesn't mind message gaps. But those are minor, so if you feel strongly about it, I don't mind making it unconditional.

aeron-cluster/src/main/java/io/aeron/cluster/ClusterSession.java
vyazelenko: Good point.

aeron-cluster/src/main/java/io/aeron/cluster/client/AeronCluster.java
vyazelenko: I would make both sides use `rejoin=false` unconditionally. Wonder what @mikeb01 thinks?

mikeb01: My inclination would be to make it unconditional, but I would like to know more about the breaking scenario that we could run into.

aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModuleAgent.java
```diff
@@ -1339,7 +1342,9 @@ long prepareForNewLeadership(final long logPosition, final long nowNs)
             clearSessionsAfter(logPosition);
             for (int i = 0, size = sessions.size(); i < size; i++)
             {
-                sessions.get(i).disconnect(aeron, ctx.countedErrorHandler());
+                final ClusterSession session = sessions.get(i);
+                session.unlinkIngressImage();
```
mikeb01, line 1346: Should `ClusterSession::unlinkIngressImage` be part of `ClusterSession::disconnect`?  Is there a reason that it needs to be separate?  Also, should we be unlinking in `ConsensusModuleAgent::onSessionClose`?

aeron-system-tests/src/test/java/io/aeron/cluster/ClusterSessionReliabilityTest.java
```diff
@@ -0,0 +1,477 @@
...
+            {
+                outOfSequenceCount++;
+                System.out.println("expected sequence " + nextSequence + ", but got " + sequence);
+            }
+            nextSequence = sequence + 1;
+        }
+    }
+
+    private static final class PortLossGenerator implements LossGenerator
+    {
```
mikeb01, line 405: Move to test-support?

mikeb01:
Overall looks good.

aeron-cluster/src/main/java/io/aeron/cluster/client/AeronCluster.java
wojciech-adaptive: If you have an old client running and start a new one on the same MD and with the same egress channel, then [this](https://github.com/aeron-io/aeron/blob/0c9eadceb1ac8e3121aa75361d8567d86054adb7/aeron-driver/src/main/java/io/aeron/driver/DriverConductor.java#L2207) will trigger and prevent the new one from connecting. It's very unlikely though, so happy to simplify and make unconditional, please let me know if I should proceed with that.

aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModuleAgent.java
mikeb01: It feels to me that it would be cleaner if it was unlinked on closed as well.  However, I think we should look at that as a separate change.  Not needed for this PR.

aeron-cluster/src/main/java/io/aeron/cluster/client/AeronCluster.java
mikeb01: Lets go with unconditional.

PR #1789, opened 2025-04-30
File page aligned mark files.

vyazelenko:
This PR implement mark file length alignment based on the `aeron.file.page.size` configured on the media driver. The file page size value is now stored in the CnC file.

### DIFF
--- a/aeron-client/src/main/java/io/aeron/Aeron.java
+++ b/aeron-client/src/main/java/io/aeron/Aeron.java
@@ -23,8 +23,25 @@ import io.aeron.exceptions.ConfigurationException;
 import io.aeron.exceptions.DriverTimeoutException;
 import io.aeron.logbuffer.FragmentHandler;
 import io.aeron.version.Versioned;
-import org.agrona.*;
-import org.agrona.concurrent.*;
+import org.agrona.BufferUtil;
+import org.agrona.CloseHelper;
+import org.agrona.DirectBuffer;
+import org.agrona.ErrorHandler;
+import org.agrona.ExpandableArrayBuffer;
+import org.agrona.Strings;
+import org.agrona.concurrent.AgentInvoker;
+import org.agrona.concurrent.AgentRunner;
+import org.agrona.concurrent.AtomicBuffer;
+import org.agrona.concurrent.BusySpinIdleStrategy;
+import org.agrona.concurrent.CountedErrorHandler;
+import org.agrona.concurrent.EpochClock;
+import org.agrona.concurrent.IdleStrategy;
+import org.agrona.concurrent.NanoClock;
+import org.agrona.concurrent.NoOpLock;
+import org.agrona.concurrent.SleepingMillisIdleStrategy;
+import org.agrona.concurrent.SystemEpochClock;
+import org.agrona.concurrent.SystemNanoClock;
+import org.agrona.concurrent.YieldingIdleStrategy;
 import org.agrona.concurrent.broadcast.BroadcastReceiver;
 import org.agrona.concurrent.broadcast.CopyBroadcastReceiver;
 import org.agrona.concurrent.ringbuffer.ManyToOneRingBuffer;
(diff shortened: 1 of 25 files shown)


aeron-client/src/main/java/io/aeron/Aeron.java
```diff
@@ -1902,68 +1886,5 @@ private void connectToDriver()
...
-
-        @SuppressWarnings("try")
-        private static MappedByteBuffer waitForFileMapping(
-            final File file, final EpochClock clock, final long deadlineMs)
-        {
-            while (true)
-            {
-                while (!file.exists() || file.length() < CncFileDescriptor.META_DATA_LENGTH)
-                {
-                    if (clock.time() > deadlineMs)
```
mikeb01, line 1914: While it is unlikely to have a millisecond timestamps go negative, I wonder if it is good style/practice for us to use the form:
```
if (clock.time() - deadlineMs < 0)
```

aeron-client/src/main/java/io/aeron/CncFileDescriptor.java
```diff
@@ -142,17 +149,18 @@ public class CncFileDescriptor
...
         START_TIMESTAMP_FIELD_OFFSET = CLIENT_LIVENESS_TIMEOUT_FIELD_OFFSET + SIZE_OF_LONG;
         PID_FIELD_OFFSET = START_TIMESTAMP_FIELD_OFFSET + SIZE_OF_LONG;
+        FILE_PAGE_SIZE_FIELD_OFFSET = PID_FIELD_OFFSET + SIZE_OF_LONG;
     }
 
     /**
      * Length of the metadata header for the CnC file.
      */
-    public static final int META_DATA_LENGTH = PID_FIELD_OFFSET + SIZE_OF_LONG;
+    public static final int META_DATA_LENGTH = CACHE_LINE_LENGTH * 2;
```
mikeb01, line 158: I'm wondering if this needed to change?  We do lose having a single constant that represents the unaligned length of the cnc metadata.  I'm not sure if this is a big drawback or not.

The Javadoc should be updated to reflect that it is the aligned length of the metadata.

aeron-cluster/src/main/java/io/aeron/cluster/service/ClusterMarkFile.java
```diff
@@ -111,40 +115,70 @@ public final class ClusterMarkFile implements AutoCloseable
...
      */
+    @Deprecated(forRemoval = true)
     public ClusterMarkFile(
         final File file,
         final ClusterComponentType type,
         final int errorBufferLength,
         final EpochClock epochClock,
         final long timeoutMs)
+    {
+        this(file, type, errorBufferLength, epochClock, timeoutMs, 0);
```
mikeb01, line 135: I would be inclined to use `MIN_PAGE_SIZE` instead of `0` here.  This will remove the extra checks for 0 to default it to that value.  Are there any other places where we end up passing a `0` to the ClusterMarkFile constructor?

aeron-cluster/src/main/java/io/aeron/cluster/service/ClusterMarkFile.java
```diff
@@ -111,40 +115,70 @@ public final class ClusterMarkFile implements AutoCloseable
...
+        final EpochClock epochClock,
+        final long timeoutMs,
+        final int filePageSize)
     {
         if (errorBufferLength < ERROR_BUFFER_MIN_LENGTH || errorBufferLength > ERROR_BUFFER_MAX_LENGTH)
         {
             throw new IllegalArgumentException("Invalid errorBufferLength: " + errorBufferLength);
         }
 
+        if (0 != filePageSize)
```
mikeb01, line 162: See the comments on disallowing `0` above, if we have to support it here.  I would do something like:
```
final int _filePageSize = 0 == filePageSize ? MIN_PAGE_SIZE : filePageSize;
```
And use the `_filePageSize`<sup>1</sup> throughout the rest of code.  This would simplify the code (i.e. 1 condition instead of 2).

<sup>1</sup> Or some better name.

aeron-client/src/main/java/io/aeron/CncFileDescriptor.java
mikeb01: It this to ensure consistent implementation between C & Java?

mikeb01:
Overall looks good.

aeron-client/src/main/java/io/aeron/CncFileDescriptor.java
vyazelenko: Yes, it was aligning C and Java impl. The former only had full aligned length.

aeron-cluster/src/main/java/io/aeron/cluster/service/ClusterMarkFile.java
vyazelenko: I went with 4KB default across all impls.

PR #1791, opened 2025-05-02
[Java] Make AeronCluster track connection status.

wojciech-adaptive:
Currently, AeronCluster does very little to track its connection status. All applications that want to react to, for example, losing network connectivity to the leader or the cluster shutting down, need to implement custom logic. This change provides such logic out of the box.

If connection is lost for more than newLeaderTimeoutNs, or a message can't be sent to a new leader for newLeaderConnectionTimeoutNs, the client will close itself.

For applications which call pollEgress/controlledPollEgress and check isClosed afterward no changes are needed. For applications which use the ingress publication or the egress subscription directly, two methods were added: trackIngressPublicationResult and pollStateChanges.

For the new leader timeout we currently use twice the default leader heartbeat timeout, in future we would ideally make the cluster announce the actual value.

Changed the semantics of pollEgress/controlledPollEgress return value, hopefully not a breaking change.

The state diagram is:

```
stateDiagram-v2
    [*] --> CONNECTED
    CONNECTED --> AWAIT_NEW_LEADER : ingress publication NOT_CONNECTED
    CONNECTED --> AWAIT_NEW_LEADER : ingress publication CLOSED
    CONNECTED --> PENDING_CLOSE : ingress publication MAX_POSITION_EXCEEDED
    CONNECTED --> AWAIT_NEW_LEADER : egress image closed
    CONNECTED --> AWAIT_NEW_LEADER_CONNECTION : NewLeaderEvent
    CONNECTED --> PENDING_CLOSE : SessionEvent CLOSED
    CONNECTED --> CLOSED : close
    AWAIT_NEW_LEADER --> AWAIT_NEW_LEADER_CONNECTION : NewLeaderEvent
    AWAIT_NEW_LEADER --> CLOSED : timeout
    AWAIT_NEW_LEADER --> CLOSED : close
    AWAIT_NEW_LEADER_CONNECTION --> CONNECTED : ingress offer successful
    AWAIT_NEW_LEADER_CONNECTION --> CLOSED : timeout
    AWAIT_NEW_LEADER_CONNECTION --> CLOSED : close
    PENDING_CLOSE --> CLOSED
    CLOSED --> [*]
```

### DIFF
--- a/aeron-cluster/src/main/java/io/aeron/cluster/client/AeronCluster.java
+++ b/aeron-cluster/src/main/java/io/aeron/cluster/client/AeronCluster.java
@@ -16,6 +16,7 @@
 package io.aeron.cluster.client;
 
 import io.aeron.*;
+import io.aeron.cluster.ConsensusModule;
 import io.aeron.cluster.codecs.*;
 import io.aeron.config.Config;
 import io.aeron.config.DefaultType;
@@ -62,11 +63,13 @@ public final class AeronCluster implements AutoCloseable
     private final long clusterSessionId;
     private long leadershipTermId;
     private int leaderMemberId;
-    private boolean isClosed;
     private final Context ctx;
     private final Subscription subscription;
+    private State state;
+    private long stateDeadline;
     private Image egressImage;
     private Publication publication;
+    private final NanoClock nanoClock;
     private final IdleStrategy idleStrategy;
     private final BufferClaim bufferClaim = new BufferClaim();
     private final UnsafeBuffer headerBuffer = new UnsafeBuffer(new byte[SESSION_HEADER_LENGTH]);
@@ -219,6 +222,7 @@ public final class AeronCluster implements AutoCloseable
         this.leaderMemberId = leaderMemberId;
         this.publication = publication;
 
+        this.nanoClock = ctx.aeron().context().nanoClock();
         this.idleStrategy = ctx.idleStrategy();
         this.egressListener = ctx.egressListener();
         this.fragmentAssembler = new FragmentAssembler(this::onFragment, 0, ctx.isDirectAssemblers());
@@ -230,6 +234,8 @@ public final class AeronCluster implements AutoCloseable
             .wrapAndApplyHeader(headerBuffer, 0, messageHeaderEncoder)
             .clusterSessionId(clusterSessionId)
             .leadershipTermId(leadershipTermId);
+
+        state(State.CONNECTED, 0);
     }
 
     /**
@@ -257,7 +263,12 @@ public final class AeronCluster implements AutoCloseable
      */
     public void close()
     {
-        if (null != publication && publication.isConnected() && !isClosed)
+        if (State.CLOSED == state)
+        {
+            return;
+        }
+
+        if (null != publication && publication.isConnected() && State.CONNECTED == state)
         {
             closeSession();
         }
@@ -269,18 +280,19 @@ public final class AeronCluster implements AutoCloseable
             CloseHelper.close(errorHandler, publication);
         }
 
-        isClosed = true;
+        state(State.CLOSED, 0);
         ctx.close();
     }
 
     /**
-     * Is the client closed? The client can be closed by calling {@link #close()} or the cluster sending an event.
+     * Is the client closed? The client can be closed by calling {@link #close()}, the cluster sending an event,
+     * or the client permanently losing a cluster connection.
      *
      * @return true if closed otherwise false.
      */
     public boolean isClosed()
     {
-        return isClosed;
+        return State.CLOSED == state;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


aeron-cluster/src/main/java/io/aeron/cluster/client/AeronCluster.java
```diff
@@ -1378,6 +1514,67 @@ public long messageTimeoutNs()
...
+        /**
+         * Set the timeout to wait for ingress publication to become connected with a new leader. Upon timeout the
+         * cluster will be considered lost and the client will close. If set to {@link Aeron#NULL_VALUE}, a reasonable
+         * default will be used.
+         *
+         * @param newLeaderConnectionTimeoutNs the new leader connection timeout in nanoseconds
+         *                                     or {@link Aeron#NULL_VALUE}.
+         * @return this for a fluent API.
+         */
+        public Context newLeaderConnectionTimeoutNs(final long newLeaderConnectionTimeoutNs)
```
mikeb01, line 1556: I don't think this configuration option is necessary, `messageTimeout` is sufficient.

aeron-cluster/src/main/java/io/aeron/cluster/client/AeronCluster.java
```diff
@@ -1378,6 +1514,67 @@ public long messageTimeoutNs()
...
+
+        /**
+         * The timeout to wait for a new leader after noticing a disconnection from the previous one. Upon timeout the
+         * cluster will be considered lost and the client will close.
+         *
+         * @return the new leader timeout in nanoseconds.
+         */
+        public long newLeaderTimeoutNs()
+        {
+            return newLeaderTimeoutNs;
```
mikeb01, line 1544: Should make use of `CommonContext::checkDebugTimeoutNs`.

aeron-cluster/src/main/java/io/aeron/cluster/client/AeronCluster.java
```diff
@@ -1378,6 +1514,67 @@ public long messageTimeoutNs()
...
+        /**
+         * The timeout to wait for a new leader after noticing a disconnection from the previous one. Upon timeout the
+         * cluster will be considered lost and the client will close. If set to {@link Aeron#NULL_VALUE}, a reasonable
+         * default will be used.
+         *
+         * @param newLeaderTimeoutNs the new leader timeout in nanoseconds or {@link Aeron#NULL_VALUE}.
+         * @return this for a fluent API.
+         */
+        public Context newLeaderTimeoutNs(final long newLeaderTimeoutNs)
+        {
```
mikeb01, line 1526: How much work would it be to embed the leader heartbeat timeout into the detail string of the login response or add it as an extra field (SessionEvent in cluster codec).  It would be nice not to require the extra configuration.  (Not a stopper for this fix, but would be good if it would fit into the next release).

aeron-cluster/src/main/java/io/aeron/cluster/client/AeronCluster.java
```diff
@@ -627,6 +679,36 @@ public void onNewLeader(
...
+    /**
+     * Updates the state of this client based on ingress publication result. Should be called with every {@code offer}
+     * and {@code tryClaim} result when {@link #ingressPublication()} is used directly. Methods of this class which send
+     * ingress messages call it automatically.
+     *
+     * @param result the result returned by the ingress publication.
+     */
+    public void trackIngressPublicationResult(final long result)
+    {
+        if (state == State.CONNECTED)
```
mikeb01, line 691: Style (constant on the left) `State.CONNECTED == state` (and instances below).

mikeb01:
Overall looks good, right direction.

aeron-cluster/src/main/java/io/aeron/cluster/client/AeronCluster.java
wojciech-adaptive: I'll have a look. For the detail string do you mean something like `leaderHeartbeatTimeoutNs=5000|foo=bar`? Having a field would be nicer, but in this case the codec is shared with other events, so for them it would be a waste of space, especially if we decide to add more information in future.

mikeb01: I think my preference would be for an additional field.  The only place this is used is session open/close so the redundant information wouldn't be too much overhead.  Having thought about it more I think that a separate field is better than encoding it in a string.  SBE has support for adding fields message versioning and compatibility where as arbitrary string encoding does not.

SBE should handle this (due to block length being encoded on the message) so older clients should safely ignore this field (but we can test for that).

mikeb01:
Only the `newLeaderTimeoutNs()` to resolve.  I think it would be worth finishing that before release.  However, this PR could be merged and handle that change as a separate PR.

PR #1792, opened 2025-05-02
[C] Fix issue for untethered slow consumers impacting whole server

fquinner:
This is related to https://github.com/aeron-io/aeron/issues/1790.

In the C aeronmd application, if the media driver is only dealing with untethered subscriptions, it will not calculate the aggregate subscription high water mark correctly, and so untethered subscriptions will never get image unavailable if they are a slow consumer among fast consumers.

This aligns aeron media driver functionality in this area with the Java media driver.


aeron-driver/src/main/c/aeron_publication_image.c
```diff
@@ -1035,11 +1035,8 @@ void aeron_publication_image_check_untethered_subscriptions(
     {
         aeron_tetherable_position_t *tetherable_position = &subscribable->array[i];
 
-        if (tetherable_position->is_tether)
-        {
-            int64_t position = aeron_counter_get_acquire(tetherable_position->value_addr);
-            max_sub_pos = position > max_sub_pos ? position : max_sub_pos;
-        }
+        int64_t position = aeron_counter_get_acquire(tetherable_position->value_addr);
```
mikeb01, line 1038: I think this change is partially correct, but if `!tetherable_position->is_tether` then we need to check the tethered state and exclude LINGERING untethered subscriptions to match with the Java implementation.

fquinner: I couldn't see where java considered RESTING in this calculation. I can see where its excluded afterwards when the subscriptions are iterated on again and state checked (as they are in C) but not in the java method which seems to calculate this position unless I'm missing something? https://github.com/aeron-io/aeron/blob/master/aeron-driver/src/main/java/io/aeron/driver/PublicationImage.java#L1138

mikeb01: They get removed from the list of `SubscriberPosition`s: https://github.com/aeron-io/aeron/blob/master/aeron-driver/src/main/java/io/aeron/driver/PublicationImage.java#L1113

fquinner: Ah yes thanks and we can't simply remove from subscription list here because (among other things) it will prevent RESTING state from transitioning to ACTIVE. OK will update the change today.

vyazelenko: @fquinner The condition is wrong, i.e. we need to check for the `RESTING` state if subscription is untethered:
```c
if (tetherable_position->is_tether || AERON_SUBSCRIPTION_TETHER_RESTING != tetherable_position->state)
```

fquinner: D'oh! OK thanks pushed in an update there now.

PR #1801, opened 2025-05-14
Add separate linger timeout for untethered subscriptions

ryanoftalltrees:
Adding a separate linger timeout for untethered subscriptions.  Code is in place.  Might need to re-order some things and add more tests.  also may need to adjust the description.

### DIFF
--- a/aeron-driver/src/main/java/io/aeron/driver/Configuration.java
+++ b/aeron-driver/src/main/java/io/aeron/driver/Configuration.java
@@ -1360,6 +1360,7 @@ public final class Configuration
      *
      * @return the timeout for an untethered subscription to remain in the linger state.
      * @see #UNTETHERED_LINGER_TIMEOUT_PROP_NAME
+     * @since 1.48.0
      */
     public static long untetheredLingerTimeoutNs()
     {
--- a/aeron-driver/src/main/java/io/aeron/driver/Configuration.java
+++ b/aeron-driver/src/main/java/io/aeron/driver/Configuration.java
@@ -1363,7 +1363,7 @@ public final class Configuration
      */
     public static long untetheredLingerTimeoutNs()
     {
-        return getDurationInNanos(UNTETHERED_LINGER_TIMEOUT_PROP_NAME, untetheredWindowLimitTimeoutNs());
+        return getDurationInNanos(UNTETHERED_LINGER_TIMEOUT_PROP_NAME, Aeron.NULL_VALUE);
     }
 
     /**
--- a/aeron-driver/src/main/java/io/aeron/driver/Configuration.java
+++ b/aeron-driver/src/main/java/io/aeron/driver/Configuration.java
@@ -843,6 +843,7 @@ public final class Configuration
     @Config
     public static final String UNTETHERED_LINGER_TIMEOUT_PROP_NAME = "aeron.untethered.linger.timeout";
 
+    private static final long DEFAULT_UNSET_VALUE_WINDOW_LIMIT_TIMEOUT_NS = TimeUnit.SECONDS.toNanos(5);
     /**
      * Default timeout for when an untethered subscription that is outside the window limit will participate in
      * local flow control.
@@ -851,7 +852,7 @@ public final class Configuration
         defaultType = DefaultType.LONG,
         defaultLong = 5_000_000_000L,
         expectedCDefaultFieldName = "AERON_UNTETHERED_WINDOW_LIMIT_TIMEOUT_NS_DEFAULT")
-    public static final long UNTETHERED_WINDOW_LIMIT_TIMEOUT_DEFAULT_NS = TimeUnit.SECONDS.toNanos(5);
+    public static final long UNTETHERED_WINDOW_LIMIT_TIMEOUT_DEFAULT_NS = DEFAULT_UNSET_VALUE_WINDOW_LIMIT_TIMEOUT_NS;
 
     /**
      * Default timeout for an untethered subscription to stay in the linger state.
@@ -860,7 +861,7 @@ public final class Configuration
         defaultType = DefaultType.LONG,
         defaultLong = 5_000_000_000L,
         expectedCDefaultFieldName = "AERON_UNTETHERED_LINGER_TIMEOUT_NS_DEFAULT")
-    public static final long UNTETHERED_LINGER_TIMEOUT_DEFAULT_NS = TimeUnit.SECONDS.toNanos(5);
+    public static final long UNTETHERED_LINGER_TIMEOUT_DEFAULT_NS = DEFAULT_UNSET_VALUE_WINDOW_LIMIT_TIMEOUT_NS;
 
     /**
      * Property name of the timeout for when an untethered subscription is resting after not being able to keep up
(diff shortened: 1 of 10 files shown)


aeron-driver/src/main/java/io/aeron/driver/PublicationParams.java
```diff
@@ -488,6 +490,28 @@ private void getUntetheredWindowLimitTimeout(final ChannelUri channelUri, final
...
+            untetheredLingerTimeoutNs = SystemUtil.parseDuration(UNTETHERED_LINGER_TIMEOUT_PARAM_NAME, timeoutString);
+        }
+        else if (null != windowLimitTimeoutString)
+        {
+            // window limit was explicitly set for the channel.  use it
+            untetheredLingerTimeoutNs = untetheredWindowLimitTimeoutNs;
+        }
+        else
+        {
+            // neither value was explicitely set for the channel.  use the context value for linger
```
nickelaway, line 509: typo: explicitely
Also, first time I read this comment I read the two lines as being a single sentence. I would favour capitalising and punctuating the comments to make things a bit clearer, otherwise it reads like this:
"use the context value for linger context value for linger needs similar checks to default to window timeout if not set"

aeron-driver/src/main/java/io/aeron/driver/MediaDriver.java
```diff
@@ -464,6 +464,7 @@ public static final class Context extends CommonContext
         private long publicationConnectionTimeoutNs = Configuration.publicationConnectionTimeoutNs();
         private long publicationLingerTimeoutNs = Configuration.publicationLingerTimeoutNs();
         private long untetheredWindowLimitTimeoutNs = Configuration.untetheredWindowLimitTimeoutNs();
+        private long untetheredLingerTimeoutNs = Configuration.untetheredLingerTimeoutNs();
```
mikeb01, line 467: I think would be simpler implementation if we set this value to be Aeron.NULL_VALUE initially, then check for that value during the `MediaDriver.Context::conclude` call and set it to the window timeout there.

mikeb01:
For the most part this looks good.

We don't have a lot of strong testing with tethering timeouts.  You could look to expand the `UntetheredSubscriptionTest` tests (there is a unit and system test).

We should also update the ChannelUriStringBuilder.  This will also need to be reflected in the C driver as well as the C++ API/Wrapper.

vyazelenko (changes requested):
Needs to be implemented in the C media driver as well.

aeron-driver/src/main/java/io/aeron/driver/PublicationParams.java
```diff
@@ -488,6 +490,28 @@ private void getUntetheredWindowLimitTimeout(final ChannelUri channelUri, final
...
+            untetheredLingerTimeoutNs = untetheredWindowLimitTimeoutNs;
+        }
+        else
+        {
+            // Neither Window Limit Timeout nor Linger Timeout was set for the channel.
+            // Use the Context value for Linger.
+            untetheredLingerTimeoutNs = ctx.untetheredLingerTimeoutNs();
+        }
+    }
+
```
vyazelenko, line 514: We should only check for `UNTETHERED_WINDOW_LIMIT_TIMEOUT_PARAM_NAME` if `UNTETHERED_LINGER_TIMEOUT_PARAM_NAME` not defined.

vyazelenko (changes requested):

vyazelenko:
Also as Mike already mentioned:
> This will also need to be reflected in the C driver as well as the C++ API/Wrapper.

aeron-driver/src/test/java/io/aeron/driver/MediaDriverContextTest.java
```diff
@@ -355,4 +356,50 @@ private static ErrorHandler getErrorHandler(final UdpTransportPoller transportPo
...
+
+        final MediaDriver.Context ctx = new MediaDriver.Context();
+        try
+        {
+            assertEquals(TimeUnit.MILLISECONDS.toNanos(222), ctx.untetheredLingerTimeoutNs());
+            assertEquals(TimeUnit.MILLISECONDS.toNanos(444), ctx.untetheredWindowLimitTimeoutNs());
+        }
+        finally
+        {
+            ctx.close();
```
vyazelenko, line 381: No need to close here.

aeron-driver/src/test/java/io/aeron/driver/MediaDriverContextTest.java
```diff
@@ -355,4 +356,50 @@ private static ErrorHandler getErrorHandler(final UdpTransportPoller transportPo
...
+
+    @Test
+    void shouldTestDefaultUntetheredLingerTimeoutAndWindowTimeout()
+    {
+        assertEquals(UNTETHERED_WINDOW_LIMIT_TIMEOUT_DEFAULT_NS, context.untetheredWindowLimitTimeoutNs());
+        assertEquals(Aeron.NULL_VALUE, context.untetheredLingerTimeoutNs());
+    }
+
+    @Test
+    void shouldHonorSystemPropertyOverrideUntetheredLingerTimeoutAndWindowTimeout()
```
vyazelenko, line 368: Split the test into two tests, i.e. test both properties separately.

aeron-driver/src/test/java/io/aeron/driver/MediaDriverContextTest.java
```diff
@@ -355,4 +356,50 @@ private static ErrorHandler getErrorHandler(final UdpTransportPoller transportPo
         field.setAccessible(true);
         return (ErrorHandler)field.get(transportPoller);
     }
+
+    @Test
+    void shouldTestDefaultUntetheredLingerTimeoutAndWindowTimeout()
+    {
+        assertEquals(UNTETHERED_WINDOW_LIMIT_TIMEOUT_DEFAULT_NS, context.untetheredWindowLimitTimeoutNs());
+        assertEquals(Aeron.NULL_VALUE, context.untetheredLingerTimeoutNs());
```
vyazelenko, line 364: Split into two tests.

aeron-driver/src/main/java/io/aeron/driver/Configuration.java
```diff
@@ -837,6 +837,12 @@ public final class Configuration
     @Config
     public static final String UNTETHERED_WINDOW_LIMIT_TIMEOUT_PROP_NAME = "aeron.untethered.window.limit.timeout";
 
+    /**
+     * Property name of the timeout for an untethered subscription to stay in the linger state.
+     */
+    @Config
```
vyazelenko, line 843: Add `@since`.

aeron-driver/src/main/java/io/aeron/driver/MediaDriver.java
```diff
@@ -1227,6 +1228,32 @@ public Context untetheredWindowLimitTimeoutNs(final long timeoutNs)
...
         }
 
+        /**
+         * The linger timeout for an untethered subscription.
+         *
+         * @return timeout that an untethered subscription will linger.
+         * @see Configuration#UNTETHERED_LINGER_TIMEOUT_PROP_NAME
+         */
+        @Config
+        public long untetheredLingerTimeoutNs()
```
vyazelenko, line 1238: `@since`

aeron-driver/src/test/java/io/aeron/driver/MediaDriverContextTest.java
```diff
@@ -355,4 +356,50 @@ private static ErrorHandler getErrorHandler(final UdpTransportPoller transportPo
...
+        try
+        {
+            assertEquals(Aeron.NULL_VALUE, ctx.untetheredLingerTimeoutNs());
+            assertEquals(TimeUnit.MILLISECONDS.toNanos(444), ctx.untetheredWindowLimitTimeoutNs());
+        }
+        finally
+        {
+            ctx.close();
+            System.clearProperty("aeron.untethered.linger.timeout");
+            System.clearProperty("aeron.untethered.window.limit.timeout");
```
vyazelenko, line 402: See above.

PR #1804, opened 2025-05-20
[Java] ConsensusModule.Context and ClusterBackup.Context to support shared Media Driver agent invoker.

the-thing:
Currently `ConsensusModule.Context` and `ClusterBackup.Context` don't support shared media driver invoker and must rely on Archive (or MediaDriver) running in non-INVOKER mode.

`ClusteredMediaDriver#launch` and `ClusterBackupMediaDriver#launch` will timeout when 

- both `MediaDriver` and `Archive` are both running in the `Invoker` threading mode 
- Aeron client is not explicitly provided

```
Exception in thread "main" io.aeron.exceptions.DriverTimeoutException: FATAL - MediaDriver (...) keepalive: age=10013ms > timeout=10000ms
	at io.aeron.ClientConductor.checkLiveness(ClientConductor.java:1601)
	at io.aeron.ClientConductor.checkTimeouts(ClientConductor.java:1564)
	at io.aeron.ClientConductor.service(ClientConductor.java:1470)
	at io.aeron.ClientConductor.awaitResponse(ClientConductor.java:1529)
	at io.aeron.ClientConductor.addCounter(ClientConductor.java:1056)
	at io.aeron.Aeron.addCounter(Aeron.java:523)
	at io.aeron.cluster.service.ClusterCounters.allocateVersioned(ClusterCounters.java:98)
	at io.aeron.cluster.ConsensusModule$Context.conclude(ConsensusModule.java:1764)
	at io.aeron.cluster.ConsensusModule.<init>(ConsensusModule.java:238)
	at io.aeron.cluster.ConsensusModule.launch(ConsensusModule.java:290)
	at io.aeron.cluster.ClusteredMediaDriver.launch(ClusteredMediaDriver.java:103)
```


vyazelenko:
@the-thing This issue is much more complicated than just passing a MediaDriver's `AgenInvoker` to the underlying Aeron client instances. For instance, the `AgentInvoker` must be called during normal duty cycle in order to ensure that the stream counters are updated and log buffers are cleaned (e.g. [ArchiveConductor#invokeDriverConductor](https://github.com/aeron-io/aeron/blob/f0b11ed03e354d2eda658429f7dd16e22104d90b/aeron-archive/src/main/java/io/aeron/archive/ArchiveConductor.java#L321)). This also applies to Archive running in the invoker mode, i.e. it needs to be called by the ConsensusModule/ClusterBackup/ClusteredServiceContainer.

Additionally, invoker mode can only work if there is only thread used, i.e. all components **must** run in the invoker mode. Nesting invoker callbacks becomes an issue.

I've added an item to the Aeron roadmap to address this issue. However, there are no immediate plans to work on it. Doing it properly will require substantial effort as we would need to ensure that all Cluster tests will pass in the invoker mode.

PR #1818, opened 2025-06-12
Cluster tool exit code bugfix.

zorba128:
Cluster tool no longer returns exit codes - what makes it impossible to use commands such as `is-leader`.


vyazelenko:
@zorba128 Thanks for the fix.

PR #1819, opened 2025-06-12
prototype response publication

nbradac:


mikeb01:
This is the general direction that I would like to go in.  It would be good to get the build passing.  There is a checkstyle failure at the moment.

mikeb01:
We should at a test to show that negative response correlation id values are rejected.  E.g. a user can't pass in the value `-2` explicitly.  They have to use the keyword.

mikeb01:
Merged through different PR.

PR #1824, opened 2025-06-26
[Java] Allow ConsensusModuleExtension to do work while election is in progress.

wojciech-adaptive:


wojciech-adaptive:
Superseded by #1827.

PR #1835, opened 2025-07-03
Fix aeron_archive/cpp_wrapper CMakeLists.txt to reference the correct source path

mystconqueror:


vyazelenko:
Fixed in https://github.com/aeron-io/aeron/commit/e89f5f639910a2f4a3eee7177f37f63077fe211d.

PR #1841, opened 2025-07-09
Aeron 225 badclientconfig

ryanoftalltrees:
Fix for https://weareadaptive.atlassian.net/browse/AERON-225


ryanoftalltrees:
A unit test for the new check is failing.  It doesn't fail locally, so I need to look into it.  Just tried changing the pattern match for how it writes to standard err and will see if that fixes it.

PR #1851, opened 2025-08-19
[Java] driver: Prefer multicast address bind to avoid port conflicts

ChongHan:
When opening a multicast receive socket, attempt to bind to the multicast group address and port first.

- Reduces likelihood of port conflicts when multicast-address binding is supported
- Preserves existing behaviour via fallback on platforms that don't support it


ChongHan:
I don't have mac to reproduce the issue flagged by CI; 

I could change `try` to `if (SystemUtil.isLinux())`?

PR #1858, opened 2025-09-02
Network Partition Testing

mikeb01:
- Add the ability set custom addresses for a cluster test.
- Bind the log and consensus channels to their associated IP addresses.
- Use IP tables to create network partitions between nodes.


aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModule.java
```diff
@@ -1558,6 +1596,8 @@ public static final class Context implements Cloneable
         private boolean useAgentInvoker = false;
         private ConsensusModuleStateExport boostrapState = null;
         private boolean acceptStandbySnapshots = Configuration.acceptStandbySnapshots();
+        private boolean bindConsensusControl = Configuration.bindConsensusControl();
+        private boolean bindLogControl = Configuration.bindLogControl();
```
vyazelenko: Naming: since these setting related to `consensusChannel` and `logChannel` this should be reflected in the names, e.g. `bindConsensusChannelControl` or `bindControlOnConsensusChannel`.

aeron-test-support/src/main/java/io/aeron/test/cluster/TestCluster.java
```diff
@@ -2251,6 +2281,17 @@ public TestCluster start(final int toStart)
...
+//            {
+//                hostnames = Stream.generate(() -> "localhost").limit(nodeCount).toList();
+//            }
+//
+//            if (nodeCount != hostnames.size())
+//            {
+//                throw new IllegalArgumentException(
+//                    "number of hostnames/address=" + hostnames + " must match nodeCount=" + nodeCount);
+//            }
+
```
vyazelenko: Debug code?

aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModule.java
mikeb01: I'm going with `enableControlOnConsensusChannel`.  I've decided not to use the work `bind` as that is more of an implementation specific detail.  We've used `enable`/`disable` naming in the past.

PR #1860, opened 2025-09-08
Fix/create subscription for extension

mikeb01:
Have the ConsensusModule create a subscription for use by the leader node instead of requiring the extension to create it.  This will prevent the extension from getting a subscription that is not in lock-step position-wise when compared to the images in the Cluster itself.

### DIFF
--- a/aeron-test-support/src/main/java/io/aeron/test/cluster/TestCluster.java
+++ b/aeron-test-support/src/main/java/io/aeron/test/cluster/TestCluster.java
@@ -172,7 +172,7 @@ public final class TestCluster implements AutoCloseable
     private String clusterBaseDir;
     private ClusterBackup.Configuration.ReplayStart replayStart;
     private List<String> hostnames;
-    private boolean useExtension = false;
+    private boolean hasExtension = false;
 
     private TestCluster(
         final int staticMemberCount,
@@ -289,7 +289,7 @@ public final class TestCluster implements AutoCloseable
         final String aeronDirName = CommonContext.generateRandomDirName();
         final File markFileDir = null != markFileBaseDir ? new File(markFileBaseDir, "mark-" + index) : null;
         final TestNode.Context context = new TestNode.Context(serviceSupplier.apply(index), nodeNameMappings());
-        context.useExtension = useExtension;
+        context.hasExtension = hasExtension;
 
         context.aeronArchiveContext
             .lock(NoOpLock.INSTANCE)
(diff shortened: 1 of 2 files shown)


aeron-cluster/src/main/java/io/aeron/cluster/ConsensusControlState.java
```diff
@@ -27,6 +28,7 @@
 public final class ConsensusControlState
 {
     private final ExclusivePublication logPublication;
+    private final Subscription leaderLogSubscription;
```
wojciech-adaptive, line 31: Needs a getter.

aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModuleAgent.java
```diff
@@ -1780,6 +1785,30 @@ void joinLogAsLeader(
             Long.MAX_VALUE,
             isStartup,
             Cluster.Role.LEADER);
+
+        connectLeaderLogSubscriptionForExtension(logPosition);
```
vyazelenko, line 1805: Wouldn't it be safer/cleaner to create extensions Subscription before waiting for services to join the log?

aeron-system-tests/src/test/java/io/aeron/cluster/ClusterTest.java
```diff
@@ -203,6 +203,19 @@ void shouldNotSnapshotOnPrimaryClusterWhenStandbySnapshotIsRequested()
...
+    @InterruptAfter(10)
+    void shouldStartCluster()
+    {
+        cluster = aCluster().withStaticNodes(3)
+            .withExtension(true)
+            .withServiceSupplier(value -> new TestNode.TestService[0])
+            .start();
+
+        systemTestWatcher.cluster(cluster);
+        cluster.awaitLeader();
```
vyazelenko, line 254: Would good to assert that `leaderLogSubscription` was created and is available from `ConsensusModuleExtension#onElectionComplete` callback.

aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModuleAgent.java
mikeb01: Generally speaking when we are running the extension we're not running with any services, so it probably doesn't matter.  If we were doing both, then they need to sync to the same point on the log, it shouldn't make too much difference.

PR #1864, opened 2025-09-24
Keep LogAdapter's buffer builder up to date in the case of a live leader

marc-adaptive:


aeron-cluster/src/main/java/io/aeron/cluster/ConsensusControlState.java
```diff
@@ -16,81 +16,24 @@
...
 /**
  * The state needed to allow control of the consensus module.
  * <p>
  * This is a record object being passed to external entities.
  *
+ * @param logPublication        publication or null.
+ * @param logRecordingId        log recording id.
+ * @param leadershipTermId      leadership term id.
  * @see ConsensusModuleExtension
  */
```
pveentjer, line 29: If this is part of the public API, could this not lead to binary incompatibility? 

So a user is using this class and compiled against an older version of Aeron (with a class based ConsensusControlState) and later runs with a newer version of Aeron (with a record based ConsensusControlState).

I made a small demo application to demonstrate binary incompatibility and till so far I was not able to demonstrate there is a problem. But this doesn't mean I have covered all cases.

aeron-cluster/src/main/java/io/aeron/cluster/ConsensusControlState.java
```diff
@@ -16,81 +16,24 @@
...
-public final class ConsensusControlState
+public record ConsensusControlState(ExclusivePublication logPublication, long logRecordingId, long leadershipTermId)
 {
```
pveentjer, line 31: Dropping the leaderLogSubscription will definitely cause incompatibility.

aeron-cluster/src/main/java/io/aeron/cluster/ConsensusControlState.java
vyazelenko: @pveentjer `ConsensusModuleExtension` is internal API despite being public. So any changes to that API is fair game.

aeron-cluster/src/main/java/io/aeron/cluster/ConsensusControlState.java
vyazelenko: @pveentjer `leaderLogSubscription` was added two weeks ago and is now removed, i.e. we didn't have public releases with it yet plus API is internal anyhow.

PR #1866, opened 2025-09-25
[Cluster] Handle ClusterTermination when using consensus module extension.

marc-adaptive:


aeron-cluster/src/main/java/io/aeron/cluster/ClusterTermination.java
```diff
@@ -23,11 +23,13 @@ class ClusterTermination
 {
     private long deadlineNs;
     private boolean haveServicesTerminated;
+    private boolean extensionCanTerminate;
 
-    ClusterTermination(final long deadlineNs, final int serviceCount)
+    ClusterTermination(final long deadlineNs, final int serviceCount, final ConsensusModuleExtension extension)
```
vyazelenko: Since `ConsensusModuleExtension` is not used after initializing `extensionCanTerminate` field I suggest to pass a boolean `extensionCanTerminate` instead.

vyazelenko (approved):

PR #1870, opened 2025-10-01
[Cluster] Update the default AuthorisationService to allow requests necessary for Cluster Standby

nickelaway:


aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModule.java
```diff
@@ -790,15 +790,12 @@ public static final class Configuration
...
             "aeron.cluster.authorisation.service.supplier";
 
-        static final AuthorisationService ALLOW_ONLY_BACKUP_QUERIES = (protocolId, actionId, type, encodedPrincipal) ->
-            MessageHeaderDecoder.SCHEMA_ID == protocolId && BackupQueryDecoder.TEMPLATE_ID == actionId;
-
         /**
          * Default {@link AuthorisationServiceSupplier} that returns {@link AuthorisationService} that forbids all
          * command from being executed (i.e. {@link AuthorisationService#DENY_ALL}).
          */
         public static final AuthorisationServiceSupplier DEFAULT_AUTHORISATION_SERVICE_SUPPLIER =
```
vyazelenko: JavaDoc should be updated describing which commands are allowed.

PR #1875, opened 2025-10-08
[Java] AeronCluster client requires explicit Aeron client when MediaDriver running in INVOKER threading mode

the-thing:
Hi,

I upgraded Aeron from 1.48.2-> 1.49.0 and I noticed that the behavior has changed around cluster client connectivity. It seems that it is not possible to run media driver in the INVOKER threading mode and use implicit `io.aeron.Aeron` client.

Please see `shouldAsyncConnectWhenMediaDriverUsesInvokerModeAndUsesImplicitAeronClient` test case. Similar test would have passed in older versions (e.g. 1.48.2).

The solution is to provide external `io.aeron.Aeron` client explicitly that uses driver agent invoker.

I had a look through recent release notes and I didn't see anything that would clearly indicate that this behavior has changed and I don't see test cases covering it. Was it intentional?

### DIFF
--- a/aeron-system-tests/src/test/java/io/aeron/cluster/ClusterTest.java
+++ b/aeron-system-tests/src/test/java/io/aeron/cluster/ClusterTest.java
@@ -3071,6 +3071,123 @@ class ClusterTest
         cluster.stopAllNodes();
     }
 
+    @Test
+    @InterruptAfter(40)
+    @Disabled
+    void shouldAsyncConnectWhenMediaDriverUsesInvokerModeAndUsesImplicitAeronClient()
+    {
+        cluster = aCluster().withStaticNodes(3).start();
+        systemTestWatcher.cluster(cluster);
+
+        final TestNode leader = cluster.awaitLeader();
+        assertEquals(1, leader.consensusModule().context().electionCounter().get());
+
+        final long leadershipTermId = leader.consensusModule().context().leadershipTermIdCounter().get();
+        assertNotEquals(-1, leadershipTermId);
+
+        final List<TestNode> followers = cluster.followers();
+
+        for (final TestNode follower : followers)
+        {
+            assertEquals(1, follower.consensusModule().context().electionCounter().get());
+            assertEquals(leadershipTermId, follower.consensusModule().context().leadershipTermIdCounter().get());
+        }
+
+        final MediaDriver.Context mediaDriverContext = cluster.clientMediaDriverCtx()
+                .threadingMode(ThreadingMode.INVOKER);
+
+        try (MediaDriver mediaDriver = MediaDriver.launch(mediaDriverContext))
+        {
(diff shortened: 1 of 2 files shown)


vyazelenko:
@the-thing What you observing with the original test is expected, i.e. any blocking call on `Aeron` instance will timeout if media driver is configured with `ThreadingMode.INVOKER` but the client is configured without `Aeron.Context#driverAgentInvoker(AgentInvoker)`. The reason is waiting for a response from media driver but no one is executing the duty cycle for it. Therefore what you are doing in a second test is correct, i.e. you need to configure the client with `Aeron.Context#driverAgentInvoker(AgentInvoker)` and `Aeron.Context#useConductorAgentInvoker(true)`.

the-thing:
@vyazelenko

I understand the explanation and I modified to code to provide the explicit Aeron client with correct settings.

My problem was:

1) The behavior has changed.

2) It is still possible to create an invalid combination of the `MediaDriver` and the `Aeron` client that are not compatible with each other by running `MediaDriver` in the `INVOKER` threading mode and either not proving `Aeron` client or providing it with invalid configuration - `useConductorAgentInvoker(false)` and `driverAgentInvoker(null)`.  In this scenario either cluster connection will immediately be established or never - `io.aeron.cluster.client.AeronCluster.AsyncConnect#poll` never advances. Configuration doesn't complain about this setup and I think it should about this particular set of flags.

vyazelenko:
@the-thing [Aeron 1.49.1](https://github.com/aeron-io/aeron/releases/tag/1.49.1) restores "compatibility" with the previous code. Running media driver with the `INVOKER` mode will "work" when using `AeronCluster`, i.e. the initial connection to the Cluster will be established. However, upon leadership change `AeronCluster` instance will break attempting to create a `Publication` to the new leader.

`Aeron` client does validate `useConductorAgentInvoker` usage:

|`useConductorAgentInvoker`|`driverAgentInvoker`| description |
|--------------------------|--------------------|-------------|
|`false`|`null`|client creates a dedicated conductor thread (default)|
|`false`|not `null`|invalid configuration (error)|
|`true`|`null`|no dedicated conductor thread is created|
|`true`|not `null`|client is wired to invoke media driver which is running in `INVOKER` mode|

the-thing:
@vyazelenko 

Thanks for "fixing" this (I'm not sure why you are quoting some words ;)).

After this change it seems that additional validation is not required, but the validation that I was referring to would be required for the 1.49.0 behavior somewhere in `io.aeron.cluster.client.AeronCluster.Context#conclude` (`useConductorAgentInvoker(false)` and `driverAgentInvoker(null)`) or `io.aeron.archive.client.AeronArchive#connect(io.aeron.archive.client.AeronArchive.Context)`. Now this change is obsolete. Validation performed in `io.aeron.Aeron.Context#conclude` never required any change. 

I will update on Monday and see how the tests behave. Thanks for looking.

PR #1880, opened 2025-10-31
allow custom aeron directory in TestCluster

ryanoftalltrees:


aeron-test-support/src/main/java/io/aeron/test/cluster/TestCluster.java
```diff
@@ -305,7 +307,7 @@ public TestNode startStaticNode(
         final int index, final boolean cleanStart, final IntFunction<TestNode.TestService[]> serviceSupplier)
     {
         final String baseDirName = clusterBaseDir + "-" + index;
-        final String aeronDirName = CommonContext.generateRandomDirName();
+        final String aeronDirName = aeronBaseDir + "-" + UUID.randomUUID();
```
vyazelenko: Use `"-" + index` like we do for the `baseDirName`.

aeron-test-support/src/main/java/io/aeron/test/cluster/TestCluster.java
```diff
@@ -411,7 +413,7 @@ public TestBackupNode startClusterBackupNode(
     {
         final int index = memberCount;
         final String baseDirName = clusterBaseDir + "-" + index;
-        final String aeronDirName = CommonContext.generateRandomDirName();
+        final String aeronDirName = aeronBaseDir + "-" + UUID.randomUUID();
```
vyazelenko: Use `"-" + index` like we do for the `baseDirName`.

aeron-test-support/src/main/java/io/aeron/test/cluster/TestCluster.java
```diff
@@ -482,7 +484,7 @@ public TestNode startStaticNodeFromBackup()
     public TestNode startStaticNodeFromBackup(final IntFunction<TestNode.TestService[]> serviceSupplier)
     {
         final String baseDirName = clusterBaseDir + "-" + backupNodeIndex;
-        final String aeronDirName = CommonContext.generateRandomDirName();
+        final String aeronDirName = aeronBaseDir + "-" + UUID.randomUUID();
```
vyazelenko: Use `"-" + index` like we do for the `baseDirName`.

aeron-test-support/src/main/java/io/aeron/test/cluster/TestCluster.java
```diff
@@ -766,7 +768,7 @@ public TestMediaDriver startClientMediaDriver()
     {
         if (null == clientMediaDriver)
         {
-            final String aeronDirName = CommonContext.generateRandomDirName();
+            final String aeronDirName = aeronBaseDir + "-" + UUID.randomUUID();
```
vyazelenko: Use `"-" + index` like we do for the `baseDirName`.

aeron-test-support/src/main/java/io/aeron/test/cluster/TestCluster.java
```diff
@@ -2137,6 +2139,7 @@ public static final class Builder
         private File markFileBaseDir = null;
         private String clusterBaseDir = System.getProperty(
             CLUSTER_BASE_DIR_PROP_NAME, CommonContext.generateRandomDirName());
+        private String aeronBaseDir = CommonContext.AERON_DIR_PROP_DEFAULT;
```
vyazelenko: This should be replaced with `CommonContext.generateRandomDirName()` to ensure that tests don't interact badly.

aeron-system-tests/src/test/java/io/aeron/cluster/AppointedLeaderTest.java
vyazelenko, line 1: Instead of changing existing tests to use `aeronBaseDir` you should write a dedicated test and add it to the `ClusterTest` class. There you should verify that drivers of all nodes as well as the client driver all use the same base dir.

ryanoftalltrees: Will do

aeron-test-support/src/main/java/io/aeron/test/cluster/TestCluster.java
ryanoftalltrees: My intention here was to leave the default behavior exactly as it is now.  Currently, it uses CommonContext.generateRandomDirName() for each individual aeron dir in the cluster.  (This was modified in July, '25).  Using a base dir + "-" + random UUID gives the exact same behavior as current if base dir isn't overridden.  Your suggestion here wouldn't allow for a different aeron dir for each node.  Happy to change it to use the index though as that makes sense.  but I'm guessing you changed the behavior to absolutely ensure each node gets a fully unique aeron dir?

vyazelenko: Using `CommonContext.generateRandomDirName()` and adding `index` for each node will ensure just that. You'll also need to ensure that client driver will use a unique `aeron.dir`.

ryanoftalltrees: ok, got it.  thanks

aeron-system-tests/src/test/java/io/aeron/cluster/TestClusterTest.java
```diff
@@ -0,0 +1,86 @@
...
+public class TestClusterTest
+{
+
+    @RegisterExtension
+    final SystemTestWatcher systemTestWatcher = new SystemTestWatcher();
+
+    private String aeronDirectory;
+
+    @BeforeEach
+    void setup(@TempDir final Path tempDir)
```
vyazelenko: Inject this directory at the test level not the class.

aeron-system-tests/src/test/java/io/aeron/cluster/TestClusterTest.java
```diff
@@ -0,0 +1,86 @@
...
+    {
+        aeronDirectory = tempDir.toString();
+    }
+
+    @Test
+    @InterruptAfter(20)
+    void testCustomAeronDirectory()
```
vyazelenko: Needs `TempDir` injected.

aeron-system-tests/src/test/java/io/aeron/cluster/TestClusterTest.java
```diff
@@ -0,0 +1,86 @@
...
+
+        final TestNode leader = cluster.awaitLeader();
+        assertNotNull(cluster.connectClient());
+
+        final Set<String> seen = new HashSet<>();
+        for (int i = 0; i < cluster.memberCount(); i++)
+        {
+            final String dir = cluster.node(i).mediaDriver().context().aeronDirectoryName();
+            assertTrue(seen.add(dir), "Duplicate Aeron dir: " + dir);
+            assertFalse(dir.startsWith("/dev/shm"), "Aeron dir under /dev/shm: " + dir);
```
vyazelenko: Instead of a negative assert use a positive one, i.e. check that the directory starts with the `baseDir`. Please use `MatcherAssert.assertThat` API, i.e.:
```java
MatcherAssert.assertThat(dir, Matchers.startsWith(aeronDirectory));
```

aeron-system-tests/src/test/java/io/aeron/cluster/TestClusterTest.java
```diff
@@ -0,0 +1,86 @@
...
+        }
+
+        final String dir = clientCtx.aeronDirectoryName();
+        assertTrue(seen.add(dir), "Duplicate Aeron dir: " + dir);
+        assertFalse(dir.startsWith("/dev/shm"), "Aeron dir under /dev/shm: " + dir);
```
vyazelenko: Same as above.

aeron-test-support/src/main/java/io/aeron/test/cluster/TestCluster.java
```diff
@@ -2137,6 +2138,7 @@ public static final class Builder
         private File markFileBaseDir = null;
         private String clusterBaseDir = System.getProperty(
             CLUSTER_BASE_DIR_PROP_NAME, CommonContext.generateRandomDirName());
+        private String aeronBaseDir = CommonContext.generateRandomDirName();
```
vyazelenko, line 2141: Should also use `aerobBaseDir` with a fixed suffix (e.g. `-client`).

vyazelenko (changes requested):

aeron-test-support/src/main/java/io/aeron/test/cluster/TestCluster.java
ryanoftalltrees: Sorry, but I don't understand this request.  You want the default aeron base dir to have "-client" at the end of it?  or just on the client media driver?

vyazelenko: Just the client driver. As for the default value it should be `CommonContext.generateRandomDirName()` like you already had.

ryanoftalltrees: Great, thanks

PR #1885, opened 2025-11-07
Update configuration to make retry attempts configurable. Deprecate `…

thughescmc:
This is a rebased version of https://github.com/aeron-io/aeron/pull/1852 as requested. Sorry I missed the notifications before.

This pull request introduces a new configuration property for controlling the number of retry attempts when offering requests to the Aeron Archive, moving this setting from ArchiveProxy to the main configuration. It also updates the usage and documentation to reflect this change, deprecates the old constant in ArchiveProxy, and ensures that the value is configurable via the context and system properties.

### DIFF
--- a/aeron-archive/src/main/java/io/aeron/archive/client/ArchiveProxy.java
+++ b/aeron-archive/src/main/java/io/aeron/archive/client/ArchiveProxy.java
@@ -68,6 +68,7 @@ import org.agrona.concurrent.NanoClock;
 import org.agrona.concurrent.SystemNanoClock;
 import org.agrona.concurrent.YieldingIdleStrategy;
 
+import static io.aeron.archive.client.AeronArchive.Configuration.MESSAGE_RETRY_ATTEMPTS_DEFAULT;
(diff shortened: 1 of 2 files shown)


aeron-archive/src/main/java/io/aeron/archive/client/ArchiveProxy.java
```diff
@@ -77,8 +78,10 @@ public final class ArchiveProxy
 {
     /**
      * Default number of retry attempts to be made when offering requests.
+     * @deprecated Use {@link AeronArchive.Configuration#MESSAGE_RETRY_ATTEMPTS_DEFAULT}.
      */
-    public static final int DEFAULT_RETRY_ATTEMPTS = 3;
+    @Deprecated(since = "1.48.6", forRemoval = true)
+    public static final int DEFAULT_RETRY_ATTEMPTS = MESSAGE_RETRY_ATTEMPTS_DEFAULT;
```
vyazelenko, line 84: Why deprecate this field?

PR #1897, opened 2025-12-02
fix: double free in aeron_archive_context_t

klittlepage:
Calling `aeron_archive_context_duplicate` on a context where `_ctx->owns_idle_strategy == true` will result in a context that also attempts to own the idle strategy. This results in a double free when calling `aeron_archive_context_close`.


vyazelenko:
@klittlepage Good catch.

PR #1901, opened 2025-12-12
[Java/C] Allow to specify network interface name in the interface channel parameter.

wojciech-adaptive:
Format with curly braces chosen so that we can unambiguously tell it apart from hostnames.

### DIFF
--- a/aeron-client/src/main/c/util/aeron_netutil.c
+++ b/aeron-client/src/main/c/util/aeron_netutil.c
@@ -51,6 +51,8 @@ __inline DWORD64 __builtin_popcountll(DWORD64 operand)
 #error Unsupported platform!
 #endif
 
+#define AERON_NETUTIL_NAMED_INTERFACE_OPENING_CHAR '{'
+
 int aeron_ip_addr_resolver(const char *host, struct sockaddr_storage *sockaddr, int family_hint, int protocol)
 {
     if (-1 == aeron_net_init())
@@ -511,6 +513,43 @@ int aeron_ip_lookup_func(
     return 0;
 }
 
+struct lookup_by_name_and_family_state
+{
+    char *name;
+    int family;
+    struct sockaddr_storage *if_addr;
+    unsigned int if_index;
+    bool found;
+    unsigned int name_matches;
+};
+
+static int aeron_ip_lookup_by_name_and_family_func(
+    void *clientd, const char *name, struct sockaddr *addr, struct sockaddr *netmask, unsigned int flags)
+{
+    if (flags & IFF_UP)
+    {
+        struct lookup_by_name_and_family_state *state = (struct lookup_by_name_and_family_state *)clientd;
+
+        if (0 == strcmp(name, state->name))
+        {
+            state->name_matches++;
+
+            if (!state->found && state->family == addr->sa_family)
+            {
+                size_t addr_len = AF_INET6 == addr->sa_family ? sizeof(struct sockaddr_in6) : sizeof(struct sockaddr_in);
+
+                memcpy(state->if_addr, addr, addr_len);
+                state->if_index = if_nametoindex(name);
+                state->found = true;
+
+                return 1;
+            }
+        }
+    }
+
+    return 0;
+}
+
 void aeron_ip_copy_port(struct sockaddr_storage *dest_addr, struct sockaddr_storage *src_addr)
 {
     if (AF_INET6 == src_addr->ss_family)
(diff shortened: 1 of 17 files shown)


aeron-client/src/main/c/util/aeron_netutil.c
```diff
@@ -529,7 +568,139 @@ void aeron_ip_copy_port(struct sockaddr_storage *dest_addr, struct sockaddr_stor
...
+
+    return 0;
+}
+
+static int aeron_find_interface_by_name_and_family(
+    int family, const char *interface_str, struct sockaddr_storage *if_addr, unsigned int *if_index)
+{
+    aeron_named_interface_t named_interface;
+
+    if (aeron_parse_named_interface(interface_str, &named_interface))
```
vyazelenko: We typically check for negative value when calling functions. So the to align with the rest of the code the check should be:
```c
if (aeron_parse_named_interface(interface_str, &named_interface) < 0)
```

aeron-driver/src/main/java/io/aeron/driver/media/InterfaceSearchAddress.java
```diff
@@ -20,38 +20,100 @@
...
 
-    InetAddress getInetAddress()
+    private InetSocketAddress resolveToAddressOfInterface(final NetworkInterface localInterface)
     {
-        return address.getAddress();
+        final InetAddress interfaceAddress = findAddressOnInterface(localInterface, address.getAddress(), subnetPrefix);
+
+        if (null == interfaceAddress)
+        {
+            throw new IllegalStateException();
```
vyazelenko: Needs more info for context.

aeron-driver/src/main/java/io/aeron/driver/media/InterfaceSearchAddress.java
```diff
@@ -20,38 +20,100 @@
...
+
+        throw new IllegalArgumentException(noMatchingInterfacesError(filteredInterfaces));
     }
 
-    int getPort()
+    private String noMatchingInterfacesError(final NetworkInterface[] filteredInterfaces) throws SocketException
     {
-        return address.getPort();
+        final StringBuilder builder = new StringBuilder()
+            .append("Unable to find multicast interface matching criteria: ")
```
vyazelenko: The `findInterfaces` method also looks for loopback interfaces which do not support multicast. So the error message should reflect that.

aeron-system-tests/src/test/java/io/aeron/ChannelInterfaceTest.java
```diff
@@ -0,0 +1,185 @@
...
+            () -> publishingClient.addPublication(channel, STREAM_ID));
+
+        assertThat(ex.getMessage(), containsString(expectedError));
+    }
+
+    private void passMessage(final Subscription subscription, final Publication publication)
+    {
+        final long payload = ThreadLocalRandom.current().nextLong();
+
+        final MutableDirectBuffer buffer = new UnsafeBuffer(new byte[SIZE_OF_LONG]);
```
vyazelenko, line 141: Can be replaced with `tryClaim` or the buffer declaration could be moved to the class field.

aeron-system-tests/src/test/java/io/aeron/ChannelInterfaceTest.java
```diff
@@ -0,0 +1,185 @@
...
+        }
+
+        throw new RuntimeException("failed to find loopback interface");
+    }
+
+    private static void assumeNotNativeDriverOnWindows()
+    {
+        assumeFalse(SystemUtil.isWindows() && TestMediaDriver.shouldRunCMediaDriver(),
+            "we don't have access to the interface names used by the native MD on Windows");
+    }
```
vyazelenko, line 184: Does it mean that this feature is not supported on Windows or simply that the test would not run?

aeron-client/src/main/c/util/aeron_netutil.c
```diff
@@ -511,6 +513,43 @@ int aeron_ip_lookup_func(
...
+};
+
+static int aeron_ip_lookup_by_name_and_family_func(
+    void *clientd, const char *name, struct sockaddr *addr, struct sockaddr *netmask, unsigned int flags)
+{
+    if (flags & IFF_UP)
+    {
+        struct lookup_by_name_and_family_state *state = (struct lookup_by_name_and_family_state *)clientd;
+
+        if (!strcmp(name, state->name))
```
vyazelenko: Can you please change the check to:
```c
if (0 == strcmp(name, state->name))
```
to make it clear that we only accept a full match. Otherwise one needs to keep [Implicit conversions](https://en.cppreference.com/w/c/language/conversion.html) rules in mind and know what `strcmp` returns to understand the logic.

This will also align it with the rest of the codebase.

aeron-system-tests/src/test/java/io/aeron/ChannelInterfaceTest.java
wojciech-adaptive: The latter.

aeron-client/src/main/c/util/aeron_netutil.c
wojciech-adaptive: Aligned.

aeron-driver/src/main/java/io/aeron/driver/media/InterfaceSearchAddress.java
wojciech-adaptive: Added.

aeron-driver/src/main/java/io/aeron/driver/media/InterfaceSearchAddress.java
wojciech-adaptive: Added.

aeron-system-tests/src/test/java/io/aeron/ChannelInterfaceTest.java
wojciech-adaptive: I don't think it's worth the additional complexity here.

aeron-client/src/main/c/util/aeron_netutil.c
wojciech-adaptive: Aligned, although not sure how that removes the need to understand `strcmp` return values or where the implicit conversion is.

aeron-client/src/main/c/util/aeron_netutil.c
```diff
@@ -529,7 +568,139 @@ void aeron_ip_copy_port(struct sockaddr_storage *dest_addr, struct sockaddr_stor
...
+        {
+            close_idx = len;
+        }
+        len++;
+    }
+
+    int name_len = close_idx - 1;
+    if (name_len <= 0 || name_len >= IF_NAMESIZE)
+    {
+        AERON_APPEND_ERR("name length %d out of bounds", name_len);
```
vyazelenko: Should be `AERON_SET_ERR` instead.

aeron-client/src/main/c/util/aeron_netutil.c
```diff
@@ -529,7 +568,139 @@ void aeron_ip_copy_port(struct sockaddr_storage *dest_addr, struct sockaddr_stor
...
+        return -1;
+    }
+
+    int port = 0;
+    int idx = close_idx + 1;
+    if (idx < len)
+    {
+        if (interface_str[idx] != ':')
+        {
+            AERON_APPEND_ERR("%s", "unexpected character after name closing brace");
```
vyazelenko: Same here.

aeron-client/src/main/c/util/aeron_netutil.c
```diff
@@ -529,7 +568,139 @@ void aeron_ip_copy_port(struct sockaddr_storage *dest_addr, struct sockaddr_stor
...
+            return -1;
+        }
+
+        idx++;
+
+        int port_len = len - idx;
+        if (port_len <= 0 || port_len > 5)
+        {
+            AERON_APPEND_ERR("port length %d out of bounds", port_len);
```
vyazelenko: And here.

aeron-client/src/main/c/util/aeron_netutil.c
```diff
@@ -529,7 +568,139 @@ void aeron_ip_copy_port(struct sockaddr_storage *dest_addr, struct sockaddr_stor
...
+            return -1;
+        }
+
+        port = aeron_udp_port_resolver(interface_str + idx, false);
+        if (port < 0)
+        {
+            return -1;
```
vyazelenko, line 623: Add `AERON_APPEND_ERR("%s", "");` to collect a stacktrace frame.

aeron-client/src/main/c/util/aeron_netutil.c
wojciech-adaptive: Good catch, thanks.

aeron-client/src/main/c/util/aeron_netutil.c
wojciech-adaptive: This doesn't really add any useful information.

PR #1906, opened 2025-12-19
mpsc scan_back_to_confirm_still_zeroed underflow UB bug

pveentjer:
It can happen that `i` can underflow and wraps around to a huge value:

```
i -= AERON_RB_ALIGNMENT
```

E.g. if `i` is 66 and AERON_RB_ALIGNMENT is 64, then after the first round `i` will be 2. After the second round `i` will be `SIZE_MAX-62`.

At that point you run into UB.

Currently is fixed by adding a guard before the decrement of `i`.

### DIFF
--- a/aeron-client/src/main/c/concurrent/aeron_mpsc_rb.c
+++ b/aeron-client/src/main/c/concurrent/aeron_mpsc_rb.c
@@ -364,6 +364,11 @@ inline static bool scan_back_to_confirm_still_zeroed(const uint8_t *buffer, size
             break;
         }
 
+        if (i < AERON_RB_ALIGNMENT)
+        {
+            break;
+        }
+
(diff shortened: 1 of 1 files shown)


pveentjer:
I have not yet been able to reproduce the error. So I'm not 100% sure if can happen with production code.

vyazelenko:
`AERON_RB_ALIGNMENT` is 8 bytes, see:
https://github.com/aeron-io/aeron/blob/c5bfa788e1bba9df21a51646c1abe4e254abf0ff/aeron-client/src/main/c/concurrent/aeron_rb.h#L52

`i` is guaranteed to be in `AERON_RB_ALIGNMENT` increments (e.g. 0, 8, 16, etc.). Therefore the situation you describe is not possible.

PR #1909, opened 2025-12-19
Fixed MappedRawLog.sliceTerms bug

pveentjer:
The constructor and the sliceTerms do not share the same condition to decide single buffer or multi buffer. So the sliceTerms can take the single buffer path while the constructor decided multi buffer and vice versa.

Concrete example.

Term length = 900 mb

Log length = 2.7 GB,

So constructor will decide multibuffer since  log lenght > 2 GB.

But sliceTerms will decide single buffer path since termLength < 1 GB.

### DIFF
--- a/aeron-driver/src/main/java/io/aeron/driver/buffer/MappedRawLog.java
+++ b/aeron-driver/src/main/java/io/aeron/driver/buffer/MappedRawLog.java
@@ -42,7 +42,6 @@ import static org.agrona.BitUtil.align;
  */
 class MappedRawLog implements RawLog
 {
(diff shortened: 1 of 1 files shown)


vyazelenko:
@pveentjer `term-length` is a power of two value ranging from `64KB` to `1GB`, i.e. it cannot be set to 900MB. See [LogBufferDescriptor#checkTermLength](https://github.com/aeron-io/aeron/blob/c5bfa788e1bba9df21a51646c1abe4e254abf0ff/aeron-client/src/main/java/io/aeron/logbuffer/LogBufferDescriptor.java#L448).

PR #1910, opened 2025-12-19
Optimized register allocation aeron_atomic64_gcc_x86_64h

pveentjer:
There was a mixture of r and q registers being used.

`q` registers are less flexible because there are only 4: rax...rdx
or their 32 bit counter parts: eax...edx

The `r` registers contain all the `q` registers, plus 12 extra; making 16 in total (32 with Intel APX)

So with more registers that can be used by the compiler, there will be less register spilling. Although I doubt it will make a significant performance different due to the cost of the atomic operations themselves, there is no point in making code more restrictive than it needs to be. Also the inconsistency isn't improving maintainability.

note:
`r`/`q` doesn't mean 32/64 bit.

Functionally there is no difference.


vyazelenko:
Good catch.

PR #1911, opened 2025-12-29
[Java] Introduce `maxPayloadLength` method to `ClientSession`

zct:
This allows service implementations to determine the maximum message size
enabling them to choose between tryClaim() for small messages (zero-copy)
and offer() for larger messages (automatic fragmentation).

### DIFF
--- a/aeron-cluster/src/main/java/io/aeron/cluster/service/ClientSession.java
+++ b/aeron-cluster/src/main/java/io/aeron/cluster/service/ClientSession.java
@@ -131,4 +131,22 @@ public interface ClientSession
      * @see BufferClaim#abort()
      */
     long tryClaim(int length, BufferClaim bufferClaim);
+
+    /**
+     * Get the maximum payload length that can be sent in a single message via {@link #tryClaim(int, BufferClaim)}.
+     * <p>
+     * This allows service implementations to determine the maximum message size enabling them to choose between
+     * {@link #tryClaim(int, BufferClaim)} for small messages (zero-copy) and {@link #offer(DirectBuffer, int, int)}
+     * for larger messages (automatic fragmentation).
+     *
+     * @return the maximum payload length in bytes.
+     */
+    int maxPayloadLength();
+
+    /**
+     * Get the maximum message length that can be sent in a single message via {@link #offer(DirectBuffer, int, int)}.
+     *
+     * @return the maximum message length in bytes.
+     */
+    int maxMessageLength();
 }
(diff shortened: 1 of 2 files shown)


zct:
@vyazelenko  Do you have any better suggestions for this use case?  Thanks!

aeron-cluster/src/main/java/io/aeron/cluster/service/ClusteredServiceAgent.java
```diff
@@ -743,6 +743,16 @@ long tryClaim(
...
     }
 
+    int maxPayloadLength(final Publication publication)
+    {
+        if (Cluster.Role.LEADER == role && null != publication)
+        {
+            return publication.maxPayloadLength() - SESSION_HEADER_LENGTH;
+        }
+
+        return headerBuffer.capacity() - SESSION_HEADER_LENGTH;
```
vyazelenko: This is wrong, i.e. the `headerBuffer.capacity()` is `65472` bytes which is far larger than a valid value for the `tryClaim` call.

aeron-cluster/src/main/java/io/aeron/cluster/service/ClusteredServiceAgent.java
```diff
@@ -743,6 +743,16 @@ long tryClaim(
         return offset;
     }
 
+    int maxPayloadLength(final Publication publication)
+    {
+        if (Cluster.Role.LEADER == role && null != publication)
+        {
+            return publication.maxPayloadLength() - SESSION_HEADER_LENGTH;
+        }
```
vyazelenko: This value is static per publication and thus should be computed only once.

aeron-cluster/src/main/java/io/aeron/cluster/service/ClusteredServiceAgent.java
zct: Ok. I'll fix it

zct:
@vyazelenko Fixed. I hope I didn't misunderstand. Add maxPayloadLength() for tryClaim() single-frame messages and maxMessageLength() for offer() fragmented messages
Both values are computed once in connect() when the responsePublication is created and subtract SESSION_HEADER_LENGTH since the cluster protocol adds a session header to every message

vyazelenko (approved):

vyazelenko:
@zct Thanks for the PR.

vyazelenko:
@zct I've declared victory too early. Your solution only work for a leader node but returns zeroes for followers. This means that it can lead to node divergence if this feature would be used, because different code branch would be taken on follower vs leader.

This feature is something we are looking to implement. However, it probably won't make the upcoming 1.50.0 release.

PR #1914, opened 2025-12-31
Fix DNS re-resolution on dynamic MDC publication change

### LINKED ISSUE #1913: Aeron dynamic MDS publication DNS re-resolution issue
## Issue

The aeron media driver does not re-resolve the DNS address of the dynamic MDC publication when:
* there are multiple publications running on the same node/IP (`node2`)
* publications use the same UDP port, but different stream ID (`1111` and `2222`)
* one of the publications changes the IP, moving to another node

Whenever `publication 1` moves to another node (`node2` -> `node3`), then the `subscription` will still get SETUP messages from the media driver running on `node2` in response to STATUS-SETUP messages. This will cause the receiver not to re-resolve the DNS address of the MDC publication because the time of last activity is continuously updated.

### Start state

Both MDC publications (`publication 1` and `publication 2`) are running on the same node, using the same UDP port but different domain names but pointing to the same IP, and different stream IDs. A subscription is continuously getting data from `publication 1` so it does not re-resolve the DNS address.

```mermaid
graph TD;

subgraph node1
    sub@{label: "subscription"}
    md1@{label: "media driver"}
end

subgraph node2
    md2@{label: "media driver"}
    pub1@{label: "publication 1"}
    pub2@{label: "publication 2"}
    style pub1 stroke:#f66
end

sub == stream ID: 1111 === md1
md1 == udp port: 5000 === md2
md2 == stream ID: 1111 === pub1
md2 -. stream ID: 2222 .- pub2
```

### Expected state

After `publisher 1` moves from a node to another, `subscription` should realize it and re-resolve the DNS address of the MDC control endpoint.

```mermaid
graph TD;

subgraph node1
    sub@{label: "subscription"}
    md1@{label: "media driver"}
end

subgraph node2
    md2@{label: "media driver"}
    pub2@{label: "publication 2"}
end

subgraph node3
    md3@{label: "media driver"}
    pub3@{label: "publication 1"}
    style pub3 stroke:#f66
end

sub == stream ID: 1111 === md1
md1 -.- md2
md1 == udp port: 5000 === md3
md2 -. stream ID: 2222 .- pub2
md3 == stream ID: 1111 === pub3
```

## Soulution

SETUP messages should not update the time of last activity field of a receiver.

leki75:
Do not update the time of last activity field of a receiver on SETUP messages.

Fixes #1913

### DIFF
--- a/aeron-driver/src/main/c/media/aeron_receive_channel_endpoint.c
+++ b/aeron-driver/src/main/c/media/aeron_receive_channel_endpoint.c
@@ -607,9 +607,8 @@ int aeron_receive_channel_endpoint_on_setup(
 {
     aeron_setup_header_t *setup_header = (aeron_setup_header_t *)buffer;
 
-    aeron_receive_destination_update_last_activity_ns(
-        destination, aeron_clock_cached_nano_time(endpoint->cached_clock));
-
+    // Do not update `destination->time_of_last_activity_ns` because it will
+    // prevent DNS re-resolution checks.
     return aeron_data_packet_dispatcher_on_setup(
         &endpoint->dispatcher, endpoint, destination, setup_header, buffer, length, addr);
 }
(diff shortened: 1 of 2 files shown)


mikeb01:
See comments on #1913.

PR #1915, opened 2026-01-07
[WIP] Fixes false sharing on mpsc and spsc queue.

pveentjer:
There is a false sharing problem in the mpsc and the spsc queues

```
typedef struct aeron_mpsc_concurrent_array_queue_stct
{
    int8_t padding[AERON_CACHE_LINE_LENGTH - sizeof(uint64_t)];
    struct
    {
        volatile uint64_t tail;
        uint64_t head_cache;
        volatile uint64_t shared_head_cache;
        int8_t padding[AERON_CACHE_LINE_LENGTH - (3 * sizeof(uint64_t))];
    }
    producer;

    struct
    {
        volatile uint64_t head;
        int8_t padding[AERON_CACHE_LINE_LENGTH - sizeof(uint64_t)];
    }
    consumer;

    size_t capacity;
    size_t mask;
    void * volatile *buffer;
}
```

For discussion sake, lets assume that tail starts at address 632.

head_cache can start at address 640.

shared_head_cache can start at 648.

Then there will be will be 64-24=40 bytes of padding.

And head will start at 648+8+40=696.

Each cacheline is 64 bytes, and therefor the head_cache (address 640) starts at the beginning of a new cacheline.

This cache line contains bytes 640-703.

So we have the head_cache and shared_head_cache (producer side) sitting on the same cache line as the (head) consumer side and as a consequence this code can suffer from false sharing.

I have run a small benchmark with and without the padding fix:

without padding fix
```
items=800000000 queue=16384 prod_cpu=2 cons_cpu=14
time=28.938893 s  throughput=27644458.057 ops/s  ns/op=36.174  sum=320000000400000000
```
with padding fix
```
items=800000000 queue=16384 prod_cpu=2 cons_cpu=14
time=1.236327 s  throughput=647078061.826 ops/s  ns/op=1.545  sum=320000000400000000
```

This is done on my AMD laptop with pretty low core to core latency. But on Intel machines or multi socket machines, the performance difference could be lot bigger.

Currently the solution is implemented with alignas; but there are some sharp corners. If you would use malloc or create an array of queues, the alignment could be ignored. So perhaps stuffing a cacheline of padding is a more reliable approach. I'm open for suggestions.

### DIFF
--- a/aeron-client/src/main/c/concurrent/aeron_mpsc_concurrent_array_queue.h
+++ b/aeron-client/src/main/c/concurrent/aeron_mpsc_concurrent_array_queue.h
@@ -22,22 +22,25 @@
 
 typedef struct aeron_mpsc_concurrent_array_queue_stct
 {
-    int8_t padding[AERON_CACHE_LINE_LENGTH - sizeof(uint64_t)];
+    int8_t padding1[AERON_CACHE_LINE_LENGTH];
+
     struct
     {
         volatile uint64_t tail;
         uint64_t head_cache;
         volatile uint64_t shared_head_cache;
-        int8_t padding[AERON_CACHE_LINE_LENGTH - (3 * sizeof(uint64_t))];
     }
     producer;
+
+    int8_t padding2[AERON_CACHE_LINE_LENGTH];
 
     struct
     {
         volatile uint64_t head;
-        int8_t padding[AERON_CACHE_LINE_LENGTH - sizeof(uint64_t)];
     }
     consumer;
+
+    int8_t padding3[AERON_CACHE_LINE_LENGTH];
 
     size_t capacity;
     size_t mask;
--- a/aeron-client/src/main/c/concurrent/aeron_spsc_concurrent_array_queue.h
+++ b/aeron-client/src/main/c/concurrent/aeron_spsc_concurrent_array_queue.h
@@ -22,21 +22,25 @@
 
 typedef struct aeron_spsc_concurrent_array_queue_stct
 {
-    int8_t padding[AERON_CACHE_LINE_LENGTH - sizeof(uint64_t)];
+
+    int8_t padding1[AERON_CACHE_LINE_LENGTH];
+
     struct
     {
         volatile uint64_t tail;
         uint64_t head_cache;
-        int8_t padding[AERON_CACHE_LINE_LENGTH - (2 * sizeof(uint64_t))];
     }
     producer;
+
+    int8_t padding2[AERON_CACHE_LINE_LENGTH];
 
     struct
     {
         volatile uint64_t head;
-        int8_t padding[AERON_CACHE_LINE_LENGTH - sizeof(uint64_t)];
     }
     consumer;
+
+    int8_t padding3[AERON_CACHE_LINE_LENGTH];
 
     size_t capacity;
     size_t mask;


vyazelenko:
@pveentjer Fails Windows build with warnings which are treated as errors:
```
D:\a\aeron\aeron\aeron-client\src\main\c\concurrent\aeron_mpsc_concurrent_array_queue.h(28,5): error C2220: the following warning is treated as an error [D:\a\aeron\aeron\cppbuild\Release\aeron-driver\src\main\c\aeron_driver_static.vcxproj]
  (compiling source file '../../../../../aeron-client/src/main/c/aeron_client.c')
  
  aeron_client.c
D:\a\aeron\aeron\aeron-client\src\main\c\concurrent\aeron_mpsc_concurrent_array_queue.h(28,5): warning C5274: behavior change: _Alignas no longer applies to the type '<unnamed-tag>' (only applies to declared data objects) [D:\a\aeron\aeron\cppbuild\Release\aeron-driver\src\main\c\aeron_driver_static.vcxproj]
  (compiling source file '../../../../../aeron-client/src/main/c/aeron_client.c')
      D:\a\aeron\aeron\aeron-client\src\main\c\concurrent\aeron_mpsc_concurrent_array_queue.h(28,5):
      to simplify migration, consider the temporary use of /Wv:18 flag with the version of the compiler with which you used to build without warnings
```

vyazelenko:
@mattgodbolt Do you see any issues with the proposed approach of using `alignas (stdalign.h)` to do padding in structs? Is there are a better alternative for `c-11/c++-11`? Or perhaps a different approach if Aeron bumps the C/C++ version?

mattgodbolt:
Hi! This appears to be C code, not C/C++. I'm afraid I don't know much about C (much less what C-11 is, I just had to google it). I don't know what version Aeron is in. I used to use Aeron at a previous job but I don't know much about how it's built.

To try and understand the question, in C++ I'd use `alignas()` like the PR does. alignas was introduced in C++11.

I have little experience with Windows compilers in the last 20 years so sadly can't help with the windows error that seems to cause.

vyazelenko:
@mattgodbolt Yes, this is a C-specific question whether we can use `alignas()` in the C code. Aeron is built with `-std=c11` (for C code) and `-std=c++11` (for C++ code).

According to the [_Alignas (since C11)(deprecated in C23), alignas (since C23)](https://en.cppreference.com/w/c/language/alignas.html) page:
* C11 defines **`_Alignas`** keyword and there is also a convenience macro `alignas`.
* C23 defines **`alignas`** keyword.

> The `_Alignas`(until C23)/`alignas`(since C23) specifier can only be used when declaring objects that are not [bit-fields](https://en.cppreference.com/w/c/language/bit_field.html), and don't have the [register](https://en.cppreference.com/w/c/language/storage_class_specifiers.html) storage class. It cannot be used in function parameter declarations, and cannot be used in a typedef.

Given it is part of the C11 standard (ISO/IEC 9899:2011) I think we could assume that it is safe to use **`alignas`** in the Aeron code base?!

Here is the [Compiler Explorer link](https://godbolt.org/#z:OYLghAFBqd5TKALEBjA9gEwKYFFMCWALugE4A0BIEAZgQDbYB2AhgLbYgDkAjF%2BTXRMiAZVQtGIHgBYBQogFUAztgAKAD24AGfgCsp5eiyagA%2BudTkVjVEQJDqzTAGF09AK5smBpwBkCTNgAcp4ARtikIADMPOQADuhKxPZMrh5eBglJdkL%2BgSFs4ZExVtg2OUwiRCykRGme3rHW2LYpVTVEecFhEdFN1bX1Gf0dXQVFfQCUVujupKicXAD0SwDUrmxxDNirAO7ESKtLSkSYaDw8AKRaAILXdwBMUQGoHjirl1HOJ4ToAHRIT64e6XJ4vN47T7fU4SAjAJgAoEgsFMV7ud5QgJEIgATzi2CUiKiwNu91x%2BJwNFWJ1I7lsH1JAHYAEL3Vbs1ZY1YsT6s24c7n0OGsJQQKIPSarVBIGqrUKXACszIVioAIryQfyOTS6UQ2RzLiz9QL2bD4SxRQA2aSS%2BhCYBSjVa9mG1VympOu6Mt2CdCe%2B5ctgsAIQSYgo3O1a%2B1bqT0C9R/UI1P6oD5RN3SB7%2ByNxUhYmgQUEPJIAL2w6ALvsmIA%2BDxVDweq1UACUAJLua21rMK5xMIvkakEMsViDqSZhqJ8m4C3P5wsNs1MEdVmug%2BuNlvtztFxW9/uC4XL9DocfZ6ccpYAKlWsyIcXcesjAtL5crx%2Brqx4DwAHMaTYujxPGtrT/S8lnDdVbi4aZ6G4BV%2BG8LgdHIdBuGcalZnmSEnj4cgiG0aDyAgFAME2bYKCoCBSK2RhIlQYAlFYOIlCQdAiAEBgiAiJRqFCAjyFCAIahxbhcME1hSBxAB5UJdBafCkP4UiOGEKSmHoETFPIHBQncYBnAkegeN4fgcCDExJC0whSHkggADcCX47B1BaB9FmQrEyn4oVQlIYTXBwfiiDzNhRP4BzSFCRJsFVbBzOAIUTAI6YaCMBiADUCGwXYpPxRDcMEYQxAkTgZDkYRlDUTQtP0WIjCS8xTEsHyeMgaZ0DiCpjIAWh%2BT5VVQC5Vm6qTgH4dAIrzHBWtDUpyhSRwmBcNwGh8Jaxh6SJYiyZIhCGRp4kSXamA2wpeiaMpbKEdpBhW4Y5quyoBk6AJujOrarGe/aDBOUZXvGc7piUTCFikGC4IQ/jUK4GNv0tbrO2AVBUwgYL3CYABrSUIHwYgyFrGJJn4BSdGmYi0HQMjaMoahqPItBXmMYBTDRzGOPoLjSB4iA%2BK08ThLCgShMkmS5NsfjlOYIg1I0/idL0gz6CMwWzKZyzkOs2yHOM5DnNcrj%2BM82CtJ8vzJIC9ziZCwWIqilRYvixLQEUlK0qUTLsty5hBcK0RxEkMrfcqjR%2BP0B5DCZkBGuaghQhm9rOpSHq%2BvTQaeGG0bxsmghpvgIHLtaBwICcb7Yj8f7NsyI6KlLw7shSU6Jgu%2Bbrq%2Bu6DuaQunr%2B/JK5GW70g757G8BmY5lBnhwa4eDyEQ5Dodh%2BHO0ZkxVlR2lMex3GSFIAnJ%2BJ5KiJIymaIiGmqJP%2Bm7NQUw7J4ABOUwjC4k5THUa12c57neeQ/nJMFv%2B0lZLyQlpTFS0t1KaQ1tgXS%2BlDLGVwqrCylttIEBsq0bWTkXKoDcobYQXkTaxzNjiC2QVrYmXILbaKDs1YBGdqTAQbsPY5Tyj7eQxUA6yCDioEONVogRwahYQwsd44oUTkIYyxxTjnHTksKSmAs4RCmo5NqD0u6LWWoPNamAR4fR2jXduVd665Aru9H6BcKg3TqIY5uj0rG6J%2Bm3LR/cXq9zMZPMeWEwaGAhrPKG3AYydjYEoa%2Bqw773z%2BM/AkRA17b3xqCQmB8XbTAxiABUWgfFcGkPwUK0gIn32/OKb80gFTSC0FoKI98tCMj8VpaG/AeIZJJoROAsBEDk1vPedilE6a0SCOwRY6g4YI2kKsJGKNWZY34NgQgO8c4GF9hw0qXD5DB2qshWq5Bdh%2BTiGFKeM857jW4FJB8XSbxUiGUvUZ4y16TOxhsU%2Bu8En7zwslFJIAHgRIeNIa0FxpCFLrFEBU99MnZPIKFdJfxrR1gKdIKIcL76WgVAqWp89uANJAE0w%2BrTkAgE6Q%2Bc%2BvSIj9I4NwS5IypT1QdOvdGUzyAzLxlNBZ7D/bLPKooHh6y9AfK2TsvZmSDn%2BK4Ccu8D5zmLwpSvaldy14PPIgTCUSTSbTCQNgFgOBIizVSfCv4PB0mMh4IyRkFT4X30%2BaCyGdT0VWExa8l2RF2koHxd02ml8%2BkDLJUEkJqZwmRJYC/Ig0zZlkHmbERZrKpArIqpy0O/DtksF2SZfZlq0XCtOWKisgTRnBNCb6qJJxZVuoiATKIRM7XKvIKq9VvQtXREtFCrQP5jXfh4FESpBrZDGzBYclC1rGnluglPB4OS0laD%2BFoHgloDWlOBYirQ99ZA9vqQO8c5BUnpMyVEFNRyuBKpaY6kAJA3DnxqAxa1p6lCqGMGUEQrFdj5SUkW0gJLBnDMRsjW5G86UMrmWcMNLKSqRvZWs2NsQ2BCBYBgOJpAAASFpmTYGYDcB86AvaBDOBQ8DrAoPBtIGhwWWHIPoCqIQB9n1SN9tOAEK9gR6C3vQPewWR76CtjOErOybBMC7BqDQTAnhOpft5Qmw%2BxtBVWq4Oxtg3UuM8b45sT9tLJTkvfRMr9W9cN7wHHK2ie8y3NNXakusFrUU7oxVi%2B1OLybMZPaQM9u7yAXpozeu9ZGiXPo9VwZToypUKc3kGxlob2VLKA9wqqsbw6EZw4yuDSgENIZQ2hmZBGINRZ3vhzDKXiNUbIz8AIlHSNOboy5pjx4WNsfoBxmTpBeP8cmXhGy2B5kZPjYmhhont29okxVqTVWavyZpZvCVy8qW%2BaxrEjTzytNPoVXpt5la1UauoO8ozXaOvLv7fpoda3rWbcodxBa0ggA%3D%3D%3D) that shows **`alignas`** working as expected when compiled with gcc, clang and msvc. The oldest compiler versions where this example works:
* `x86-64 gcc 4.8.1`
* `x86-64 clang 4.0.0`
* `x64 msvc v19.28 VS 16.8`

P.S. MSVC is unhappy when `alignas` is applied at the `struct`-level. Moving it to the first field in the struct fixes it.

CC @mjpt777 @mikeb01

aeron-client/src/main/c/concurrent/aeron_mpsc_concurrent_array_queue.h
```diff
@@ -16,29 +16,30 @@
...
 
+#include <stdalign.h>
 #include "util/aeron_bitutil.h"
 #include "aeron_atomic.h"
 #include "aeron_concurrent_array_queue.h"
 
 typedef struct aeron_mpsc_concurrent_array_queue_stct
 {
-    int8_t padding[AERON_CACHE_LINE_LENGTH - sizeof(uint64_t)];
+    alignas(AERON_CACHE_LINE_LENGTH)
```
vyazelenko: Should be applied to the `tail` field declaration instead, because MSVC.

aeron-client/src/main/c/concurrent/aeron_mpsc_concurrent_array_queue.h
```diff
@@ -16,29 +16,30 @@
...
     struct
     {
         volatile uint64_t tail;
         uint64_t head_cache;
         volatile uint64_t shared_head_cache;
-        int8_t padding[AERON_CACHE_LINE_LENGTH - (3 * sizeof(uint64_t))];
     }
     producer;
 
+    alignas(AERON_CACHE_LINE_LENGTH)
```
vyazelenko: Should be applied to the `head` field declaration instead, because MSVC.

aeron-client/src/main/c/concurrent/aeron_mpsc_concurrent_array_queue.h
```diff
@@ -16,29 +16,30 @@
...
     struct
     {
         volatile uint64_t head;
-        int8_t padding[AERON_CACHE_LINE_LENGTH - sizeof(uint64_t)];
     }
     consumer;
 
+    alignas(AERON_CACHE_LINE_LENGTH)
     size_t capacity;
```
vyazelenko: Reformat to one line: `alignas(AERON_CACHE_LINE_LENGTH) size_t capacity;`

aeron-client/src/main/c/concurrent/aeron_spsc_concurrent_array_queue.h
```diff
@@ -16,28 +16,30 @@
...
+#include <stdalign.h>
 #include "util/aeron_bitutil.h"
 #include "aeron_atomic.h"
 #include "aeron_concurrent_array_queue.h"
 
 typedef struct aeron_spsc_concurrent_array_queue_stct
 {
-    int8_t padding[AERON_CACHE_LINE_LENGTH - sizeof(uint64_t)];
+
+    alignas(AERON_CACHE_LINE_LENGTH)
```
vyazelenko: Should be applied to the `tail` field declaration instead, because MSVC.

aeron-client/src/main/c/concurrent/aeron_spsc_concurrent_array_queue.h
```diff
@@ -16,28 +16,30 @@
...
     struct
     {
         volatile uint64_t tail;
         uint64_t head_cache;
-        int8_t padding[AERON_CACHE_LINE_LENGTH - (2 * sizeof(uint64_t))];
     }
     producer;
 
+    alignas(AERON_CACHE_LINE_LENGTH)
```
vyazelenko: Should be applied to the `head` field declaration instead, because MSVC.

aeron-client/src/main/c/concurrent/aeron_spsc_concurrent_array_queue.h
```diff
@@ -16,28 +16,30 @@
...
     struct
     {
         volatile uint64_t head;
-        int8_t padding[AERON_CACHE_LINE_LENGTH - sizeof(uint64_t)];
     }
     consumer;
 
+    alignas(AERON_CACHE_LINE_LENGTH)
     size_t capacity;
```
vyazelenko: Reformat to one line: `alignas(AERON_CACHE_LINE_LENGTH) size_t capacity;`

aeron-client/src/main/c/benchmark.c
vyazelenko, line 1: Delete this file or move to samples if it is relevant after the fixes are applied.

vyazelenko (changes requested):

PR #1917, opened 2026-01-07
Rollback the consensus module leader's uncommitted state

marc-adaptive:
This PR is to rollback the consensus module leader's uncommitted state in the case of an unexpected election.

In the cases of using control toggle for suspend, snapshot, shutdown, and resume - the consensus module leader updates its ConsensusModuleAgent#state. The corresponding cluster action added to the log is initially uncommitted. If an unexpected election occurs, the leader will need to rollback this uncommitted state to its committed value.

The leader's consensus module will track its previous states and log position corresponding to cluster action, while cluster action is uncommitted. For now, avoided using aeron-test-support TestCluster as using consensus module invoker for ease of testing and wanted to avoid making changes to TestCluster.

Feel free to close if I'm wrong, misunderstanding something, or there's a better approach.

### DIFF
--- a/aeron-system-tests/src/test/java/io/aeron/cluster/ClusterUncommittedStateTest.java
+++ b/aeron-system-tests/src/test/java/io/aeron/cluster/ClusterUncommittedStateTest.java
@@ -102,7 +102,7 @@ public class ClusterUncommittedStateTest
         if (!hasService)
         {
             clusterBuilder
-                .withExtensionSuppler(TestCounterExtension::new)
+                .withExtensionSuppler(TestNode.TestConsensusModuleExtension::new)
                 .withServiceSupplier(value -> new TestNode.TestService[0]);
         }
         cluster = clusterBuilder.start();
@@ -139,7 +139,7 @@ public class ClusterUncommittedStateTest
         if (!hasService)
         {
             clusterBuilder
-                .withExtensionSuppler(TestCounterExtension::new)
+                .withExtensionSuppler(TestNode.TestConsensusModuleExtension::new)
                 .withServiceSupplier(value -> new TestNode.TestService[0]);
         }
         cluster = clusterBuilder.start();
@@ -189,7 +189,7 @@ public class ClusterUncommittedStateTest
         if (!hasService)
         {
             clusterBuilder
-                .withExtensionSuppler(TestCounterExtension::new)
+                .withExtensionSuppler(TestNode.TestConsensusModuleExtension::new)
                 .withServiceSupplier(value -> new TestNode.TestService[0]);
         }
         cluster = clusterBuilder.start();


mjpt777:
The approach seems reasonable. However the use of IP Tables for testing feels heavyweight and Linux dependant.

vyazelenko:
> However the use of IP Tables for testing feels heavyweight and Linux dependant.

We use this approach for testing Cluster in other places. See [ClusterNetworkPartitionTest](https://github.com/aeron-io/aeron/blob/master/aeron-system-tests/src/test/java/io/aeron/cluster/ClusterNetworkPartitionTest.java).

aeron-system-tests/src/test/java/io/aeron/cluster/ClusterUncommittedStateTest.java
```diff
@@ -0,0 +1,184 @@
...
+        final List<String> leaderHostname = List.of(HOSTNAMES.get(firstLeader.memberId()));
+        final List<String> followerHostnames = new ArrayList<>(HOSTNAMES);
+        followerHostnames.remove(firstLeader.memberId());
+
+        IpTables.makeSymmetricNetworkPartition(CHAIN_NAME, leaderHostname, followerHostnames);
+
+        cluster.suspendCluster(firstLeader);
+        Tests.await(() -> ConsensusModule.State.SUSPENDED == firstLeader.moduleState());
+
+        Tests.await(() -> null != cluster.findLeader(firstLeader.memberId()));
```
vyazelenko, line 121: Could replaced with `cluser.awaitLeader(firstLeader.memberId()`.

mjpt777:
Loss and other channel behaviour can be introduced in an OS independent way. For example: https://github.com/aeron-io/aeron/blob/master/aeron-system-tests/src/test/java/io/aeron/GapFillLossTest.java.

aeron-system-tests/src/test/java/io/aeron/cluster/ClusterUncommittedStateTest.java
```diff
@@ -0,0 +1,184 @@
...
+
+        IpTables.flushChain(CHAIN_NAME);
+        Tests.await(() -> ConsensusModule.State.ACTIVE == firstLeader.moduleState());
```
vyazelenko, line 124: You should wait for election completion first with `awaitElectionClosed(firstLeader)` and then assert the CM state.

marc-adaptive:
For testing needed a way to generate uncommitted data deterministically and then have 2 followers start new leadership term. I could think of:

1. Use system test cluster allowing consensus module invoker. Think more decisions beyond the scope of this PR needed to be made there, for example, calls to ClusterTool without a way of polling consensus module.
2. Using private test class to pull out required methods from TestCluster seems like may add long term maintenance cost. Used this approach initially.
3. Using loss generator. Seems like it would require complex logic - originally drop specific frames and from test thread set a volatile to allow those frames. Is this the way to go?
4. Using ip tables.

aeron-system-tests/src/test/java/io/aeron/cluster/ClusterUncommittedStateTest.java
```diff
@@ -0,0 +1,184 @@
...
+        {
+            boolean allNodesSuspended = true;
+            for (int i = 0; i < cluster.memberCount(); ++i)
+            {
+                if (ConsensusModule.State.SUSPENDED != cluster.node(i).moduleState())
+                {
+                    allNodesSuspended = false;
+                }
+            }
+            return allNodesSuspended;
```
vyazelenko: Can be simplified with returning `false` if state is not `SUSPENDED`. It is also faster.

aeron-system-tests/src/test/java/io/aeron/cluster/ClusterUncommittedStateTest.java
```diff
@@ -0,0 +1,184 @@
...
+        Tests.await(() ->
+        {
+            boolean allNodesSuspended = true;
+            for (int i = 0; i < cluster.memberCount(); ++i)
+            {
+                if (ConsensusModule.State.SUSPENDED != cluster.node(i).moduleState())
+                {
+                    allNodesSuspended = false;
+                }
+            }
```
vyazelenko: Can be simplified with returning `false` if state is not `SUSPENDED`. It is also faster.

vyazelenko:
@marc-adaptive I suggest to stick to existing ip tables tests for this release. Perhaps move them to the `ClusterNetworkPartitionTest` so that we have a single place to improve. We can then iterate on the approach for the next major release.

aeron-system-tests/src/test/java/io/aeron/cluster/ClusterUncommittedStateTest.java
marc-adaptive: Thanks

aeron-system-tests/src/test/java/io/aeron/cluster/ClusterUncommittedStateTest.java
marc-adaptive: Thanks

aeron-system-tests/src/test/java/io/aeron/cluster/ClusterUncommittedStateTest.java
marc-adaptive: The reason awaiting for active is to use least amount of awaits + assertions. This test only cares about recovering leader's state. By looking at the state - only need to assert was suspended and eventually became active again, simulating a truncation. Waiting for election closed would maybe get confused by prior election closed and reading previous counter value. Likely would have to also assert leadershipTermId or election count has increased too. Is this okay to leave?

aeron-system-tests/src/test/java/io/aeron/cluster/ClusterUncommittedStateTest.java
marc-adaptive: Thanks you're right. Initially used this assert since cluster#awaitLeader checks for all nodes to have a closed election. Given this test, the node is suspended and won't enter a new election if network partition. Left to keep aligned with other tests; the other tests need this await because active node will detect inactive follower quorum, start a new election, and await would timeout. Okay to keep aligned with other tests?

marc-adaptive:
Thanks @vyazelenko. Started looking at potentially using loss generator and maybe unit testing these scenarios. Hoping to have something soon.

aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModuleAgent.java
```diff
@@ -2661,6 +2662,8 @@ private int checkClusterControlToggle(final long nowNs)
                     final long timestamp = clusterClock.time();
                     if (appendAction(ClusterAction.SUSPEND, timestamp, CLUSTER_ACTION_FLAGS_DEFAULT))
                     {
+                        uncommittedState.offerLong(logPublisher.position());
+                        uncommittedState.offerLong(ConsensusModule.State.ACTIVE.code());
```
mikeb01: Perhaps this should store the `state` variable instead of the constant so that it is clear that we are storing the current state prior to a transition rather that a specific state.

aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModuleAgent.java
```diff
@@ -3521,6 +3536,23 @@ private void restoreUncommittedEntries(final long commitPosition)
                 addSession(session);
             }
         }
+
+        while (uncommittedState.peekLong() <= commitPosition)
```
mikeb01: Just to verify my understanding here we are using `<=` here as this queue will always be storing state prior to the one we actually want to restore to.  So we then either restore to the next state present in the queue or leave it in the current state?

marc-adaptive: We're storing the log position of the cluster action and the corresponding previous state. <= is used to verify log position corresponding to the cluster action was committed, then we don't need to remember the corresponding previous state.

image

aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModuleAgent.java
marc-adaptive: That makes sense. Thanks

marc-adaptive:
I've pushed a change to use java sender/receiver endpoint suppliers to toggle loss for now. Have done so with other PRs as well. These tests are only enabled using java media driver for the time being.

aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModuleAgent.java
```diff
@@ -2681,6 +2682,8 @@ private int checkClusterControlToggle(final long nowNs)
                     final long timestamp = clusterClock.time();
                     if (appendAction(ClusterAction.SUSPEND, timestamp, CLUSTER_ACTION_FLAGS_DEFAULT))
                     {
+                        uncommittedState.offerLong(logPublisher.position());
+                        uncommittedState.offerLong(state.code());
```
mikeb01: These two lines are repeated a few times, I think it should be pulled into a method with a descriptive name.  Also because then idea the value stored as a pair of entries can be encapsulated.  We should also have a single method for popping those values off again (as we do that in two places).  Probably something like:
```
offerUncommittedStateAndPosition(long logPublisherPosition, State state)
```

PR #1918, opened 2026-01-11
Rollback the consensus module leader's next committed session id

marc-adaptive:
This PR is to fix ConsensusModuleAgent#nextCommittedSessionId.

Previously the leader would bump nextCommittedSessionId once message was appended to log, which doesn't guarantee the session open is committed. In the case session opens are not committed and the log is truncated, a leader becoming a follower could have an incorrect nextCommittedSessionId resulting in an inconsistent snapshot across nodes.

Feel free to close if I'm wrong, misunderstanding something, or there's a better approach.

### DIFF
--- a/aeron-cluster/src/main/java/io/aeron/cluster/SessionManager.java
+++ b/aeron-cluster/src/main/java/io/aeron/cluster/SessionManager.java
@@ -492,7 +492,10 @@ class SessionManager
         if (clusterSessionId >= nextSessionId)
         {
             nextSessionId = clusterSessionId + 1;
-            nextCommittedSessionId = nextSessionId;
+        }
+        if (clusterSessionId >= nextCommittedSessionId)
+        {
+            nextCommittedSessionId = clusterSessionId + 1;
         }
 
         if (null != consensusModuleExtension)
@@ -1090,7 +1093,6 @@ class SessionManager
             nextCommittedSessionId = Math.max(nextCommittedSessionId, uncommittedNextCommittedSessionIds.pollLong());
         }
         uncommittedNextCommittedSessionIds.clear();
-        nextSessionId = nextCommittedSessionId;
     }
 
     void updateTimeOfLastActivity()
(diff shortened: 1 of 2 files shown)


aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModuleAgent.java
```diff
@@ -686,12 +687,6 @@ public void onLoadClusterSession(
         session.loadSnapshotState(correlationId, openedPosition, timeOfLastActivity, closeReason);
 
         addSession(session);
-
```
mikeb01: @marc-adaptive Could you run me though the logic of remove this?  It is because the consensus module state values are now guaranteed to be higher than the ids of the stored cluster sessions?

mikeb01: Is it possible that after loading a snapshot stored from an older cluster into a cluster of this version that the nextSessionId/nextCommitedSessionId could be incorrect?

marc-adaptive: I don't see how we could end up in that code path as snapshotted/bootstrapped session id value should always be >= values in sessions. But I may be wrong and it doesn't hurt to have this check so I will reverted this. I didn't consider case of older clusters. Thanks for catching.

Considering the cases
- leader replay, live follower, follower replay, follower catchup - nextSessionId and nextCommittedSessionId will be the same. Loaded from snapshot/bootstrapper, #onReplaySessionOpen can increment, and value is snapshotted
- live leader - nextSessionId and nextCommittedSessionId will diverge during the leadership term. Both loaded from snapshot/bootstrapper. #onSessionConnect will increment nextSessionId. #sweepUncommittedEntriesTo and #restoreUncommittedEntries can increment nextCommittedSessionId. nextCommittedSessionId is snapshotted. And on new election nextSessionId is brought down to nextCommittedSessionId.

marc-adaptive:
This approach rolls back nextSessionId to nextCommittedSessionId when leadership term is over. Am unsure if this could be problematic if an authenticator assumes unique session ids. The other approach is to not roll back nextSessionId and update nextSessionId and nextCommittedSessionId independently in onReplaySessionOpen.

PR #1919, opened 2026-01-11
Fix Consensus Module Snapshotting with 0 services

marc-adaptive:
This PR is to fix consensus module snapshotting when there are 0 services.

It revives these 2 PRs that were reverted for future work
- https://github.com/aeron-io/aeron/pull/1865
- https://github.com/aeron-io/aeron/pull/1866

In the case of 0 services, currently the leader will take a snapshot once it detects snapshot toggle without waiting for uncommitted data to become committed. This could result in an inconsistent consensus module snapshotted state in addition to an inconsistent extension snapshotted state across nodes. We don't have to worry about the case of 1+ services, since the consensus module will wait for service acks, which is bounded by the commit position since the services cannot read cluster action in log until it is committed.

In the case of 0 services, the consensus module will now wait for uncommitted data to become committed. This PR builds on top of https://github.com/aeron-io/aeron/pull/1917 (allowing the consensus module to rollback state in the case of an unexpected election). If the consensus module is unable to rollback it could remain stuck in snapshotting state, even if the snapshot cluster action was truncated.

Feel free to close if wrong, misunderstanding something, or there's a better approach.

### DIFF
--- a/aeron-system-tests/src/test/java/io/aeron/cluster/ClusterUncommittedStateTest.java
+++ b/aeron-system-tests/src/test/java/io/aeron/cluster/ClusterUncommittedStateTest.java
@@ -18,8 +18,19 @@ package io.aeron.cluster;
 import io.aeron.Counter;
 import io.aeron.ExclusivePublication;
 import io.aeron.Image;
+import io.aeron.cluster.client.AeronCluster;
 import io.aeron.cluster.codecs.MessageHeaderDecoder;
 import io.aeron.cluster.codecs.MessageHeaderEncoder;
+import io.aeron.driver.DataPacketDispatcher;
+import io.aeron.driver.MediaDriver;
+import io.aeron.driver.ReceiveChannelEndpointSupplier;
+import io.aeron.driver.SendChannelEndpointSupplier;
+import io.aeron.driver.ext.DebugReceiveChannelEndpoint;
+import io.aeron.driver.ext.DebugSendChannelEndpoint;
+import io.aeron.driver.ext.LossGenerator;
+import io.aeron.driver.media.ReceiveChannelEndpoint;
+import io.aeron.driver.media.SendChannelEndpoint;
+import io.aeron.driver.media.UdpChannel;
 import io.aeron.logbuffer.ControlledFragmentHandler;
 import io.aeron.logbuffer.FrameDescriptor;
 import io.aeron.logbuffer.Header;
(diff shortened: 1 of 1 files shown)


aeron-system-tests/src/test/java/io/aeron/cluster/ClusterUncommittedStateTest.java
```diff
@@ -0,0 +1,218 @@
+/*
+ * Copyright 2014-2024 Real Logic Limited.
```
mjpt777: Copyright date incorrect.

aeron-cluster/src/main/java/io/aeron/cluster/ClusterTermination.java
```diff
@@ -22,12 +22,11 @@
 class ClusterTermination
 {
     private long deadlineNs;
-    private boolean haveServicesTerminated;
+    private boolean isTerminationReady = false;
```
mjpt777: It may make more sense to now invert the logic and say `isAwaitingServices`?

aeron-system-tests/src/test/java/io/aeron/cluster/ClusterUncommittedStateTest.java
```diff
@@ -0,0 +1,218 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+import static io.aeron.test.cluster.TestCluster.aCluster;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+@ExtendWith({ EventLogExtension.class, InterruptingTestCallback.class })
+@EnabledOnOs(OS.LINUX)
```
mjpt777: I think the implementation looks reasonable. However it feels heavyweight to test using IP Tables. Can this be tested by a simpler means that works on all operating systems?

aeron-system-tests/src/test/java/io/aeron/cluster/ClusterUncommittedStateTest.java
marc-adaptive: Thanks

aeron-cluster/src/main/java/io/aeron/cluster/ClusterTermination.java
marc-adaptive: Thanks

PR #1920, opened 2026-01-12
[Java] Get rid of excessive allocations in AeronCluster.AsyncConnect.awaitPublicationConnected

zyulyaev:
Currently AeronCluster makes allocations on each invocation of AeronCluster.AsyncConnect#poll while in `AWAIT_PUBLICATION_CONNECTED`. This creates massive GC pressure when Aeron Cluster does not respond in time. This PR fixes this by caching `egressSubscription.tryResolveChannelEndpointPort()` result.

One thing I am not 100% sure - whether `egressSubscription.tryResolveChannelEndpointPort()` call should be extracted into a separate step like `AWAIT_EGRESS_CHANNEL_ENDPOINT_RESOLVED`. Let me know.


vyazelenko:
This is already solved by caching in the [Subscription#tryResolveChannelEndpointPort](https://github.com/aeron-io/aeron/blob/3593d8b34742b43dcc3f14d64e1c3ea3e9c830c9/aeron-client/src/main/java/io/aeron/Subscription.java#L540-L565). The fix will be released this week as part of `1.50.0` version.

PR #1933, opened 2026-01-25
[Cluster] Add inter-cluster bridge module for bidirectional cluter-cluster messaging

anirudhpundhir:
Introduces aeron-cluster-bridge, a deterministic messaging bridge between independent Aeron Clusters using UDP publications, Archive replay, and ReplayMerge for crash-recovery catch-up with application-level idempotency.

### DIFF
--- a/aeron-cluster-bridge/src/main/java/io/aeron/cluster/bridge/BridgeCheckpoint.java
+++ b/aeron-cluster-bridge/src/main/java/io/aeron/cluster/bridge/BridgeCheckpoint.java
@@ -1,0 +1,184 @@
+/*
+ * Copyright 2014-2025 Real Logic Limited.
+ *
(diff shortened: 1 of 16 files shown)


nickelaway:
Hi Anirudh,

Given this PR is adding a new feature that uses Aeron, rather than changing existing functionality of Aeron, this looks like something that could live in its own repo rather than being incorporated into Aeron itself. Every change we accept into Aeron comes with extra maintenance and support burden, and this is a fairly substantial new feature. There is also another option for solving the cluster-to-cluster communication problem using a sequencer-based architecture (see https://aeron.io/aeron-sequencer/).

For a change like this, I would suggest publishing it as its own open source project. If you're looking for support in building or maintaining a new feature like this, you can also contact us for a commercial agreement. See https://aeron.io/contact/

PR #1943, opened 2026-02-15
[C] Fix data corruption resulting from aeron_append_block ignoring term_offset

klittlepage:
`aeron_append_block` ignores `term_offset` and writes the header and payload to the start of the term buffer instead of `term_buffer->addr + term_offset`. This results in silent data corruption, and is inconsistent with the java client implementation, which does not have this issue.


mikeb01:
This looks valid at first glance.  I am adding additional tests to cover this case before merging.

mikeb01:
We've added some tests in this area and manually pulled in this change.

PR #1956, opened 2026-03-09
add ReplayMerge sample

Nadia-Adaptive:


aeron-samples/src/main/java/io/aeron/samples/archive/ReplayMergeSubscriber.java
```diff
@@ -0,0 +1,144 @@
...
+                        descriptor.recordingId(),
+                        0))
+                {
+                    final AtomicBoolean isLive = new AtomicBoolean(false);
+
+                    while (running.get())
+                    {
+                        if (replayMerge.hasFailed())
+                        {
+                            System.out.println("ReplayMerge has failed, " + replayMerge);
```
vyazelenko: This error message should be added to the exception and not logged to stdout.

aeron-samples/src/main/java/io/aeron/samples/archive/ReplayMergeSubscriber.java
```diff
@@ -0,0 +1,144 @@
...
+                            throw new RuntimeException();
```
vyazelenko: `IllegalStateException` is more appropriate in this case.

aeron-samples/src/main/java/io/aeron/samples/archive/ReplayMergeSubscriber.java
```diff
@@ -0,0 +1,144 @@
...
+
+                        if (replayMerge.isMerged() && !isLive.get())
+                        {
+                            System.out.println("===========");
+                            System.out.println("ReplayMerge has joined live stream.");
+                            System.out.println("===========");
+                            isLive.set(true);
+                        }
+
+                        replayMerge.poll(
```
vyazelenko: The return value from `ReplayMerge#poll` is ignored. At the very least it should be passed on to `IdleStrategy#idle` method. Something like:
```java
// Before the loop
final IdleStrategy idleStrategy = ...;


// In the loop
idleStrategy.idle(replayMerge.poll(...));
```

aeron-samples/src/main/java/io/aeron/samples/archive/ReplayMergeSubscriber.java
```diff
@@ -0,0 +1,144 @@
...
+                            (final DirectBuffer buffer, final int offset, final int length, final Header header) ->
```
vyazelenko: This is a _capturing_ lambda expression which is allocated on every call to `replayMerge.poll`. We want to teach Aeron users how to use it best and thus should show the correct patter, i.e. reuse lambda across all invocation:

```java
// Before the loop
final FragmentHandler fragmentHandler = new FragmentAssembler(<my lambda expression>);

// in the loop
idleStrategy.idle(replayMerge.poll(fragmentHandler));
```

aeron-samples/src/main/java/io/aeron/samples/archive/ReplayMergeSubscriber.java
```diff
@@ -0,0 +1,144 @@
...
+                ctx.aeronDirectoryName(driver.aeronDirectoryName());
+            }
+
+            final AeronArchive.Context aeronArchiveCtx = new AeronArchive.Context();
+
+            // Create Aeron and AeronArchive instances using the configured Context.
+            try (Aeron aeron = Aeron.connect(ctx);
+                AeronArchive aeronArchive = AeronArchive.connect(aeronArchiveCtx.aeron(aeron)))
+            {
+                final RecordingDescriptor descriptor = getLastDescriptor(aeronArchive);
```
vyazelenko, line 92: What if no recording found or recording hasn't started yet?

aeron-samples/src/main/java/io/aeron/samples/archive/ReplayMergeSubscriber.java
```diff
@@ -0,0 +1,144 @@
...
+            }
+        }
+    }
+
+    private static RecordingDescriptor getLastDescriptor(final AeronArchive aeronArchive)
+    {
+        final AtomicReference<RecordingDescriptor> descriptorRef = new AtomicReference<>(null);
+        final RecordingDescriptorCollector collector = new RecordingDescriptorCollector(Integer.MAX_VALUE);
+
+        aeronArchive.listRecordings(0, collector.poolSize(), collector.reset());
```
vyazelenko: This finds any existing recordings but what you need is to find recordings for a specific stream instead that `RecordedBasicPublisher` has created.

Use `AeronArchive#listRecordingsForUri` instead and pass both channel and streamId to search by.

PR #1958, opened 2026-03-10
[C] Improve loss detection and recovery

zyulyaev:
Currently Aeron detects and recovers a single loss at a time. This works well enough for low-latency low-loss networks. However in case of long-range WAN connections and loss spikes it takes a lot of time to recover multiple losses. 
In our system we have a connection with RTT ~250ms, and lately we've been experiencing spikes to 5-10% packet loss. Such scenario completely freezes Aeron connections for several seconds.

This PR makes Aeron detects and request retransmission for all gaps instead of the closest one.


vyazelenko:
@zyulyaev Aeron was designed for low loss networks and can comfortably operate with loss way below one percent (e.g. 0.1%). Other protocols such as TCP are better suited for high loss congested networks. Improving loss recovery, while addressing some of the issues, won't solve the fundamental problems. Therefore we believe a better solution will be to add a TCP transport to Aeron which we have plans for on our future roadmap.

PR #1963, opened 2026-03-17
Stop Consensus Module Agent on Clustered Service start failure

marc-adaptive:


aeron-cluster/src/main/java/io/aeron/cluster/service/ClusteredServiceAgent.java
```diff
@@ -768,28 +768,47 @@ private void recoverState(final CountersReader counters)
...
-        while (!consensusModuleProxy.ack(logPosition, clusterTime, id, aeron.clientId(), serviceId))
+        final long relevantId = (null == exception) ? aeron.clientId() : NULL_VALUE;
+        while (!consensusModuleProxy.ack(logPosition, clusterTime, id, relevantId, serviceId))
         {
             idle();
         }
+
+        if (null != exception)
+        {
+            final String message = "failed to load for service=" + ctx.serviceId() +
```
vyazelenko, line 805: `failed to load for service` a word is missing (i.e. `state` or `snapshot`)?

marc-adaptive: It can fail for variety of reasons - find recordingId from recovery counter, archive/aeron client calls, snapshot, or potential service onStart logic. Unsure of the best wording - will change to better wording in next day or so

PR #1990, opened 2026-04-08
[Driver/C] Fixes aeron_receive_channel_endpoint_create memory leak

pveentjer:
The aeron_receive_channel_endpoint_create allocates memory for the _endpoint. And then it does a step by step initialization.

In almost all steps, if the initialization step fails, the _endpoint was not deleted.

I have added an error handler at the end and made sure that on all error paths, the error handler is called.

### DIFF
--- a/aeron-driver/src/main/c/media/aeron_receive_channel_endpoint.c
+++ b/aeron-driver/src/main/c/media/aeron_receive_channel_endpoint.c
@@ -67,35 +67,35 @@ int aeron_receive_channel_endpoint_create(
     if (aeron_alloc((void **)&_endpoint, sizeof(aeron_receive_channel_endpoint_t)) < 0)
     {
         AERON_APPEND_ERR("%s", "could not allocate receive_channel_endpoint");
-        return -1;
+        goto error;
     }
 
     if (aeron_data_packet_dispatcher_init(
         &_endpoint->dispatcher, context->conductor_proxy, context->receiver_proxy->receiver) < 0)
     {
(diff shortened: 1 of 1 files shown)


mikeb01:
I think we need some tests around these changes as some of the case could result in segfaults from the `_delete` function.

aeron-driver/src/main/c/media/aeron_receive_channel_endpoint.c
```diff
@@ -133,14 +132,21 @@ int aeron_receive_channel_endpoint_create(
...
+            goto error;
         }
     }
 
     // Only take ownership on successful construction.
     _endpoint->conductor_fields.udp_channel = channel;
     *endpoint = _endpoint;
     return 0;
+
+    error:
```
vyazelenko, line 144: Formatting, i.e. `error:` should have no offset.

aeron-driver/src/main/c/media/aeron_receive_channel_endpoint.c
```diff
@@ -67,35 +67,35 @@ int aeron_receive_channel_endpoint_create(
     if (aeron_alloc((void **)&_endpoint, sizeof(aeron_receive_channel_endpoint_t)) < 0)
     {
         AERON_APPEND_ERR("%s", "could not allocate receive_channel_endpoint");
-        return -1;
+        goto error;
```
vyazelenko, line 70: Should `return -1`.

vyazelenko (changes requested):

PR #1994, opened 2026-04-09
Config annotation updates

nbradac:


.github/workflows/config-doc.yml
```diff
@@ -0,0 +1,44 @@
...
+  contents: read
+
+jobs:
+  config-annotations-build:
+    name: Config Annotations
+    runs-on: ubuntu-24.04
+    timeout-minutes: 30
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v5
```
vyazelenko: Outdated Github actions should use instead:
* `actions/checkout@v6`
* `gradle/actions/setup-gradle@v6`
* `actions/upload-artifact@v7`

.github/workflows/config-doc.yml
```diff
@@ -0,0 +1,44 @@
...
+        with:
+          distribution: 'zulu'
+          java-version: ${{ env.JAVA_VERSION }}
+      - name: Setup BUILD_JAVA_HOME & BUILD_JAVA_VERSION
+        run: |
+          java -Xinternalversion
+          echo "BUILD_JAVA_HOME=${JAVA_HOME}" >> $GITHUB_ENV
+          echo "BUILD_JAVA_VERSION=${{ env.JAVA_VERSION }}" >> $GITHUB_ENV
+      - name: Setup Gradle
+        uses: gradle/actions/setup-gradle@v5
```
vyazelenko: Outdated Github actions (see above).

.github/workflows/config-doc.yml
```diff
@@ -0,0 +1,44 @@
...
+      - name: Build and validate config annotations
+        run: ./gradlew generateCombinedConfigDoc --console=plain
+      - name: Upload generated docs
+        uses: actions/upload-artifact@v5
```
vyazelenko: Outdated Github actions (see above).

.github/workflows/config-doc.yml
```diff
@@ -0,0 +1,44 @@
+name: Config Documentation
```
vyazelenko, line 1: What is the idea behind this workflow? Who will be checking/consuming generated artifacts?

aeron-annotations/src/main/java/io/aeron/utility/JavadocCleaner.java
```diff
@@ -0,0 +1,166 @@
...
+ * limitations under the License.
+ */
+package io.aeron.utility;
+
+import java.util.regex.Pattern;
+
+/**
+ * Converts Javadoc markup (inline tags, HTML) in description strings to Markdown equivalents,
+ * and strips block Javadoc tags such as {@literal @}return, {@literal @}param, etc.
+ */
```
vyazelenko, line 23: Using regex to parse HTML? Why?

* [Regex Can't Parse HTML](https://www.linkedin.com/posts/laurie-kirk_its-impossible-to-regex-html-it-also-activity-7415808273028689923-KvAP)
* [RegEx match open tags except XHTML self-contained tags](https://stackoverflow.com/questions/1732348/regex-match-open-tags-except-xhtml-self-contained-tags)

build.gradle
```diff
@@ -1425,6 +1458,75 @@ tasks.register('uploadArtifactsToCentralPortal', io.aeron.build.SonatypeCentralP
...
+    final def docFileProviders = subprojectNames.collect {
+        project(":${it}").layout.buildDirectory.file('generated-doc/out.md')
+    }
+    final def counterDocFileProvider = project(':aeron-client').layout.buildDirectory.file('generated-doc/counters.md')
+
+    inputs.files(docFileProviders)
+    inputs.files(counterDocFileProvider)
+
+    final def outputFile = layout.buildDirectory.file('generated-doc/config.md')
+    final def counterOutputFile = layout.buildDirectory.file('generated-doc/counters.md')
```
vyazelenko: We have settled on using AsciiDoc for all documentation across all projects.

vyazelenko (changes requested):

vyazelenko:
@nbradac Have you looked at the https://github.com/asciidoctor/asciidoclet?

vyazelenko:
Also https://logging.apache.org/log4j/tools/log4j-docgen-asciidoctor-extension.html might be useful.

.github/workflows/config-doc.yml
nbradac: Good question!  I had put together this whole @Config annotation a couple years ago, but never really polished it up.  The dream was to have some way to generate all our configuration documentation based on the source code, and to have it produce the doc for Java and for C.  In some futuristic world, we'd run this workflow when we release Aeron, so as to provide a nice big document with all the configuration options in one place.

aeron-annotations/src/main/java/io/aeron/utility/JavadocCleaner.java
nbradac: :-) So... one of the reasons I revisted this after 2 years was that it seemed like a reasonable project against which to kick the tires of Claude.  When I asked it to build me something that would sanitize the javadoc to make it more fitting for a .md file, this is what it came up with.  I know there's no HTML in our javadoc, but it didn't seem like it was hurting anything to leave in the code it came up with.

build.gradle
nbradac: Ahh yes, that's right.  I can rework this thing to generate asciidoc instead of markdown.

nbradac:
The other thing that is still TODO is the contents of the CDriverConfiguration class.  All that javadoc was generated by Claude, and it... did a terrible job.  It mentions java objects all over the place.

PR #1995, opened 2026-04-10
WIP - update DriverNameResolver to align more with C code

nbradac:


aeron-driver/src/main/java/io/aeron/driver/DriverNameResolver.java
```diff
@@ -48,6 +49,8 @@
  */
 final class DriverNameResolver implements AutoCloseable, UdpNameResolutionTransport.UdpFrameHandler, NameResolver
 {
+    private static final int MAX_BOOTSTRAP_NEIGHBORS = 20;
```
vyazelenko, line 61: Where is this limit coming from, i.e. why do we want to limit the list to 20 entries?

aeron-driver/src/main/java/io/aeron/driver/DriverNameResolver.java
```diff
@@ -390,6 +408,11 @@ private void sendResolutionFrameTo(final ByteBuffer buffer, final InetSocketAddr
         {
             shortSends.increment();
         }
+
+        if (0 == bytesSent) // TODO ????
+        {
+            errors.increment();
+        }
```
vyazelenko: If this is an error? Why no error message logged?

vyazelenko:
Personally I think that Java code is more correct and we should align C with the current state of Java code instead.

aeron-driver/src/main/java/io/aeron/driver/DriverNameResolver.java
nbradac: Yeah that's the whole TODO thing on 412.  The jira ticket talks about adding some distinct error logging stuff here as well.

aeron-driver/src/main/java/io/aeron/driver/DriverNameResolver.java
nbradac: The limit comes from trying to achieve parity with the C code.

nbradac:
> Personally I think that Java code is more correct and we should align C with the current state of Java code instead.

Yeah I guess the powers that be have already looked at these differences and made some calls about when we should align with Java and when we should align with C.  There's one case where I've updated the C code (about to check that in), but mostly the C code was deemed 'better'.

mikeb01:
Would probably be best to hold off on `aeron_driver_name_resolver.c` changes as I'm in the process of porting the bootstrap neighbour changes across.

aeron-driver/src/main/java/io/aeron/driver/DriverNameResolver.java
nbradac: Turns out we don't need that bit o' code anyway.

PR #2005, opened 2026-04-15
[C] implement aeron_image_release()

nbradac:
These changes should address https://github.com/aeron-io/aeron/issues/1874


mikeb01:
It would good to add a small unit test for this.

PR #2006, opened 2026-04-15
[C] Fixed possible leak of ClientConductorIsLengthSufficientTest test parameters

strangelydim:
If the `shouldCheckIfLengthIsSufficient` test was skipped, running the rest of the test suite resulted in a memory leak, since we allocated some things via `aeron_alloc` statically at setup time but only free'd them at the end of the unit test itself. This change just avoids the heap allocation altogether since it wasn't truly needed.


vyazelenko:
@strangelydim The code does not compile.

strangelydim:
> @strangelydim The code does not compile.

Yep, forgot to add a commit that fixed a `missing-field-initializers` warning.  macOS clang and Windows MSVC were apparently just fine with the code... but Linux clang not so much.  Should be fixed now.

PR #2008, opened 2026-04-15
Added drain of queued commands during close to fix memory leaks

strangelydim:


strangelydim:
@vyazelenko hold off on this for just a bit... new test failure I need to address.

vyazelenko:
@strangelydim The existing approach assumed user's ownership of the `async` object's memory. Therefore application thread was responsible for freeing said memory by:
* Polling async object to completion.
* Cancelling async operation (new API added recently).
* Manually freeing async objects if no longer polling and/or aeron client is closed.

C++ Wrapper relies on this and does manual freeing once `Aeron` is closed, see:

https://github.com/aeron-io/aeron/blob/93e45901d06a32e490b536700e8244f0a8b411de/aeron-client/src/main/cpp_wrapper/Aeron.h#L91-L125

If we are going to change the existing logic than we need to update C++ Wrapper as well. 

I wonder what @mikeb01 thinks?

mikeb01:
The problem is if you have an cancel request inflight when the Aeron client is closed it will end up leaking because we have no other handle to that memory for a user to remove.

Another option would be to do something similar to the change I made in the driver, which was to use a ring buffer rather than a queue for the commands so that allocation is not required.  A command can just occupy space in the buffer.  For all of the add operations we would still need to allocate an object and for that store a pointer in the ring buffer, but for simple commands that don't need to be tracked by the user (e.g. cancels) the data can just be stored inline.

vyazelenko:
> The problem is if you have an cancel request inflight when the Aeron client is closed it will end up leaking because we have no other handle to that memory for a user to remove.

@mikeb01 You are right. I was thinking about the `async` stucts only but cancel requests are a problem.

The proposed approach will work, albeit not being the most efficient, since it will effectively execute all commands in the queue including create requests. The driver will remove all client-specific resources once the client is closed (provided a client can send a request to remove itself from the driver that is). 

The C++ Wrapper is effectively already using the same approach by calling `aeron_add_close_handler` which implicitly drains the command queue. So when the rest of the close proceeds there will be no commands in flight anymore.
_**NB:** All add/remove handler methods (e.g. `aeron_add_close_handler`, `aeron_remove_close_handler`, `aeron_add_available_counter_handler` etc.) are blocking and force command queue to drain in order to execute the request. Unlike the rest of the aeron client API which is asynchronous._

So the questions are:

1. Do we want to execute all pending commands or shall we just free the memory associated with them without executing?
   I'm inclined to simply free memory.

2. What shall we do with the `async` structs on close?
   Currently they are an end user responsibility. However, actual resources (e.g. `aeron_subscription_t`, `aeron_publication_t`) are freed in the close method. Therefore it seems sensible that the `async` structs in the `registering_resources` array should be also freed.

strangelydim:
OK, just pushed a new fix that I _think_ should work:

1. Keep track of all outstanding `aeron_client_remove_resource_cmd_t` objects in a list that is only ever modified by the user thread
2. Added a volatile `processed` flag to `aeron_client_remove_resource_cmd_t`, false initially, set to true by the conductor thread after it has gotten to the remove and handled it
3. `aeron_client_conductor_on_close` walks the list of outstanding `aeron_client_remove_resource_cmd_t` objects and frees all of them
4. To prevent that list from just accumulating and accumulating... it's also walked by `aeron_client_conductor_async_remove_resource`, but _only_ cmds that have already been processed are free'd and removed from the list.  But that should be good enough to keep the max size of the list fairly small in practice.

So... that's at least one possible approach.  Seems to work in my testing so far....

vyazelenko:
> OK, just pushed a new fix that I _think_ should work:
> 
>     1. Keep track of all outstanding `aeron_client_remove_resource_cmd_t` objects in a list that is only ever modified by the user thread
> 
>     2. Added a volatile `processed` flag to `aeron_client_remove_resource_cmd_t`, false initially, set to true by the conductor thread after it has gotten to the remove and handled it
> 
>     3. `aeron_client_conductor_on_close` walks the list of outstanding `aeron_client_remove_resource_cmd_t` objects and frees all of them
> 
>     4. To prevent that list from just accumulating and accumulating... it's also walked by `aeron_client_conductor_async_remove_resource`, but _only_ cmds that have already been processed are free'd and removed from the list.  But that should be good enough to keep the max size of the list fairly small in practice.
> 
> 
> So... that's at least one possible approach. Seems to work in my testing so far....

I don't think it will work.

First of all, there could be multiple application threads using a single `aeron_client_t`, i.e. the client is thread safe and could be shared across multiple threads.

Second, there are at least two threads writing to that list: application thread (when submitting) and the conductor thread (when removing). So the list pointer access must be atomic and volatile. 

Why not simply scan command queue and free all instances of the `aeron_client_remove_resource_cmd_t`?

strangelydim:
> > OK, just pushed a new fix that I _think_ should work:
> > ```
> > 1. Keep track of all outstanding `aeron_client_remove_resource_cmd_t` objects in a list that is only ever modified by the user thread
> > 
> > 2. Added a volatile `processed` flag to `aeron_client_remove_resource_cmd_t`, false initially, set to true by the conductor thread after it has gotten to the remove and handled it
> > 
> > 3. `aeron_client_conductor_on_close` walks the list of outstanding `aeron_client_remove_resource_cmd_t` objects and frees all of them
> > 
> > 4. To prevent that list from just accumulating and accumulating... it's also walked by `aeron_client_conductor_async_remove_resource`, but _only_ cmds that have already been processed are free'd and removed from the list.  But that should be good enough to keep the max size of the list fairly small in practice.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > So... that's at least one possible approach. Seems to work in my testing so far....
> 
> I don't think it will work.
> 
> First of all, there could be multiple application threads using a single `aeron_client_t`, i.e. the client is thread safe and could be shared across multiple threads.
> 
> Second, there are at least two threads writing to that list: application thread (when submitting) and the conductor thread (when removing). So the list pointer access must be atomic and volatile.
> 
> Why not simply scan command queue and free all instances of the `aeron_client_remove_resource_cmd_t`?

Ah nuts, didn't realize the client needed to be thread-safe... So nope, this won't work as-is.

I don't think the conductor thread modifies the list, though - at the point `aeron_client_conductor_on_close` is called, isn't the conductor thread already stopped?  I might be wrong...

strangelydim:
> Ah nuts, didn't realize the client needed to be thread-safe... So nope, this won't work as-is.

Could band-aid it all even more and make the list of outstanding `aeron_client_remove_resource_cmd_t` objects thread-local, and have every user thread that calls `aeron_client_conductor_async_remove_resource` register a pointer to its thread-local outstanding `aeron_client_remove_resource_cmd_t` list to a master list o' lists (just on that thread's very first call to `aeron_client_conductor_async_remove_resource`) that is protected by an actual lock.

Each individual user thread would then be responsible for the "lazy" frees of its own remove_cmds that had already been processed, and the rest would then still be cleaned up by `aeron_client_conductor_on_close` which would walk through each of the thread-local lists at that point rather than just the one big list that's in my current commit.

Too complicated?

strangelydim:
OK, pushed yet another update to make this now one of the most complicated ways to ever fix a small and infrequent memory leak ever written... but... probably works and isn't slow?

strangelydim:
> OK, pushed yet another update to make this now one of the most complicated ways to ever fix a small and infrequent memory leak ever written... but... probably works and isn't slow?

Seems to work great with pthreads... not so great on Windows, where the Aeron TLS code doesn't support TLS destructors. But could be modified to support them ala https://iangrunert.com/2025/08/04/fiber-local-storage-slow . Boy howdy, and I thought this would be simple.

strangelydim:
> > OK, pushed yet another update to make this now one of the most complicated ways to ever fix a small and infrequent memory leak ever written... but... probably works and isn't slow?
> 
> Seems to work great with pthreads... not so great on Windows, where the Aeron TLS code doesn't support TLS destructors. But could be modified to support them ala https://iangrunert.com/2025/08/04/fiber-local-storage-slow . Boy howdy, and I thought this would be simple.

Windows should probably support TLS destructors anyway for parity with Linux/macOS (we probably have a tiny leak of aeron_errmsg there now), regardless of the approach we take here, so: https://github.com/aeron-io/aeron/pull/2009

(And yes - with the commit in the above PR added to this one, the Windows tests pass for me as well.)

vyazelenko:
> > OK, pushed yet another update to make this now one of the most complicated ways to ever fix a small and infrequent memory leak ever written... but... probably works and isn't slow?
> 
> Seems to work great with pthreads... not so great on Windows, where the Aeron TLS code doesn't support TLS destructors. But could be modified to support them ala https://iangrunert.com/2025/08/04/fiber-local-storage-slow . Boy howdy, and I thought this would be simple.

@strangelydim Can we step back and think about what problem we are trying to solve?

The main issue is pending `cancel` requests that were submitted but not yet executed by the client conductor thread, right? There is no problem if:
* `cancel` request failed during submission, i.e. no space in the conductor command queue in which case the request is freed on the application thread.
* `cancel` request was submitted and client conductor thread executed it. In this case the `cancel` struct will be freed on the conductor thread.

The remaining case is when such `cancel` request sits on the conductor command queue (`command_queue`). So how about we drain this queue in the `aeron_client_conductor_on_close` method and simply free every pending command? Similar to how we delete resources in the `resource_by_id_map` map. This would also solve the issue with non-polled `async` structs and will simplify our C++ Wrapper impl.

We use a similar approach when closing `async_executor`:

https://github.com/aeron-io/aeron/blob/a092f5bd9f04f882be4bc7cdec5a3e86c115edd5/aeron-client/src/main/c/concurrent/aeron_executor.c#L79-L93

/cc @mikeb01

strangelydim:
> So how about we drain this queue in the `aeron_client_conductor_on_close` method and simply free every pending command?

With or without calling the user's `on_complete` function?  We'd probably want to call it... presumably the user set it for a good reason it. Could that cause any issues?

vyazelenko:
> > So how about we drain this queue in the `aeron_client_conductor_on_close` method and simply free every pending command?
> 
> With or without calling the user's `on_complete` function? We'd probably want to call it... presumably the user set it for a good reason it. Could that cause any issues?

With `on_complete` function for parity with normal close operation. However, `_cancel` (e.g. `aeron_async_add_subscription_cancel`) and `_remove` (e.g. `aeron_async_remove_subscription`) APIs were added just 2 weeks ago (i.e. are not yet released) so we can change them not to have the `on_complete` functions. At the momemnt those are used to free the corresponding `async` struct.

strangelydim:
I really do think the `on_complete` callbacks are useful, and so wouldn't want to remove them, but also agree that the fancy TLS approach was overkill.

So new approach: tagged pointers. Set the LSB on remove cmds to mark them as such, and then we don't need to dereference each command in `aeron_client_conductor_on_close` to tell if it was a remove command or not (because if not, it was probably already free'd, so we'd blow up just trying to check what type of command it was). Now if the LSB is set, we know it's a remove command and safe to deref during close, calling the user's `on_complete` callback inline inside `aeron_client_conductor_on_close` and then free'ing the cmd.

No insane TLS machinery needed, preserves user callbacks and ordering, simple code.  @vyazelenko - just pushed to this branch if you'd like to take a look. I think this solution is fine as long as we can stomach using tagged pointers.  Tests all pass with this approach as well.

vyazelenko:
> Set the LSB on remove cmds to mark them as such, and then we don't need to dereference each command in aeron_client_conductor_on_close to tell if it was a remove command or not (because if not, it was probably already free'd, so we'd blow up just trying to check what type of command it was). 

When a command is processed it is removed from the list. So by definition the only items in the list are non-processed items and hence not yet freed.

To me the approach with pointer tagging seems like an overkill.

aeron-client/src/main/c/aeron_client_conductor.c
```diff
@@ -245,12 +245,54 @@ static int aeron_client_conductor_release_log_buffer(aeron_client_conductor_t *c
...
 {
-    aeron_client_command_base_t *cmd = (aeron_client_command_base_t *)item;
+    aeron_client_command_base_t *cmd =
+        (aeron_client_command_base_t *)aeron_client_conductor_cmd_untag(item);
     cmd->func(clientd, cmd);
 }
 
+static void aeron_client_conductor_drain_pending_cmd_on_close(void *clientd, void *item)
+{
+    (void)clientd;
```
mikeb01, line 282: Is this line here to prevent unused parameter warnings?

strangelydim: Yep.

PR #2015, opened 2026-04-23
[C] Fix bug aeron_spsc_rb_try_claim when insufficient space begin/end

pveentjer:
When there is insufficient space at the end and at the beginning, the claim_capacity doesn't return -1 (AERON_RB_FULL) unlike the Java version.

So the caller doesn't know there was no space and starts to overwrite content in the ringbuffer. 

It can lead to:
1) corruption of the ringbuffer content 
2) the content corruption can lead to a data race between the producer and consumer, which can lead to UB.

### DIFF
--- a/aeron-client/src/main/c/concurrent/aeron_spsc_rb.c
+++ b/aeron-client/src/main/c/concurrent/aeron_spsc_rb.c
@@ -71,7 +71,7 @@ inline static int32_t aeron_spsc_rb_claim_capacity(aeron_spsc_rb_t *ring_buffer,
 
         if (required_capacity > (ring_buffer->capacity - (size_t)(tail - head)))
         {
-            return -1;
+            return AERON_RB_FULL;
         }
 
         ring_buffer->descriptor->head_cache_position = head;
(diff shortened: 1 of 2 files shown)


aeron-client/src/main/c/concurrent/aeron_spsc_rb.c
```diff
@@ -135,7 +135,7 @@ inline static int32_t aeron_spsc_rb_claim_capacity(aeron_spsc_rb_t *ring_buffer,
         next_header->msg_type_id = 0;
     }
```
pveentjer, line 137: Here is the actual bug fix; instead of returning record_index, the write_index should be returned. The write_index contains AERON_RB_FULL if there is no space.

The Java version is correctly returning the write_index. 

https://github.com/aeron-io/agrona/blob/cfd91f0e0ae6f0a4f980f674b823a5f3eb31d66a/agrona/src/main/java/org/agrona/concurrent/ringbuffer/OneToOneRingBuffer.java#L487

aeron-client/src/main/c/concurrent/aeron_spsc_rb.c
```diff
@@ -71,7 +71,7 @@ inline static int32_t aeron_spsc_rb_claim_capacity(aeron_spsc_rb_t *ring_buffer,
 
         if (required_capacity > (ring_buffer->capacity - (size_t)(tail - head)))
         {
-            return -1;
+            return AERON_RB_FULL;
```
pveentjer, line 74: AERON_RB_FULL is defined as -1. So I'm replacing the magic numbers by the appropriate constant.

vyazelenko:
@pveentjer Tests are broken.

PR #2016, opened 2026-04-23
Persistent Subscriptions

nickelaway:
[Java/C/C++ Wrapper] Introduce the Persistent Subscriptions feature for seamlessly switching between live and replay channels, supporting different media types.

---------


aeron-archive/src/main/java/io/aeron/archive/client/PersistentSubscription.java
```diff
@@ -0,0 +1,2328 @@
...
+        try
+        {
+            this.uncontrolledFragmentHandler = fragmentHandler;
+            return doWork(fragmentLimit, false);
+        }
+        finally
+        {
+            this.uncontrolledFragmentHandler = null;
+        }
+    }
```
vyazelenko, line 221: Have you considered using a different approach to this API without passing `FragmentHandler` to every call? Similar to how `AeronCluster` API for polling egress works?

The API then becomes:
```
public int poll(){...}
public int controlledPoll(){...}
```

The main issue with the existing approach is that field write might be expensive (GC dependent) due to card marking. Also the code is more compilicated and more error prone to use, i.e. what if user passes different handler every time poll is called?

aeron-archive/src/main/java/io/aeron/archive/client/PersistentSubscription.java
```diff
@@ -0,0 +1,2328 @@
...
+        {
+            controlledFragmentHandler = fragmentHandler;
+            return doWork(fragmentLimit, true);
+        }
+        finally
+        {
+            controlledFragmentHandler = null;
+        }
+    }
+
```
vyazelenko, line 251: Same as above.

aeron-archive/src/main/java/io/aeron/archive/client/PersistentSubscription.java
```diff
@@ -0,0 +1,2328 @@
...
+    {
+        return new PersistentSubscription(ctx);
+    }
+
+    /**
+     * Poll for the next available message(s).
+     * <p>
+     * Either this method or {@link #controlledPoll(ControlledFragmentHandler, int)} must be called in a duty cycle for
+     * the {@code PersistentSubscription} to perform its work.
+     *
```
vyazelenko, line 205: Would be good to mention how errors are propagated here. I was surprised that we don't throw and that `PersistentSubscription` is not closed on terminal error unlike `AeronCluster` for instance.

aeron-archive/src/main/java/io/aeron/archive/client/PersistentSubscription.java
```diff
@@ -0,0 +1,2328 @@
...
+        {
+            this.uncontrolledFragmentHandler = null;
+        }
+    }
+
+    /**
+     * Poll for the next available message(s).
+     * <p>
+     * Either this method or {@link #poll(FragmentHandler, int)} must be called in a duty cycle for the
+     * {@code PersistentSubscription} to perform its work.
```
vyazelenko, line 229: Mention error handling.

aeron-archive/src/main/java/io/aeron/archive/client/PersistentSubscription.java
```diff
@@ -0,0 +1,2328 @@
...
+                   ", errorMessage='" + errorMessage + '\'' +
+                   ", responseReceived=" + responseReceived +
+                   '}';
+        }
+    }
+
+    /**
+     * Configuration of a {@code PersistentSubscription} to be created.
+     */
+    public static class Context implements Cloneable
```
vyazelenko: Should be `final`.

aeron-archive/src/main/java/io/aeron/archive/client/PersistentSubscription.java
```diff
@@ -0,0 +1,2328 @@
...
+            {
+                throw new ConfigurationException("recordingId must be set");
+            }
+
+            if (Aeron.NULL_VALUE == liveStreamId)
+            {
+                throw new ConfigurationException("liveStreamId must be set");
+            }
+
+            if (null == liveChannel)
```
vyazelenko: Use `Strings.isEmpty(liveChannel)` to cover for empty string case.

aeron-archive/src/main/java/io/aeron/archive/client/PersistentSubscription.java
```diff
@@ -0,0 +1,2328 @@
...
+            {
+                throw new ConfigurationException("liveChannel must be set");
+            }
+
+            if (null == replayChannel)
```
vyazelenko: Use `Strings.isEmpty(replayChannel)` to cover for empty string case.

aeron-archive/src/main/java/io/aeron/archive/client/PersistentSubscription.java
```diff
@@ -0,0 +1,2328 @@
...
+                        "Channel media type mismatch. " +
+                            "When using `control-mode=response`, the `replayChannel` media type must match the media" +
+                            " type for the archive control channel."
+                    );
+                }
+            }
+
+            if (null == aeron)
+            {
+                final Aeron.Context aeronCtx = new Aeron.Context()
```
vyazelenko, line 1723: Have you considered running `Aeron` instance in the invoker mode by default, i.e. to avoid creating an additional thread?

aeron-archive/src/main/java/io/aeron/archive/client/PersistentSubscriptionException.java
```diff
@@ -0,0 +1,75 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.aeron.archive.client;
+
+/**
+ * Exception raised when using a {@link PersistentSubscription}.
+ */
+public class PersistentSubscriptionException extends RuntimeException
```
vyazelenko: Why does not it extend `AeronException`? Also should be final.

aeron-archive/src/main/java/io/aeron/archive/client/PersistentSubscription.java
```diff
@@ -0,0 +1,2328 @@
...
+        liveChannel = ctx.liveChannel;
+        liveStreamId = ctx.liveStreamId;
+        replayChannel = ctx.replayChannel;
+        replayChannelUri = ChannelUri.parse(replayChannel);
+        replayChannelType = ReplayChannelType.of(replayChannelUri);
+        replayStreamId = ctx.replayStreamId;
+        listener = ctx.listener;
+        aeron = ctx.aeron;
+        nanoClock = aeron.context().nanoClock();
+        asyncAeronArchive = new AsyncAeronArchive(ctx.aeronArchiveContext().aeron(aeron), new ArchiveListener());
```
vyazelenko: I don't think it is correct to re-assign `Aeron` reference on the supplied `AeronArchive.Context`? If none was provided then the local `Aeron` should be used but if `AeronArchive.Context` has one assigned we should used that.

Also this configuration step should move to the `conclude` method.

aeron-agent/src/main/java/io/aeron/agent/ArchiveEventCode.java
```diff
@@ -234,7 +235,15 @@ public enum ArchiveEventCode implements EventCode
...
-        45, MaxRecordedPositionRequestDecoder.TEMPLATE_ID, ArchiveEventDissector::dissectControlRequest);
+        45, MaxRecordedPositionRequestDecoder.TEMPLATE_ID, ArchiveEventDissector::dissectControlRequest),
+    /**
+     * Archive logging event for {@link PersistentSubscription} state change.
+     */
+    PERSISTENT_SUBSCRIPTION_STATE_CHANGE(46, -1, ArchiveEventDissector::dissectPersistentSubscriptionStateChange),
+    /**
+     * Archive logging event for {@link PersistentSubscription} joining live.
+     */
+    PERSISTENT_SUBSCRIPTION_JOINED_LIVE(47, -1, ArchiveEventDissector::dissectPersistentSubscriptionJoinedLive);
```
vyazelenko: Why a separate logging event for joining live, i.e. it is covered by state change logging?
Also if this is important then why we don't have a separate event for dropping from live?

vyazelenko (changes requested):
`CHANGELOG.adoc` needs to be updated to include `PersistenSubscription` in the noteworthy section. We also need a wiki page about PS usage.

aeron-agent/src/main/java/io/aeron/agent/ArchiveEventCode.java
nickelaway: The difference between this and the state-transition logging is that this one logs the position at which we switch. We added this one as it might help diagnose issues where a live transition impacts other nodes. For example, if your Persistent Subscription adds live, but doesn't switch right away, it could back-pressure other min-flow-control consumers. Including the position that we add live could be a useful bit of debug information there. That's also why we expose the joinDifference.

Switching to replay is a less sensitive transition. Although, in retrospect, I think it would also be useful to see the position at which that happens. So we could maybe add a logging event for that, too.

aeron-archive/src/main/java/io/aeron/archive/client/PersistentSubscription.java
nickelaway: Yeah, we considered requiring people add the fragment handler when creating the Persistent Subscription. One of the goals was to present a similar API to Subscription (as much as is practical), which is why we went with the approach of passing in the FragmentHandler. That's also how you use ReplayMerge.

The field write is an interesting point. Do you think that is likely to be problematic? Worth changing the API for that?

Is passing a different handler on different calls an error? Is the error that someone might accidentally create a new one for each poll without realising? If people do have a legitimate use-case for passing different handlers on different calls, I would expect that to work fine. Is this an issue we've seen with Subscription?

aeron-agent/src/main/java/io/aeron/agent/ArchiveEventCode.java
vyazelenko: Yes, we should add the other logging event.

aeron-archive/src/main/java/io/aeron/archive/client/PersistentSubscription.java
vyazelenko: > The field write is an interesting point. Do you think that is likely to be problematic? Worth changing the API for that?

This is a performance issue which might be a problem for some use cases. It is hard to tell without testing. And even then a benchmark might not reflect the real impact.

> Is passing a different handler on different calls an error?

It is not an error per se but I would say it probably is. I don't see why anyone would pass different handlers when consuming a stream of messages. If there is any logic that depends on the messages processes then it should be a part of the lambda function passed in.

aeron-archive/src/main/java/io/aeron/archive/client/PersistentSubscriptionException.java
pveentjer: Fixed

aeron-archive/src/main/java/io/aeron/archive/client/PersistentSubscription.java
vyazelenko: > It is in invoker mode by default:

                if (null == aeron)
                {
                    final Aeron.Context aeronCtx = new Aeron.Context()
                        .clientName("PersistentSubscription")
                        .useConductorAgentInvoker(true);<-----
                    ...
    

> Or am I missing something?

@pveentjer It was not there when I started my review. It was added later in this commit https://github.com/aeron-io/aeron/commit/5a2eaf5e031002ff7e07074bca54c3c7e0bce309.

pveentjer: Yes. I figured that part already out. That is why I removed my comment.

aeron-archive/src/main/c/client/aeron_archive_persistent_subscription.h
```diff
@@ -0,0 +1,469 @@
...
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
+#define AERON_ARCHIVE_PERSISTENT_SUBSCRIPTION_FROM_START (INT64_C(-1))
+#define AERON_ARCHIVE_PERSISTENT_SUBSCRIPTION_FROM_LIVE  (INT64_C(-2))
+
+typedef struct aeron_archive_persistent_subscription_context_stct aeron_archive_persistent_subscription_context_t;
+typedef struct aeron_archive_persistent_subscription_stct aeron_archive_persistent_subscription_t;
```
vyazelenko: Any particular reason why these structs are not declared in the header file?

aeron-archive/src/main/c/client/aeron_archive_persistent_subscription.c
```diff
@@ -0,0 +1,2425 @@
...
+        NULL,
+        NULL,
+        NULL,
+        NULL) < 0)
+    {
+        aeron_archive_persistent_subscription_transition(persistent_subscription, FAILED);
+
+        aeron_archive_persistent_subscription_on_terminal_error(
+            persistent_subscription,
+            aeron_errcode(),
```
vyazelenko: Naked error code should not be passed to the user code, i.e. Aeron codes are set as negative values (e.g. `-12` for storage space error). However, constants themselves are defined as positive values (e.g. `AERON_ERROR_CODE_STORAGE_SPACE (12)`). Therefore negative error codes should be converted to positive error code before calling user defined callbacks.

aeron-archive/src/main/c/client/aeron_archive_persistent_subscription.c
```diff
@@ -0,0 +1,2425 @@
...
+}
+
+static int aeron_archive_persistent_subscription_await_replay_subscription(
+    aeron_archive_persistent_subscription_t *persistent_subscription)
+{
+    if (aeron_async_add_subscription_poll(
+        &persistent_subscription->replay_subscription,
+        persistent_subscription->add_replay_subscription) < 0)
+    {
+        int errcode = aeron_errcode();
```
vyazelenko: Naked error code should not be passed to the user code (see above).

aeron-archive/src/main/c/client/aeron_archive_persistent_subscription.c
```diff
@@ -0,0 +1,2425 @@
...
+        const char *errmsg = aeron_errmsg();
+        persistent_subscription->add_replay_subscription = NULL;
+        aeron_archive_persistent_subscription_clean_up_replay(persistent_subscription);
+
+        if (-AERON_ERROR_CODE_RESOURCE_TEMPORARILY_UNAVAILABLE == errcode)
```
vyazelenko: If error code is translated to positive value first then direct constant comparison can be used instead. So instead of calling `aeron_errcode` directly call a function that will convert negative to positive code. This will solve all usage issues.

aeron-archive/src/main/c/client/aeron_archive_persistent_subscription.c
```diff
@@ -0,0 +1,2425 @@
...
+    aeron_uri_string_builder_close(&builder);
+
+    if (failed)
+    {
+        AERON_APPEND_ERR("%s", "failed to build request publication channel");
+        aeron_archive_persistent_subscription_transition(persistent_subscription, FAILED);
+
+        aeron_archive_persistent_subscription_on_terminal_error(
+            persistent_subscription,
+            aeron_errcode(),
```
vyazelenko: Naked `aeron_errcode` usage.

aeron-archive/src/main/c/client/aeron_archive_persistent_subscription.c
```diff
@@ -0,0 +1,2425 @@
...
+        persistent_subscription->context->aeron,
+        channel,
+        stream_id) < 0)
+    {
+        AERON_APPEND_ERR("%s", "failed to add request publication");
+        aeron_archive_persistent_subscription_transition(persistent_subscription, FAILED);
+
+        aeron_archive_persistent_subscription_on_terminal_error(
+            persistent_subscription,
+            aeron_errcode(),
```
vyazelenko: Naked `aeron_errcode` usage.

aeron-archive/src/main/c/client/aeron_archive_persistent_subscription.c
```diff
@@ -0,0 +1,2425 @@
...
+    if (0 == result)
+    {
+        return 0;
+    }
+
+    persistent_subscription->add_request_publication = NULL;
+
+    if (result < 0)
+    {
+        int errcode = aeron_errcode();
```
vyazelenko: Naked `aeron_errcode` usage.

aeron-archive/src/main/c/client/aeron_archive_persistent_subscription.c
```diff
@@ -0,0 +1,2425 @@
...
+        persistent_subscription->context->archive_context,
+        persistent_subscription->request_publication,
+        AERON_ARCHIVE_MESSAGE_RETRY_ATTEMPTS_DEFAULT) < 0)
+    {
+
+        aeron_archive_persistent_subscription_transition(persistent_subscription, FAILED);
+
+        aeron_archive_persistent_subscription_on_terminal_error(
+            persistent_subscription,
+            aeron_errcode(),
```
vyazelenko: Naked `aeron_errcode` usage.

aeron-archive/src/main/c/client/aeron_archive_persistent_subscription.c
```diff
@@ -0,0 +1,2425 @@
...
+        NULL,
+        NULL,
+        NULL) < 0)
+    {
+        AERON_APPEND_ERR("%s", "failed to add live subscription");
+        aeron_archive_persistent_subscription_transition(persistent_subscription, FAILED);
+
+        aeron_archive_persistent_subscription_on_terminal_error(
+            persistent_subscription,
+            aeron_errcode(),
```
vyazelenko: Naked `aeron_errcode` usage.

aeron-archive/src/main/c/client/aeron_archive_persistent_subscription.c
```diff
@@ -0,0 +1,2425 @@
...
+        return 1;
+    }
+
+    if (NULL == persistent_subscription->live_subscription && NULL != persistent_subscription->add_live_subscription)
+    {
+        if (aeron_async_add_subscription_poll(
+            &persistent_subscription->live_subscription,
+            persistent_subscription->add_live_subscription) < 0)
+        {
+            int errcode = aeron_errcode();
```
vyazelenko: Naked `aeron_errcode` usage.

aeron-archive/src/main/c/client/aeron_archive_persistent_subscription.c
```diff
@@ -0,0 +1,2425 @@
...
+static int aeron_archive_persistent_subscription_await_live(
+    aeron_archive_persistent_subscription_t *persistent_subscription)
+{
+    if (NULL == persistent_subscription->live_subscription)
+    {
+        if (aeron_async_add_subscription_poll(
+            &persistent_subscription->live_subscription,
+            persistent_subscription->add_live_subscription) < 0)
+        {
+            int errcode = aeron_errcode();
```
vyazelenko: Naked `aeron_errcode` usage.

vyazelenko (changes requested):
C error handling.

aeron-archive/src/main/java/io/aeron/archive/client/PersistentSubscription.java
vyazelenko: @nickelaway So any internal error is treated as terminal while if user code throws it is not a terminal error? 

> Exceptions thrown from the {@link FragmentHandler} will be propagated.

This is not true, i.e. it depends on how the `Aeron.Context#subscriberErrorHandler(org.agrona.ErrorHandler)` is configured. By default it will delegate to the `Aeron.Context#errorHandler(org.agrona.ErrorHandler)` which will print exception stacktrace but won't rethrow anything. 

There is `RethrowingErrorHandler` that can be used to force rethrowing of the frame handler exceptions. See:

https://github.com/aeron-io/aeron/blob/30e02ded8bc7217da3d5c3ab3be50f650762c8a0/aeron-archive/src/main/java/io/aeron/archive/client/AeronArchive.java#L3111-L3120

I think PS should also configure `RethrowingErrorHandler` by default.

vyazelenko (changes requested):

aeron-archive/src/main/c/client/aeron_archive_persistent_subscription.c
pveentjer: Is it sufficient to flip the sign of the error code? And what about the already positive errno codes like EOI? Could it be that the error code becomes ambiguous; so when you receive it you don't know if it e.g. a EBUSY or INVALID_POSITION (both have value 16).

CHANGELOG.adoc
```diff
@@ -4,6 +4,10 @@
 
 === New & Noteworthy
 
+* **[Archive]** Added `PersistentSubscription` to Aeron Archive for all languages (Java, C++ and C).
++
+`PersistentSubscription` is an alternative to ReplayMerge with IPC and Spy support, plus a fallback mechanism for slow consumers.
+
```
vyazelenko, line 10: @Nadia-Adaptive Can we add a bit more description in here or point people to a wiki/doc/samples/tests for more description? It seems unfair that the main item of the release has barely any mention in the release notes.

vyazelenko (changes requested):

aeron-archive/src/main/java/io/aeron/archive/client/PersistentSubscription.java
nickelaway: > @nickelaway So any internal error is treated as terminal while if user code throws it is not a terminal error?

Not quite. Most of the internal errors are terminal, but some are not. For example, if we get a ErrorCode.RESOURCE_TEMPORARILY_UNAVAILABLE, that's not terminal.

Exceptions from user code don't fail the persistent subscription, though. 

> This is not true, i.e. it depends on how the
Ok. I will look at this part again.

CHANGELOG.adoc
nickelaway: We have some docs that we will publish to the Wiki. We can include the link to that here, once we've published it. I was inclined to not publish that page until we are ready to release.

vyazelenko: @nickelaway Agree with the approach of adding docs once code is merged.

aeron-archive/src/main/java/io/aeron/archive/client/PersistentSubscription.java
vyazelenko: > Exceptions from user code don't fail the persistent subscription, though.

That's fine. But the Javadoc mentions that exceptions in user code will be re-thrown which is not true with the default configuration anyway. My comment above was bout that (i.e. using `RethrowingErrorHandler`).

aeron-archive/src/main/c/client/aeron_archive_persistent_subscription.h
pveentjer: The above issue is fixed with this PR:

https://github.com/aeron-io/aeron/pull/2023

PR #2019, opened 2026-04-24
persistent_subscription invoker-mode by default for aeron

pveentjer:


vyazelenko:
The agent invoker must be called on every duty cycle. Not just when adding counters. See java implementation.

PR #2024, opened 2026-04-27
Added check to skip re-resolution for response control mode endpoints

strangelydim:
No need to re-resolve things for an endpoint that should just be null anyway. Already enforced here:

https://github.com/aeron-io/aeron/commit/411f7b6a64e99ea834fcda41af4f95b0a7741034#diff-a29f3df5050a13706ce29b053e4f5b062097b2e5b01f67a8b365ae6875a62ac1L105

when first constructing the endpoint, but the same check was missing in `checkForReResolution`.


vyazelenko:
@strangelydim The same fix should be applied to the C code as well.

strangelydim:
> @strangelydim The same fix should be applied to the C code as well.

Good point, and a couple unit tests wouldn't hurt, either... so added both.

PR #2037, opened 2026-05-01
[Java/C] Refresh the recording descriptor before attempting the replay again if a Persistent Subscription loses its image during a replay.

nickelaway:
This prevents it from attempting to start a replay from the recording stop position when the recording and live publication have stopped.


strangelydim (approved):
Makes sense to me!

PR #2039, opened 2026-05-01
Bug/archive dir conflicts

ryanoftalltrees:
Detect mis-configured mark file location to avoid corrupting an archive in use

### DIFF
--- a/aeron-archive/src/main/java/io/aeron/archive/Archive.java
+++ b/aeron-archive/src/main/java/io/aeron/archive/Archive.java
@@ -1244,6 +1244,11 @@ public final class Archive implements AutoCloseable
                 archiveDir = new File(archiveDirectoryName);
             }
 
+            if (null != markFile)
+            {
+                markFileDir = markFile.parentDirectory();
+            }
+
             if (null == markFileDir)
             {
                 final String markFileDirPath = Configuration.markFileDir();
@@ -1288,10 +1293,7 @@ public final class Archive implements AutoCloseable
                 epochClock = SystemEpochClock.INSTANCE;
             }
 
-            if (null == markFile)
-            {
-                ensureNoConflictingMarkFile(archiveDir, markFileDir, epochClock);
-            }
+            ensureNoConflictingMarkFile(archiveDir, markFileDir, epochClock);
 
             if (null == nanoClock)
             {
--- a/aeron-archive/src/main/java/io/aeron/archive/Archive.java
+++ b/aeron-archive/src/main/java/io/aeron/archive/Archive.java
@@ -1290,7 +1290,7 @@ public final class Archive implements AutoCloseable
 
             if (null == markFile)
             {
-                ensureNoConflictingMarkFile(archiveDir, epochClock);
+                ensureNoConflictingMarkFile(archiveDir, markFileDir, epochClock);
             }
 
             if (null == nanoClock)
@@ -3805,13 +3805,9 @@ public final class Archive implements AutoCloseable
             }
         }
 
-        private static void ensureNoConflictingMarkFile(final File archiveDir, final EpochClock epochClock)
+        private static void ensureNoConflictingMarkFile(
+            final File archiveDir, final File markFileDir, final EpochClock epochClock)
         {
-            if (ArchiveMarkFile.isActive(archiveDir, epochClock))
-            {
-                throw new ConfigurationException("active archive detected, archive.dir=" + archiveDir + " is in use");
-            }
-
             final File linkFile = new File(archiveDir, ArchiveMarkFile.LINK_FILENAME);
             if (linkFile.exists())
             {
@@ -3820,12 +3816,15 @@ public final class Archive implements AutoCloseable
                     final String path = new String(Files.readAllBytes(linkFile.toPath()), US_ASCII).trim();
                     if (!path.isEmpty())
                     {
-                        final File markFileDir = new File(path);
-                        if (ArchiveMarkFile.isActive(markFileDir, epochClock))
+                        final File linkedMarkFileDir = new File(path).getCanonicalFile();
+                        if (!linkedMarkFileDir.equals(markFileDir) &&
+                            ArchiveMarkFile.isActive(linkedMarkFileDir, epochClock))
                         {
                             throw new ConfigurationException(
-                                "active archive detected, archive.dir=" + archiveDir +
-                                " mark file dir=" + markFileDir + " is in use");
+                                "There is an active Archive at archive.dir=" + archiveDir +
+                                    " detected through an existing link to " +
+                                    "mark file dir=" + linkedMarkFileDir + ", but this Archive is configured " +
+                                    "with mark file dir=" + markFileDir);
                         }
                     }
                 }
@@ -3833,6 +3832,13 @@ public final class Archive implements AutoCloseable
                 {
                     throw new UncheckedIOException(ex);
                 }
+            }
+
+            if (!markFileDir.equals(archiveDir) && ArchiveMarkFile.isActive(archiveDir, epochClock))
+            {
+                throw new ConfigurationException(
+                    "archive.dir=" + archiveDir + " has an active mark file but this Archive is configured" +
+                    " with mark file dir=" + markFileDir);
             }
         }
 
--- a/aeron-archive/src/main/java/io/aeron/archive/Archive.java
+++ b/aeron-archive/src/main/java/io/aeron/archive/Archive.java
@@ -1288,6 +1288,11 @@ public final class Archive implements AutoCloseable
                 epochClock = SystemEpochClock.INSTANCE;
             }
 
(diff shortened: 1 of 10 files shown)


aeron-archive/src/main/java/io/aeron/archive/ArchiveMarkFile.java
```diff
@@ -556,6 +556,47 @@ private static MarkFile openExistingMarkFile(
...
+        final int mapOffset = Math.min(versionOffset, timestampOffset);
+        final int mapLength = Math.max(versionOffset, timestampOffset) + Long.BYTES - mapOffset;
+
+        final MappedByteBuffer byteBuffer = MarkFile.mapExistingFile(markFile, null, mapOffset, mapLength);
+        if (null == byteBuffer)
+        {
+            return false;
+        }
+
+        try
```
vyazelenko: The logic of handling different versions of  already exists in the `ArchiveMarkFile` constructor. So just call it here instead.

aeron-archive/src/main/java/io/aeron/archive/Archive.java
```diff
@@ -1288,6 +1288,11 @@ public void conclude()
                 epochClock = SystemEpochClock.INSTANCE;
             }
 
+            if (null == markFile)
```
vyazelenko: I think the check should be unconditional.

aeron-archive/src/main/java/io/aeron/archive/Archive.java
```diff
@@ -3800,6 +3805,43 @@ private void concludeArchiveId()
             }
         }
 
+        private static void ensureNoConflictingMarkFile(
```
vyazelenko, line 3810: There are two issues:
* We need similar check for `ClusterMarkFile`.

* The check is racy, i.e. two processes starting simultaneously using different configuration will pass the check and will violate the condition of a single Archive running in the `archive.dir`.
  One option would be enforce the same value `mark.file.dir` or even remove this config option and put it under some known location (e.g. `aeron.dir`) unconditionally. Requires some thinking...

vyazelenko (changes requested):

aeron-archive/src/main/java/io/aeron/archive/ArchiveMarkFile.java
ryanoftalltrees: Done.  I was trying to avoid reading in the whole file just to look at the activity timestamp thinking it would interfere with the other archive that could be using it, but that's probably not an issue here.

aeron-archive/src/main/java/io/aeron/archive/Archive.java
ryanoftalltrees: Done.  Also made sure the mark file dir matches the user supplied mark file for consistency.

aeron-archive/src/main/java/io/aeron/archive/Archive.java
ryanoftalltrees: Added the checks for Cluster. ClusteredService, and ClusterBackup.  I'm not sure how to solve the race problem.  At least now if they start at slightly different times, it will be caught.  Maybe we can defer that to later?

vyazelenko:
`MarkFile` is designed to ensure there is only one instance of a component running at a time. The expected usage is to simply try to create a `MarkFile` instance in the target directory and it takes care of the rest (e.g. atomically creating file, dealing with racy start etc.).

The issue in Aeron is that it allows relocating `MarkFile` to a different directory. Therefore it is possible to start two instances of the same component sharing the same data directory but placing `MarkFile` in different directories. With two `MarkFiles` we can no longer ensures atomicity of the access which breaks the premise of the approach.

I see several possible approaches to solve it:
* Do not relocate mark files.
  This is bad, because of the periodic activity timestamp updates which will casue stalls if mark file is located on a real disc instead of `/dev/shm`.
  
* Use the same predefined location for relocating mark files of a component.
  The location should be independent of other config in order to ensure the final mark file path is the same, i.e. should not be relative to the `aeron.dir`. One possible approach is to use component dir name (i.e. `archive.dir`, `cluster.dir`) to compute the mark file path (e.g. `/dev/shm/${archive.dir.name}`).

* Atomically create and write link file.
  Currently the link file is optionally created if the mark file is relocated and otherwise it is deleted.
  
  The idea is to make link file required and atomically create/update it in order to prevent other thread/process from overwriting it. Once this is in place the other checks added by this PR are no longer necessary, i.e. since the link file is created in the target dir it becomes the missing concurrency control.

  Making link file concurrent would require changing its format to have a header where we can CAS some value. The simplest is to add 8 byte `timestamp` at the beginning of the file. Then do what `MarkFile` does with its own `timestamp`, i.e.:
  1. Read the timestamp value from the file.
  2. Get current time.
  3. Compare file timestamp with the current time and throw if _active_ file detected
  4. CAS a special sentinel value (`Long#MAX_VALUE`) into the `timestamp` slot, i.e. if CAS fails then throw _active_ file detected exception.
    _NB: The special value is there to ensure that a concurrent writer will either fail at the validation step (3) or at step (4)._
  5. Write the rest of the fields, i.e. path to the mark file dir.
  6. Update `timestamp` with the current time.
  
  I think these changes should be made in Agrona, i.e. [MarkFile#ensureMarkFileLink](https://github.com/aeron-io/agrona/blob/aac2cda7359816154fb3a989f46a93a297f71f10/agrona/src/main/java/org/agrona/MarkFile.java#L817) method. We will also need to add a method to read link contents, i.e. to skip header. Perhaps there should be a separate class `MarkFileLink` or something.

mikeb01:
Were we to scrap the idea of a link file altogether and we go back to having a mark file in the archive/cluster directory.  Then for optimisation the application can create a file in /dev/shm to store actively written data.  Likely timestamp and error logs.  The mark file will still contain a timestamp, but it will have a special value to indicate if the timestamp should be read from somewhere else.  The Markfile can contain the location of that data file.

This gets away from the idea of searching for a mark file or a link file.  Access is always through the mark file, but the timestamp (and potentially other data) might need to be redirected.  It should be easier to manage the concurrency this way.  The start up process of an new file would be:

1. create mark file
2. write current timestamp
3. create data file in /dev/shm.
4. write timestamp to data file
5. put location of data file in the mark file
6. set the timestamp value in the mark file to indicate a redirect.

The if the mark file exists, read the timestamp, if it is positive to a timestamp check.  If value indicates a redirect, get the location from the mark file and read the timestamp from there.

We would probably need to have some backward compatibility checks for older versions.

vyazelenko:
@mikeb01 I see three possible solutions:

1. Keep using the link and creation/update atomic.
   Pros:
   - No changes to the Archive/Cluster config.
   - Can be made backwards compatible.
   - The complexity is hidden inside of the `org.agrona.MarkFile#ensureMarkFileLink`.

   Cons:
   - Link is required and always created. 
   - Some nasty concurrency code around link management in addition to the mark file, i.e. we double the concurrency efforts.
   - Requires new link file format, i.e. backwards compatibility can be preserved but mixing old and new versions will not work.

2. Keep mark file in the data dir but externalize time keeping, i.e. similar to your proposal.
   Pros:
   - No link file to deal with.

   Cons:
   - Configuration changes to Archive/Cluster.
   - Mark file format changes.
   - Time keeping redirection adds complexity and more failure/concurrency issues.

3. Keep mark file in a known location outside of the data dir and `aeron.dir`.

   If we can map data dir (`archive.dir`/`cluster.dir`) path to the unique name (e.g. compute SHA-2 hash or some other hashing scheme). Then we can use that name as a directory name in a well known location (i.e. `/dev/shm` on Linux etc.) in order to store mark file in it. This way the location is dependent on the data dir path only and does not reside in the persistent storage (so no latency penalty for updates either). 
   
   Pros:
   - There is only a mark file.
   - No link file.
   - The logic remains as it was, i.e. no changes to the mark file format needed.

   Cons:
   - Configuration changes to Archive/Cluster.
   - Backwards compability needs to be factored in.

I personaly would go for option 3 (or 1 if 3 is a no go).

I would also like us to extract error log buffer out of the mark files into a separate `error.log` file that will be located in the data dir directly, i.e. in order to make errors persistent by default. This will also make preserving errors from previous runs nicers, i.e. place `error-$date-time.log` in the same directory upon component restart.

If we keep the error buffer in the mark file. We should at the very least place those `error-$date-time.log` in the data dir instead of keeping them in the `dev/shm` as we currently do.

mikeb01:
I'm not a fan of option 1 specifically around the con that you mention:

> Some nasty concurrency code around link management in addition to the mark file, i.e. we double the concurrency efforts.

The tricky part of option 3 is the 'known location' part of it.  This would mean hardcoding some sort of root path to that location.  We would not be able to make this user configurable as it would result in the same issues that triggered this particular bug in the first place.  I worry about a solution that relies on fixed configuration, e.g. file system location as it can possibly cause unseen issues

PR #2042, opened 2026-05-06
Support using Cpusets to handle thread affinity.

mikeb01:
- Read the current process's cgroup and effective cpuset.
- Use the thread affinity setting as an offset into the cpuset.
- Validate that all siblings are in the same cpuset.
- Validate that all of the cpus share the same L3.
- Validate that all of the cpus share the same cluster.
- Toggle the ability to fail if any warnings detected.


vyazelenko:
Compilation error on MSVC:
```bash
D:\a\aeron\aeron\aeron-driver\src\main\c\aeron_cpuset.c(22,1): error C1083: Cannot open include file: 'unistd.h': No such file or directory [D:\a\aeron\aeron\cppbuild\Release\aeron-driver\src\main\c\aeron_driver.vcxproj]
```

PR #2043, opened 2026-05-06
fix a race condition in client conductor

egnchen:
## [C] Fix TSan race conditions from missing AERON_GET_ACQUIRE

Replace plain reads of cross-thread-shared fields with `AERON_GET_ACQUIRE` to ensure correct memory ordering and prevent stale values detected by ThreadSanitizer.

### `aeron_client_conductor.c`

Eight sites iterating `registering_resources` and reading `resource->registration_status` now acquire-read into a local `status` before the comparison:

| `aeron_client_conductor.c` function | Field | Thread interaction |
|---|---|---|
| `check_registering_resources` | `registration_status` | conductor side of registration state machine |
| `on_error` | `registration_status` | driver response handler reads state written by conductor |
| `on_publication_ready` | `registration_status` | same |
| `on_subscription_ready` | `registration_status` | same |
| `on_counter_ready` | `registration_status` | same |
| `on_static_counter` | `registration_status` | same |
| `on_next_available_session_id` | `registration_status` | same |
| `on_operation_success` | `registration_status` | same |

### `aeron_network_publication.c`

Two sender-thread reads of `has_receivers` now use `AERON_GET_ACQUIRE`, matching `AERON_SET_RELEASE` writes from the conductor thread:

| Location | Field | Thread interaction |
|---|---|---|
| `setup_message_check` | `has_receivers` | sender reads, conductor writes via `AERON_SET_RELEASE` |
| `send` | `has_receivers` | same |

### `aeron_publication_image.c`

One receiver-thread read of `is_end_of_stream` now uses `AERON_GET_ACQUIRE`, matching the `AERON_SET_RELEASE` write from the conductor thread:

| Location | Field | Thread interaction |
|---|---|---|
| `insert_packet` | `is_end_of_stream` | receiver reads, conductor writes via `AERON_SET_RELEASE` |

courtesy of TSAN, deepseek v4 pro and GLM 5.1.

---

> old pr description:
> fix a race condition in client conductor
> image
> discovered by tsan, didn't run the full test case, but you get the idea.

### DIFF
--- a/aeron-client/src/main/c/aeron_client_conductor.c
+++ b/aeron-client/src/main/c/aeron_client_conductor.c
@@ -1437,7 +1437,7 @@ static int aeron_client_conductor_check_registering_resources(aeron_client_condu
     {
         aeron_client_registering_resource_t *resource = conductor->registering_resources.array[index].resource;
         aeron_client_registration_status_t status;
-        AERON_GET_ACQUIRE(status, resource->registeration_status);
+        AERON_GET_ACQUIRE(status, resource->registration_status);
 
         if (AERON_CLIENT_REGISTRATION_STATUS_AWAITING == status && now_ns > resource->registration_deadline_ns)
         {
--- a/aeron-client/src/main/c/aeron_client_conductor.c
+++ b/aeron-client/src/main/c/aeron_client_conductor.c
@@ -1436,14 +1436,15 @@ static int aeron_client_conductor_check_registering_resources(aeron_client_condu
     for (size_t size = conductor->registering_resources.length, last_index = size - 1, i = size, index = i - 1; i > 0; i--, index--)
     {
         aeron_client_registering_resource_t *resource = conductor->registering_resources.array[index].resource;
+        aeron_client_registration_status_t status;
+        AERON_GET_ACQUIRE(status, resource->registeration_status);
 
-        if (AERON_CLIENT_REGISTRATION_STATUS_AWAITING == resource->registration_status &&
-            now_ns > resource->registration_deadline_ns)
+        if (AERON_CLIENT_REGISTRATION_STATUS_AWAITING == status && now_ns > resource->registration_deadline_ns)
         {
-            AERON_SET_RELEASE(resource->registration_status, AERON_CLIENT_REGISTRATION_STATUS_TIMED_OUT);
+            AERON_SET_RELEASE(status, AERON_CLIENT_REGISTRATION_STATUS_TIMED_OUT);
             work_count++;
         }
-        else if (AERON_CLIENT_REGISTRATION_STATUS_POLL_COMPLETED == resource->registration_status)
+        else if (AERON_CLIENT_REGISTRATION_STATUS_POLL_COMPLETED == status)
         {
             aeron_array_fast_unordered_remove(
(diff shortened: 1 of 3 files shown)


strangelydim:
Hi @egnchen . Good find - thanks for the report!  I believe there may be a few other places in the code where `resource->registration_status` is read without an acquire and would have a similar issue.

If you can review and fix the other reads of `registration_status` as well, and fix the `registeration_status` typo in your current commit, then we can go ahead and merge.  Thanks!

egnchen:
> If you can review and fix the other reads of registration_status as well

will do that in the morning

> and fix the registeration_status typo in your current commit, then we can go ahead and merge. Thanks!

awkward lol... fixed.

Also, since you're awake, there's another tsan related problem -

Have you guys ever compiled the c driver and c client under tsan? because while macros like `AERON_GET_ACQUIRE` and `AERON_SET_RELEASE` are correct memory order & assembly wise, they're not particularly thread-sanitizer friendly.

thread sanitizer by default can only understand C/C++ native atomic intrinsics, like C11 `atomic_thread_fence` or `std::atomic_thread_fence`. But it does not recognize inline asm like `__asm__ __volatile("" ::: "memory")`, which is what both macros does in `aeron_atomic_gcc_x86_64.h`.

This means if we compile aeron c driver/client with tsan and run it, it will report tons of false positives. So we either use an explicit tsan supression file or other magic to supress those warnings.

But a better way is to replace those macro defintions. Here's a proposed patch.
image

An alternative is to use `aeron_atomic_c11.h` under both X86 and ARM, but I'm not sure if that's what you want. Or we can just leave it, since it's not a correctness problem, just an ergonomics improvment, so really it's up to you guys, I could include a fix for that too if that's what you guys want.

egnchen:
@strangelydim updated.

egnchen:
> > If you can review and fix the other reads of registration_status as well
> 
> will do that in the morning
> 
> > and fix the registeration_status typo in your current commit, then we can go ahead and merge. Thanks!
> 
> awkward lol... fixed.
> 
> Also, since you're awake, there's another tsan related problem -
> 
> Have you guys ever compiled the c driver and c client under tsan? because while macros like `AERON_GET_ACQUIRE` and `AERON_SET_RELEASE` are correct memory order & assembly wise, they're not particularly thread-sanitizer friendly.
> 
> thread sanitizer by default can only understand C/C++ native atomic intrinsics, like C11 `atomic_thread_fence` or `std::atomic_thread_fence`. But it does not recognize inline asm like `__asm__ __volatile("" ::: "memory")`, which is what both macros does in `aeron_atomic_gcc_x86_64.h`.
> 
> This means if we compile aeron c driver/client with tsan and run it, it will report tons of false positives. So we either use an explicit tsan supression file or other magic to supress those warnings.
> 
> But a better way is to replace those macro defintions. Here's a proposed patch. image
> 
> An alternative is to use `aeron_atomic_c11.h` under both X86 and ARM, but I'm not sure if that's what you want. Or we can just leave it, since it's not a correctness problem, just an ergonomics improvment, so really it's up to you guys, I could include a fix for that too if that's what you guys want.

@vyazelenko @strangelydim Can I get a response on this? Much appreciated.

PR #2050, opened 2026-05-15
Roll debug log files based on file size

mikeb01:


aeron-agent/src/main/java/io/aeron/agent/EventLogReaderAgent.java
```diff
@@ -266,5 +277,99 @@ private static void write(final ByteBuffer buffer, final FileChannel fileChannel
...
+    {
+        final String maxFileLengthStr = configOptions.get(ConfigOption.LOG_FILE_MAX_LENGTH);
+        try
+        {
+            return null != maxFileLengthStr ?
+                parseSize(ConfigOption.LOG_FILE_MAX_LENGTH, maxFileLengthStr) : Long.MAX_VALUE;
+        }
+        catch (final NumberFormatException ex)
+        {
+            System.err.println(ex.getMessage() + "  Disabling log rotation.");
```
vyazelenko: Why bother logging this? Shouldn't this be an error instead?

aeron-agent/src/main/java/io/aeron/agent/EventLogReaderAgent.java
```diff
@@ -266,5 +277,99 @@ private static void write(final ByteBuffer buffer, final FileChannel fileChannel
...
+            rolledFile = Path.of(filename + "." + nextFileIndex);
+            nextFileIndex++;
+        }
+        while (Files.exists(rolledFile));
+
+        Files.move(file, rolledFile);
+
+        fileChannel = open(Paths.get(filename), CREATE_NEW, APPEND, WRITE);
+
+        appendFileHeader(nanoClock.nanoTime(), startTimeMs);
```
vyazelenko: This logs the original start of the logging message but with a new nanos timestamp? Isn't this confusing, i.e. should we instead show the time when this file was created?

aeron-agent/src/main/java/io/aeron/agent/EventLogReaderAgent.java
mikeb01: Given that we're in the scope of the agent, I didn't want to have it interact with the DistinctErrorLog or the error handler as the driver context or any of that configuration is not in scope.  I also didn't really want to prevent the service from starting or stopping the debug logging in this case either so it seemed the most appropriate.

aeron-agent/src/main/java/io/aeron/agent/EventLogReaderAgent.java
mikeb01: Fixed.

mikeb01: Fixed.

aeron-agent/src/main/java/io/aeron/agent/EventLogReaderAgent.java
```diff
@@ -60,28 +65,60 @@ public final class EventLogReaderAgent implements Agent
...
     private final Int2ObjectHashMap<ComponentLogger> loggers = new Int2ObjectHashMap<>();
     private final PrintStream out;
     private final NanoClock nanoClock;
     private final EpochClock epochClock;
+    private final long maxFileLength;
+    private final String filename;
+    private int nextFileIndex = 1;
+
+    private FileChannel fileChannel;
+    private long startTimeMs;
```
vyazelenko: Should be a local variable since it is only used once.

aeron-agent/src/main/java/io/aeron/agent/EventLogReaderAgent.java
vyazelenko: Then I suggest we put prefix before the exception message, e.g. `Disabling log rotation due to: xxx`.

vyazelenko (changes requested):

aeron-agent/src/main/java/io/aeron/agent/EventLogReaderAgent.java
mikeb01: Implemented better message.

PR #2054, opened 2026-05-22
when onTruncateLogEntry is called, make sure to update the log position to the new position

nbradac:


vyazelenko:
This is not correct, because in the common case we have `logPosition <= commitPosition < appendPosition`. If post truncation `logPosition > appendPosition` it is a symptom of larger problem whereby truncation erased committed data.

PR #2055, opened 2026-05-28
truncateRecording now does a stopAllReplays and waits for outstanding replays to complete

nbradac:


aeron-archive/src/main/java/io/aeron/archive/GatedDeleteSegmentsSession.java
```diff
@@ -0,0 +1,59 @@
...
+import org.agrona.ErrorHandler;
+
+import java.io.File;
+import java.util.ArrayDeque;
+import java.util.function.BooleanSupplier;
+
+class GatedDeleteSegmentsSession extends DeleteSegmentsSession
+{
+    private final BooleanSupplier gate;
+
```
vyazelenko: I think even simpler implementation is to move the logic to the parent class and make it state-specific. 

The `DeleteSegmentsSession` can be in one of three states:
* `AWAIT_REPLAYS_STOP` - this is used by the truncate command as an initial state. Here we call `conductor.hasInProgressReplays(recordingId)` and if it returns `false` transition to delete.
* `DELETE_FILES` - all other delete commands (e.g. purge) will start with this state.
* `DONE` - terminal state when all files were deleted.

Allow state transition: `AWAIT_REPLAYS_STOP -> DELETE_FILES -> DONE`.

CHANGELOG.adoc
```diff
@@ -2,6 +2,7 @@
 
 == 1.52.0 (Unreleased)
 
+* **[Archive]** Truncating a recording now stops any concurrent replays of that recording before proceeding with the deletion of segments.
```
vyazelenko, line 10: Should be moved to `New & Noteworthy` section.

aeron-archive/src/main/java/io/aeron/archive/client/AeronArchive.java
```diff
@@ -1667,6 +1667,7 @@ public long findLastMatchingRecording(
     /**
      * Truncate a stopped recording to a given position that is less than the stopped position. The provided position
      * must be on a fragment boundary. Truncating a recording to the start position effectively deletes the recording.
+     * Truncating a recording will stop any concurrent replays of that recording.
```
vyazelenko, line 1671: Add `<p>` before the paragraph.

aeron-archive/src/main/java/io/aeron/archive/client/ArchiveProxy.java
```diff
@@ -821,6 +821,7 @@ public boolean getRecordingPosition(final long recordingId, final long correlati
     /**
      * Truncate a stopped recording to a given position that is less than the stopped position. The provided position
      * must be on a fragment boundary. Truncating a recording to the start position effectively deletes the recording.
+     * Truncating a recording will stop any concurrent replays of that recording.
```
vyazelenko, line 825: Add <p> before the paragraph.

vyazelenko (changes requested):

PR #2057, opened 2026-05-29
[Java] CreateReplayPublicationSession retries publication create on session id clash

nbradac:


vyazelenko:
Session conflicts will be dealt with at the driver conductor level. Also `CreateReplayPublicationSession` will probably go away.

PR #2058, opened 2026-06-03
Fixed accidentally-added "final" modifiers for several classes

strangelydim:
Classes that are either required to be extended or might plausibly be usefully extended:

- io.aeron.cluster.AppVersionValidator
- io.aeron.Aeron.Context
- io.aeron.status.HeartbeatTimestamp
- io.aeron.agent.ArchiveComponentLogger
- io.aeron.Aeron.Configuration
- io.aeron.cluster.ClusterBackup.Configuration
- io.aeron.driver.status.ClientHeartbeatTimestamp

I've reverted those & made sure checkstyle / javadoc passes.

There were a lot of other classes (40+) marked `final` in 1.51.0 as well, but I did not revert those - the ones in the list above are the ones that I think someone might actually have a reason to be non-final.  My opinion is it's nice to make things final if you can, so we should leave the rest of the `final`s in until someone actually complains.

### DIFF
--- a/aeron-agent/src/main/java/io/aeron/agent/ArchiveComponentLogger.java
+++ b/aeron-agent/src/main/java/io/aeron/agent/ArchiveComponentLogger.java
@@ -36,7 +36,7 @@ import static net.bytebuddy.matcher.ElementMatchers.named;
  * Implementation of a component logger for archive log events.
  */
 @Versioned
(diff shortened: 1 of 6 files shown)


vyazelenko:
@strangelydim Out of all those classes only `AppVersionValidator` should not be final and even that should be replaced with an interface instead of allowing sub-classing.

PR #2065, opened 2026-06-19
[C] fix race condition between sender thread and conductor adding spy…

nbradac:
… to subscribable list


vyazelenko:
@nbradac The same issue exists in the Java driver btw. I think a much simpler approach is to eliminate this race entirely, i.e. compute max spy position on the conductor duty cycle and expose it is as volatile field to the sender thread. I'll work on a fix later today.

PR #2067, opened 2026-06-22
Fixed accidental send channel endpoint map removal

strangelydim:
Removing an endpoint from the map in `aeron_send_channel_endpoint_entry_delete` just removed it keyed off the canonical form of the endpoint without checking that the entry in the map really did refer to the endpoint supposedly being removed.  Which meant that if a _new_ endpoint had been added to the map in between the old endpoint's socket being closed and `aeron_send_channel_endpoint_entry_delete` being called for the old endpoint, then `aeron_send_channel_endpoint_entry_delete` would remove the map entry for what was now the _new_ endpoint.

### DIFF
--- a/aeron-driver/src/main/c/aeron_driver_conductor.c
+++ b/aeron-driver/src/main/c/aeron_driver_conductor.c
@@ -1574,10 +1574,19 @@ bool aeron_send_channel_endpoint_entry_has_reached_end_of_life(
 void aeron_send_channel_endpoint_entry_delete(
     aeron_driver_conductor_t *conductor, aeron_send_channel_endpoint_entry_t *entry)
 {
-    aeron_str_to_ptr_hash_map_remove(
+    // Only remove the map entry if it still refers to this endpoint. A new endpoint
+    // with the same canonical form may have already been created by the time we get
+    // here, in which case we don't want to remove that new endpoint's map entry.
+    if (entry->endpoint == aeron_str_to_ptr_hash_map_get(
         &conductor->send_channel_endpoint_by_channel_map,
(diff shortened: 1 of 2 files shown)


vyazelenko:
What is the exact set of steps to trigger this?

I can think of three scenarios on how `add/removePublication` might play out:
1. `removePublication` completes first, i.e. endpoint is removed and socket is closed.
   In this case `addPublication` will complete without an error and will create a new `send_channel_endpoint_t`.

2. `addPublication` executes before `removePublication`.
   In this case ref count on the `send_channel_endpoint_t` will be incremented by one and this will prevent `send_channel_endpoint_t` from being removed when `removePublication` eventually executes.

4. `addPublication` and `removePublication` race with each other, i.e. `removePublication` executes first and notifies `sender` thread:
    1. `sender` finishes first and conductor processes timer events before `add`:
        1. `endpoint` is completely removed by the time `add` is processed (i.e. just like case 1).
        2. `endpoint` is marked as `AERON_SEND_CHANNEL_ENDPOINT_STATUS_CLOSING` before notifying the sender.
          If `add` is processed while endpoint is in this state it will fail with `AERON_ERROR_CODE_RESOURCE_TEMPORARILY_UNAVAILABLE`
https://github.com/aeron-io/aeron/blob/570c86a96013837569eef1d6e64daba6cd4170fd/aeron-driver/src/main/c/aeron_driver_conductor.c#L4114-L4121
          
    2. `sender` takes longer or conductor does not processes timer events in time, i.e. `add` is executed first.
     This case is the same as the case 2.

strangelydim:
Closing this PR in favor of https://github.com/aeron-io/aeron/pull/2075

PR #2080, opened 2026-06-24
Added sequencer duty cycle tracker counters for snapshot/replay-service

pveentjer:
The counters were added based on these 2 PRs

https://github.com/weareadaptive/aeron-sequencer/pull/435
https://github.com/weareadaptive/aeron-sequencer/pull/434


vyazelenko:
These also must be added to [aeron_counters.h](https://github.com/aeron-io/aeron/blob/master/aeron-client/src/main/c/aeron_counters.h).

PR #2087, opened 2026-06-28
Test add regression tests for subscription

amirdeljouyi:
Hi,

I added two regression tests for `Subscription`.

The tests cover removing an image by correlation ID and describing a subscription after its connection state and metadata have been set up. This checks the image-list update, resource cleanup.

Impact on coverage:

- Covers the relevant `Subscription` logic, including [image removal and cleanup](https://github.com/real-logic/aeron/blob/a28f85e7e76df00ff8f2533db11324a686410ed9/aeron-client/src/main/java/io/aeron/Subscription.java#L605-L629).
- Adds coverage for `Subscription` lines, about 18.03% line coverage.

Why:

While looking through `Subscription`, I noticed these behaviors were not covered directly by the existing tests.

I hope you find these tests useful. 😄 Please let me know if you have any questions or concerns.

Thanks,
Amir


vyazelenko:
@amirdeljouyi `Subscription#removeImage` is internal API that is used by `ClientConductor` when image goes unavailable. It is tested indirectly by any test that receives `UnavailableImageHandler#onUnavailableImage` notifications which have lots of given how core this functionality is.

`Subscription#toString` format is for debugging purposes only. We don't guarantee any format of that string.

Given the above I'm inclined to close this PR.

PR #2093, opened 2026-07-12
[Cluster] Send a termination request to the ClusteredServiceContainer when the ConsensusModuleAgent is closed

mikeb01:
Use a clean termination when the ClusteredServiceContainer closed.

---

> [!NOTE]
> **Medium Risk**
> Changes cluster shutdown coordination and bumps the inter-component SBE schema (v17), which can affect rolling upgrades and termination ordering; behavior is covered by a new graceful-stop test.
> 
> **Overview**
> When **ConsensusModuleAgent** closes, it now notifies clustered services at the current **commit position** with **`ackRequired=false`**, so services run **`onTerminate`** and exit without blocking on a consensus-module ack while the module is already shutting down. Cluster-driven shutdown, abort, and follower termination paths still send **`ackRequired=true`**.
> 
> The **`ServiceTerminationPosition`** wire message gains an **`ackRequired`** field (schema **v17**). **`ClusteredServiceAgent`** only sends the termination ack to the consensus module when that flag is set.
> 
> **`ClusterTerminationException`** now inherits expected/unexpected termination from Agrona’s **`AgentTerminationException`** instead of storing its own flag.
> 
> Adds a system test for graceful follower stop plus test helpers; bumps Agrona to **2.6.0-SNAPSHOT** and adds **`-Xshare:off`** for JVM tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8c4a1d09c76ea8b44c5c8cdb6cac253b5a928d4d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModuleAgent.java
+++ b/aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModuleAgent.java
@@ -282,6 +282,30 @@ final class ConsensusModuleAgent
     {
         if (!aeron.isClosed())
         {
+            final long logPosition = commitPosition.get();
+            serviceProxy.terminationPosition(logPosition, ctx.countedErrorHandler());
+
+//            try
+//            {
+//                idleStrategy.reset();
+//                while (!ServiceAck.hasReached(logPosition, serviceAckId, serviceAckQueues))
+//                {
+//                    final int workCount = consensusModuleAdapter.poll();
+//                    if (Thread.currentThread().isInterrupted())
+//                    {
+//                        throw new InterruptedException();
+//                    }
+//                    aeronClientInvoker.invoke();
+//
+//                    idleStrategy.idle(workCount);
+//                }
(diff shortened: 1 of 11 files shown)


marc-adaptive:
LGTM. 

Wondering what your thoughts are on waiting for service ack (and commented out code)?

Also wondering your thoughts on the error handlers being invoked? From the CS side - if the CM has closed CMAdapter I’m guessing you may get a false positive error on service ack offer, but maybe that's okay?

mikeb01:
I had originally though to wait for the ack, but things get hairy when waiting for things to happen during close.  I tested it and found it created more problems than it solved.  There is a very small window where a false positive failure could still occur, but with several thousand test iterations I was unable to replicate it.  The window could be removed, but the added complexity didn't seem worth it.

marc-adaptive (approved):

PR #2095, opened 2026-07-16
Added "isLeadershipConfirmedSince" method to the ConsensusModuleAgent API

strangelydim:
Given a timestamp, isLeadershipConfirmedSince returns true if a quorum of cluster members has actively ACK'd and confirmed that the leader is still the leader sometime _after_ the timestamp.

This allows users to confirm a given quorum read is linearizably "fresh" (issue the read, then check that isLeadershipConfirmedSince == true); otherwise, a user would have to issue a read and then perform an expensive no-op write operation as a barrier to ensure linearizable freshness.

---

> [!NOTE]
> **High Risk**
> Changes consensus wire format and leader/follower commit-path behaviour; incorrect quorum or token handling could allow stale linearizable reads or break mixed-version clusters until all nodes upgrade.
> 
> **Overview**
> Adds **ReadIndex-style leadership confirmation** so a leader can verify a quorum still recognises it **after** a read point, without a log barrier write. Extensions call **`triggerQuorumConfirmation()`** for a term-scoped token, then poll **`isLeadershipConfirmedSince(token)`** from consensus work.
> 
> The wire protocol bumps cluster codecs to **v17**: **`CommitPosition`** carries a monotonic **`confirmationCounter`** (advanced only on explicit confirmation rounds, not every commit broadcast), and followers reply with a new **`LeadershipConfirmAck`**. Quorum logic counts members who echoed a counter **strictly beyond** the token in the **current leadership term** (wrap-safe), so stale in-flight acks and prior-term echoes cannot confirm.
> 
> **`ConsensusModuleAgent`** coalesces confirmation rounds per duty cycle, sends an absent sentinel during elections, and resets confirmation state on **`electionComplete`**. **`ClusterMember.hasQuorumConfirmedSince`** implements the quorum check with per-member echo tracking.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3963362e92fe1c20f95f5cbc3432d26fd6245005. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/aeron-cluster/src/main/java/io/aeron/cluster/ClusterMember.java
+++ b/aeron-cluster/src/main/java/io/aeron/cluster/ClusterMember.java
@@ -53,6 +53,11 @@ public final class ClusterMember
     private long changeCorrelationId = NULL_VALUE;
     private long logPosition = NULL_POSITION;
     private long timeOfLastAppendPositionNs = NULL_VALUE;
+    // ReadIndex leadership confirmation: the highest counter this member echoed (LeadershipConfirmAck) and the
+    // term it echoed it in. The term gate is essential -- a leftover counter from a prior leadership must never
+    // count towards a confirmation in the current term. Compared wrap-safe.
+    private int confirmationCounter;
+    private long confirmationCounterTermId = NULL_VALUE;
     private ExclusivePublication publication;
     private String consensusChannel;
     private final String consensusEndpoint;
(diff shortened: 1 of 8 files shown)


strangelydim:
bugbot run

strangelydim:
bugbot run

PR #2117, opened 2026-08-10
Fixes delay in sending the response setup message.

pveentjer:
What happens is that the conductor sets the responseSessionId on the PublicationImage, but the receiver is the thread that builds and sends the response setup message. And it will only send it after the PublicationImage nextSmDeadlineNs expires. And that could be up to 200ms later.

And this delays finishing the connection up to 200ms.

This change does 2 things:

1) The conductor will ask the PublicationImage for a nextSmDeadline reset
   using the new requestNextSmDeadlineReset method (and its C version
   aeron_publication_image_request_next_sm_deadline_reset)

2) In the old code, there was a race condition and data race on the
   nextSmDeadlineNs. The onTimeEvent is called from the Conductor and was
   writing to this nextSmDeadlineNs directly without furter synchronization;
   this field is a plain field.

   This has now been fixed; so external threads can raise the request for
   resetting the timer by arming the isNextSmDeadlineResetRequested.
   This field is accessed using release/acquire semantics; so that solves the
   data race and race condition. And the receiver thread will eventually
   detect that the timer needs to be reset, and will then reset it. So the
   actual timer field is now only access by the receiver thread exclusively.

This fix will not bring the 200ms down to nearly zero; in my benchmarks it went down from 214 ms → 117 ms.

---

> [!NOTE]
> **Medium Risk**
> Touches cross-thread synchronization on the hot receiver SM path and response-connection setup; behavior change is localized but timing-sensitive for media driver correctness.
> 
> **Overview**
> **Response Setup** was gated on the receiver’s status-message timer (`nextSmDeadlineNs`), so after the conductor set `responseSessionId` the client could wait up to ~200ms before the setup frame went out.
> 
> The conductor now calls **`requestNextSmDeadlineReset()`** (C: `aeron_publication_image_request_next_sm_deadline_reset`) whenever it assigns a response session id, and uses the same path when forcing EOS SM on publication revoke instead of writing `nextSmDeadlineNs` from the conductor thread.
> 
> A new **`isNextSmDeadlineResetRequested`** flag is set with release semantics from any thread; the receiver clears it and sets `nextSmDeadlineNs = now - 1` in `sendPendingStatusMessage`, so only the receiver mutates the deadline and the prior cross-thread write race is removed. Java and C drivers are aligned.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6bf35b9c84f3b128ef36812e4869dc9ece747baf. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/aeron-driver/src/main/c/aeron_driver_conductor.c
+++ b/aeron-driver/src/main/c/aeron_driver_conductor.c
@@ -4204,6 +4204,7 @@ aeron_driver_conductor_command_state_t aeron_driver_conductor_execute_add_networ
     {
         aeron_publication_image_set_response_session_id(
             async_command->response_publication_image, (int64_t)publication->session_id);
+        aeron_publication_image_request_next_sm_deadline_reset(async_command->response_publication_image);
     }
 
     return AERON_DRIVER_CONDUCTOR_COMMAND_STATE_DONE;
--- a/aeron-driver/src/main/c/aeron_publication_image.c
+++ b/aeron-driver/src/main/c/aeron_publication_image.c
@@ -380,6 +380,7 @@ int aeron_publication_image_create(
     _image->time_of_last_packet_ns = now_ns;
     _image->next_sm_deadline_ns = 0;
     _image->is_sm_enabled = true;
+    _image->is_next_sm_deadline_reset_requested = false;
     _image->conductor_fields.clean_position = initial_position;
     _image->conductor_fields.time_of_last_state_change_ns = now_ns;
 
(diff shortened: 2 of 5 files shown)


mikeb01:
I'll dig into this tomorrow.

pveentjer:
Thanks @mikeb01 . There will be a follow up PR because I believe there are some more data races on fields (trackRebuild, isSendingEosSm)

PR #2125, opened 2026-08-24
[Driver/Java] Add cgroup validation for Java Driver.

EmilJohn24:


vyazelenko:
@EmilJohn24 If this intended for the media driver the code should live in the `aeron-driver` module **not** in the `aeron-client`.

aeron-client/src/main/java/io/aeron/topology/AffinityParser.java
```diff
@@ -0,0 +1,72 @@
...
+     */
+    public static void parse(final String affinity, final IntConsumer consumer)
+    {
+        for (final String rawPart : affinity.split(","))
+        {
+            final String part = rawPart.trim();
+            final int dashIndex = part.indexOf('-');
+            if (-1 != dashIndex)
+            {
+                final int start = Integer.parseInt(part, 0, dashIndex, 10);
```
vyazelenko: Agrona has better functions to parse int/long from a String, i.e. `org.agrona.AsciiEncoding#parseIntAscii` and `org.agrona.AsciiEncoding#parseLongAscii`. Please used those.

EmilJohn24:
> @EmilJohn24 If this intended for the media driver the code should live in the `aeron-driver` module **not** in the `aeron-client`.

@vyazelenko I put it in `aeron-client` originally because the validation utilities might be used for archive and cluster thread affinities eventually once we start fleshing out the Java CPU pinning feature. I'll move it to the `aeron-driver` module for now since this is currently  the only place it is used for now.

vyazelenko:
Alright, that makes sense. Keep the code in the `aeron-client`.

aeron-client/src/main/java/io/aeron/topology/ThreadAlignmentValidator.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    record MissingSibling(int siblingCpu, int coreCpu)
+    {
+    }
+
+    List<MissingSibling> findMissingSiblings(final IntArrayList cpuList) throws IOException
+    {
+        final IntHashSet presentCpus = new IntHashSet();
+        for (int i = 0; i < cpuList.size(); i++)
```
mikeb01: Can use presentCpus.addAll I think.

mikeb01:
Apart from the one small change, I think this is ready to go.

PR #2128, opened 2026-08-29
Support for 32-bit Aeron

xbotuk:


vyazelenko:
See https://github.com/aeron-io/aeron/issues/2101#issuecomment-5572432840.

PR #2135, opened 2026-09-09
[C++] Read data frame header values lazily in the wrapper Header

Magnushst:
# [C++] Read data frame header values lazily in the wrapper Header

`Header`'s constructor calls `aeron_header_values`, copying 48 bytes of frame metadata out of the log buffer. A `Header` is built for **every fragment** delivered to a C++ handler, so the copy is paid whether or not the handler reads it. A `FragmentAssembler` chain pays twice per fragment and the first copy is always wasted: `Subscription::poll` builds a `Header`, `FragmentAssembler` passes only `header.hdr()` to the C assembler, which builds a second `Header` for the delegate.

The values are now read on the first accessor that needs them; `position()`, `context()` and `hdr()` go through the wrapped `aeron_header_t` and never trigger the copy.

## Correctness

Values, accessor semantics and exception mapping are unchanged. The wrapped `aeron_header_t` is valid for the duration of the poll callback, which is already a `Header`'s lifetime and was already required by `position()` and `context()`; this matches the pre-wrapper C++ and the Java `Header`. The constructor still rejects a null header, and a `Header` is per-fragment and never shared, so the `mutable` state introduces no sharing.

New `HeaderTest` covers every accessor, read order, repeated and `const` reads, the fragmented-frame position and a null header. New `PubSubTest.shouldDeliverHeaderValuesForAssembledMessages` asserts payload and header values received through a `FragmentAssembler` over IPC against an embedded C driver, at 32 bytes, `maxPayloadLength`, and 2 and 3 fragments. Both also pass against the previous implementation.

## Performance

`aeron_header_values` calls counted with an `LD_PRELOAD` interposer over exactly 1,000,000 delivered fragments; no configuration copies more than before:

| Delivery path | before | after |
| --- | ---: | ---: |
| `FragmentAssembler`, delegate ignores header | 2,000,000 | 0 |
| `FragmentAssembler`, delegate reads one field | 2,000,000 | 1,000,000 |
| direct handler, ignores header | 1,000,000 | 0 |
| direct handler, reads one field | 1,000,000 | 1,000,000 |

On an IPC receive benchmark (embedded C driver, 32-byte validated messages, poll phase only) the `FragmentAssembler` path improves from 10.5 to 9.3 ns/fragment on gcc 14.2 and 15.4 to 10.7 on clang 19.1, control unchanged. WSL2 on an i9-13905H, ~±15% spread, so the counts above are the reliable evidence.

## Trade-offs

Each value accessor now tests a `bool`, costing a header-reading handler up to ~0.5 ns/fragment on gcc and nothing measurable on clang; `Header` grows from 56 to 64 bytes as a per-fragment stack object. A `Header` first read after the poll callback returns now reads through a pointer the C client no longer keeps alive (already true of `position()` and `context()`).


pveentjer:
Hi @Magnushst thanks you PR. I'm currently looking into it. I'm checking out a different approach and will come back to you.

pveentjer:
I have been playing with the following approach: https://github.com/aeron-io/aeron/pull/2142

So the idea is to provide access to the aeron_data_header_t (which is already part of the API) so this whole field copying isn't needed. Using this approach you get a direct pointer to the frame header in the term buffer.

PR #2136, opened 2026-09-10
[Cluster] Introduce ExtendedTerminationHook.

wojciech-adaptive:
Needed to tell apart expected and unexpected terminations. Not changing the existing hook to maintain compatibility.

### DIFF
--- a/aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModule.java
+++ b/aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModule.java
@@ -1677,6 +1677,7 @@ public final class ConsensusModule implements AutoCloseable
         private Counter electionCounter;
         private Counter leadershipTermId;
         private Runnable terminationHook;
+        private ExtendedTerminationHook extendedTerminationHook;
 
         private AeronArchive.Context archiveContext;
         private AuthenticatorSupplier authenticatorSupplier;
@@ -2108,6 +2109,11 @@ public final class ConsensusModule implements AutoCloseable
             if (null == terminationHook)
             {
                 terminationHook = () -> {};
+            }
+
+            if (null == extendedTerminationHook)
+            {
+                extendedTerminationHook = cause -> {};
             }
 
             if (null == authenticatorSupplier)
@@ -4089,8 +4095,10 @@ public final class ConsensusModule implements AutoCloseable
 
         /**
          * Set the {@link Runnable} that is called when the {@link ConsensusModule} processes a termination action.
+         * <p>
+         * Both extendedTerminationHook and terminationHook run on termination, in that order.
          *
-         * @param terminationHook that can be used to terminate a consensus module.
+         * @param terminationHook that is called when the {@link ConsensusModule} processes a termination action.
          * @return this for a fluent API.
          */
         public Context terminationHook(final Runnable terminationHook)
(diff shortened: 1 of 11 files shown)


aeron-cluster/src/main/java/io/aeron/cluster/ExtendedTerminationHook.java
```diff
@@ -0,0 +1,29 @@
...
+ * Represents a callback run when a cluster component terminates.
+ */
+public interface ExtendedTerminationHook
+{
+    /**
+     * Called when a cluster component terminates.
+     *
+     * @param cause the exception that caused this termination or {@code null} if not caused by an exception
+     */
+    void run(Throwable cause);
```
vyazelenko: The only exception type that can lead to termination is `org.agrona.concurrent.AgentTerminationException` and its subclasses. Why have `Throwable` in the signature?

aeron-cluster/src/main/java/io/aeron/cluster/ExtendedTerminationHook.java
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * Represents a callback run when a cluster component terminates.
+ */
+public interface ExtendedTerminationHook
+{
+    /**
+     * Called when a cluster component terminates.
+     *
+     * @param cause the exception that caused this termination or {@code null} if not caused by an exception
```
vyazelenko: We should clarify how to distinguish between expected and unexpected terminations. I guess `null` should be treated as expected termination?

aeron-cluster/src/test/java/io/aeron/cluster/service/ClusteredServiceAgentTest.java
```diff
@@ -150,4 +169,30 @@ void shouldLogErrorInsteadOfThrowingIfSessionIsNotFoundOnClose()
...
+            .terminationHook(terminationHook)
+            .extendedTerminationHook(extendedTerminationHook);
+
+        final ClusteredServiceAgent clusteredServiceAgent = new ClusteredServiceAgent(ctx);
+
+        final AgentTerminationException terminationException =
+            assertThrows(AgentTerminationException.class, clusteredServiceAgent::doWork);
+
+        verify(terminationHook).run();
+        verify(extendedTerminationHook).run(terminationException);
```
vyazelenko: Do we want to assert/verify an exact order of these hooks being called?

aeron-cluster/src/main/java/io/aeron/cluster/ConsensusModule.java
```diff
@@ -4102,13 +4108,43 @@ public Context terminationHook(final Runnable terminationHook)
...
+        }
+
+        /**
+         * Get the {@link ExtendedTerminationHook} that is called when the {@link ConsensusModule} processes
+         * a termination action.
+         * <p>
+         * Identical to {@link #terminationHook(Runnable)}, but receives information about the termination cause.
+         *
+         * @return the {@link ExtendedTerminationHook} that is called when the {@link ConsensusModule} processes
+         * a termination action.
```
vyazelenko, line 4149: We should document that *both* hooks will be called and the order of the invocation. I guess in the Javadoc of each method.

vyazelenko (changes requested):
Overall looks good. I think the interface should be type specific.

wojciech-adaptive:
> I think the interface should be type specific.

You mean an interface for each component, e.g. `ConsensusModuleTerminationHook`?

aeron-cluster/src/main/java/io/aeron/cluster/ExtendedTerminationHook.java
wojciech-adaptive: To be more generic, just like allowing nulls. If you're confident we won't need that in future, happy to change.

vyazelenko: `AgentTerminationException` is pretty stable and we should specialise for it.

vyazelenko:
> > I think the interface should be type specific.
> 
> You mean an interface for each component, e.g. `ConsensusModuleTerminationHook`?

No, just the `AgentTerminationException` part.

aeron-cluster/src/main/java/io/aeron/cluster/ExtendedTerminationHook.java
wojciech-adaptive: We wanted null for unknown future cases, so hard to say. If you want the contract to be very specific, then maybe we should not allow nulls?

vyazelenko: I agree. Let's remove `null` as an option.

PR #2144, opened 2026-09-16
AbstractMinMulticastFlowControl drops live receivers.

pveentjer:
If you have the following set of receivers
    
[r0, R1, r2]
    
where r0, r2 are idle and R1 is not. Then after the onIdle completes, the array will be [r2] instead of [R1].
    
So the live receiver R1 is dropped while not dropping the idle receiver r2.
    
In case of fc=min
    
Imagine that R1 would be the receiver furthest behind. Because it is dropped from the flow control, the producer can now blast past R1.

SendChannelEndpoint has a similar bug. I'll send a PR shortly.


pveentjer:
The aeron_min_flow_control_strategy_on_idle has a similar problem. I'll a PR for that one as well.

PR #2145, opened 2026-09-18
DynamicSndMultiDestination drops live destinations.

pveentjer:
If you have the following set of MDC destinations

[d0, D1, d2]

where d0, d2 are idle and D1 is not. Then after removeInactiveDestinations completes, the array will be [d2] instead of [D1].

So the live destination D1 is dropped while the idle destination d2 is kept. The publisher stops sending to a receiver that is still there, and keeps sending to one that has timed out.

Same defect as AbstractMinMulticastFlowControl.onIdle: https://github.com/aeron-io/aeron/pull/2144


vyazelenko:
I fixed this already in https://github.com/aeron-io/aeron/commit/753834136eb4d43650ace9018639bfd348167078.

PR #2147, opened 2026-09-18
[Cluster] Commits can be lost

pveentjer:
Commits can be lost because the quorumPosition is determined based on the available members and not the members that share the current leader. So a member that doesn't share the same leader, can still signal writes as committed.

Here is an example situation with the above problem. The arrow is the member that is leader.
```
step 0   initial state, term 1, leader m3
        m1  t1 []
        m2  t1 []
    ->  m3  t1 []

step 1   1 is written to m3
        m1  t1 []
        m2  t1 []
    ->  m3  t1 [1]

step 2   m3 goes down
        m1  t1 []
        m2  t1 []
        m3  down, on disk [1]

step 3   m2 is elected, term 2
        m1  t2 []
    ->  m2  t2 []
        m3  down, on disk [1]

step 4   2 is written to m2
        m1  t2 []
    ->  m2  t2 [2]
        m3  down, on disk [1]

step 5   m3 comes up, still reporting term 1
        m1  t2 []
    ->  m2  t2 [2]
        m3  t1 [1]
    m3's log position makes the quorum look reached, m2 acks [2] as committed

step 6   m2 goes down
        m1  t2 []
        m2  down, on disk [2]
        m3  t1 [1]

step 7   election, m1 wins on term, m3 truncates
    ->  m1  t3 []
        m3  t3 []
    [2] was acknowledged and is gone
```


pveentjer:
I'm going to close this one for now. I'll come back shortly.

pveentjer:
I'm working in a reproducer with an actual cluster and a lot of extra synchronization to trigger the behavior.

